PR #160, opened 2018-02-21
increase inline asset size limit to 32k

maxlk:
Let's increase inline asset size limit to 32k
discussed here: https://wix.slack.com/archives/C802VSGS1/p1519138341000392?thread_ts=1519136618.000577&cid=C802VSGS1


ronami:
Closing. Please re-open an issue if this is still relevant. Thanks!

maxlk:
@ronami
We have a control that is built up from several thumbnails, and some of these thumbnails looks ugly being compressed to 10k size.
We need a way to bundle those thumbnails together. Sprites doesn't suit well, because some of those thumbnails are photo-like and they are too big being saved as PNG, they have good quality/size balance being saved as JPEG, but JPEG doesn't suit well for sprites.
So, I want to bundle them together with webpack, but I cannot do it in Yoshi, because I don't have access to url-loader configuration.

PR #205, opened 2018-04-16
Add babel-preset-yoshi.

yavorsky:
Allow users to use our own preset that pre-configured to be the best solution for users who use yoshi + babel.

The basis is:

- [preset-env](https://babeljs.io/docs/plugins/preset-env) for ESNext to ES5 transform. Moreover, you can customize current targets and module type to build for. Will use all targets and commonjs module type as a default values.
- [preset-react](https://babeljs.io/docs/plugins/preset-react) for JSX and Flow transforms.
- Stage 2 and 3 plugins: [transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties), [transform-decorators](https://babeljs.io/docs/plugins/transform-decorators) (legacy).
- [remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) only for production builds.
- [syntax-dynamic-import](https://babeljs.io/docs/plugins/syntax-dynamic-import) - just syntax support since all transforms by webpack.

**Supported Options**
- `targets`: Avoid redundant transformations if specified targets already support some of ESNext features. See more [preset-env#targets](https://babeljs.io/docs/plugins/preset-env/#targets).
- `modules` (default: commonjs): Enable transformation of ES6 module syntax to another module type. See [preset-env#modules](https://babeljs.io/docs/plugins/preset-env/#modules).

In future, we could update babel to 7 or 8 and just update config. So we could publish yoshi with any babel updated w/o breaking changes.


yavorsky:
awesome!

PR #255, opened 2018-05-03
Support test debugging using `--debug` option

hadar111:
### 🔦 Summary
Add `--debug` flag support.
Enable tests (mocha and jest) debugging.
* Mocha, Jest and protractor is now running through cli api instead of haste task.  
* Yoshi is now normalizing debugging args (turn `--inspect(--brk)` into `--debug`  which is the debug flags passed by IDEs like VScode and Webstorm)
* Add deprecation warning to users of `./debug/mocha.js`


ranyitz:
Continued on #264 and released on [2.3.0](https://github.com/wix/yoshi/blob/master/CHANGELOG.md#230-may-9-2018)

PR #278, opened 2018-05-12
[WIP] Replace mocha with jest

yavorsky:
### 🔦 Summary
This PR is an attempt to replace mocha and all related stuff with jest. I'll update a description during this PR progress.

### 🗺️ Test Plan
😫

### DIFF
--- a/packages/yoshi/.babelrc
+++ b/packages/yoshi/.babelrc
@@ -1,0 +1,11 @@
+{
+  "env": {
+    "test": {
+      "presets": ["react", [ "env", {
+        "targets": {
+          "node": "current"
+        }
+      }]]
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


ranyitz:
This is a good attempt, we do want to move to jest, but we should find a way to do it gradually. The tests always changes and the conflicts going to be harder and harder to merge.

We can start by using the `expect` library (instead of `chai`), and then use `jest --runInBand` before going parallelized.

Right now, i've seen some problems with the output that should be solved before moving to jest.

closing for now.

ranyitz:
Great idea ❤️

I think that we need to make the `build`/`start` tests work with jest, and than the rest would be fairly easy.

Right now there is a problem with the reporting but I couldn't managed to invest more than a few minutes.

PR #287, opened 2018-05-14
Force Typescript module configuration for client code

yairhaimo:
### 🔦 Summary
Force webpack's typescript loader to use `es2015` modules in order to support tree shaking out of the box.
Force haste's typescript task to use `commonjs` modules in order to support serverside typescript.

### 🗺️ Test Plan
 - Check that an unused function was removed from the production minified bundle
 - WIP


yairhaimo:
@felixmosh I understand, thanks.
@ranyitz I didnt feel comfortable to put the tree shaking test under the `'[ts,json,html,haml]-loaders with yoshi.separateCss: false'` suite.
On the other hand, i didnt feel comfortable to run a yoshi build just for one small test.
What do you say?

felixmosh:
You solution would be great, to enforce minimum of `es2015` :].

I'm not sure that 2.9 will allow `import()` as part of `es2015` config.

ronami:
Can we change `typescript` to be a dependency instead of a peerDependency and bring in version `2.9.0`? That way we can enforce `esnext` and let all users use dynamic import statements. Also, less configuration.

yairhaimo:
Closing this as we will have this change as part of Yoshi 3's typescript changes

PR #297, opened 2018-05-19
move to using npm-ci-scripts

shahata:
### 🔦 Summary
...

### 🗺️ Test Plan
...


ronami:
I think we should have our own custom publish script (at least for now) so we can fine tune it before using the default one. We might want to make changes to our publish script and I think it will be hard to make large scale changes for eveyone whenever we do.

Also, we are expecting CI too have it installed globally (might be confusing).

Wdyt?

shahata:
1) I strongly disagree, we should have standard solution for building monorepos and our projects should be the example, I don't see why you would need something that all other monorepos don't need as well
2) we don't expect it to be available globally, I don't know what you mean

ranyitz:
@shahata We want to test and refine it before we use it for a wider range. We all want to find solutions that will suit all teams in Wix, and for this specific use-case, for all monorepos.

shahata:
ping

yairhaimo:
Dont we want Yoshi to be open source and `npm-ci-scripts` is a private Wix repo/package?

shahata:
we can make npm-ci-scripts public

shahata:
npm-ci-scripts is public now

PR #298, opened 2018-05-19
support es transpilation also for typescript

shahata:
### 🔦 Summary
...

### 🗺️ Test Plan
...


docs/faq/EXPORT-MODULES.md
```diff
@@ -17,7 +17,8 @@ Yoshi will also use this field to infer whether to create `/es` directory with n
 1. Don't include `babel-plugin-transform-es2015-modules-commonjs` to your `.babelrc`.
     * If you are using `babel-preset-env`, `babel-preset-es2015` or other preset which includes this plugin under the hood, use `{ modules: false }` to configure it.
     * If you are using `babel-preset-wix` there is nothing special you need to do in babel config.
-2. Specify path to your entry file with `module: 'dist/es/src/entry.js'`. Please note that Yoshi will create `es` directory with untranspiled modules near your usual transformation output (`dist/src` and `dist/es/src`).
+2. For typescript project, use `{"module": "es6"}` in `tsconfig.json` (under `compilerOptions`).
```
ronami: I think if the user should only configure his `package.json` with `{"module": "dist/es/src/entry.js"}`. If that happens, we can transpile a 2nd time into `./dist/es` and enforce `{"module": "es6"}`. Wdyt?

packages/yoshi/test/build.spec.js
```diff
@@ -629,6 +630,55 @@ describe('Aggregator: Build', () => {
...
+    });
+    after(() => {
+      test.teardown();
+    });
+
+    it('should build w/o errors', () => {
+      expect(resp.code).to.equal(0);
+    });
+
+    it('should not transpile modules for `/es` content if `module` field in `package.json` was specified', () => {
```
ronami: I think that should be `should` instead of `should not`, I'll fix after merging.

docs/faq/EXPORT-MODULES.md
shahata: it doesn't really matter imo, we should simply do proper zero config and then user doesn't need to do anything

PR #325, opened 2018-06-03
add karma stylable support

ranyitz:
#### Should add stylable support, so component libraries like `wix-ui` could use krama to test their components.

* Added `stylable-webpack-plugin` to the specs bundle.
* Karma now understands `test/karma-setup.js` file when exists and add it as the first entry point.
* Mocha reporter defaults to `html`, should not break any current behavior but adds the tests results on the debug tab in karma.

### 🗺️ Test Plan
Verify that when you require an `st.css` file, the contents will be in the result bundle. Also the `test/karma-setup.js` contents should be there when the file is in the file system.


packages/yoshi/config/webpack.config.specs.js
```diff
@@ -1,15 +1,24 @@
...
 const glob = require('glob');
+const StylableWebpackPlugin = require('stylable-webpack-plugin');
 const webpackConfigCommon = require('./webpack.config.common');
 const mergeByConcat = require('../src/utils').mergeByConcat;
 const { cssModules, tpaStyle } = require('./project');
 const globs = require('../src/globs');
 const projectConfig = require('./project');
 
 const specsGlob = projectConfig.specs.browser() || globs.specs();
+const karmaSetupPath = path.join(process.cwd(), 'test', `karma-setup.js`);
```
ronami: let's use `require.resolve()` instead?

PR #329, opened 2018-06-04
yoshi - patch stylable to always be part of the app's js bundle

shlomitc:
### 🔦 Summary
`stylable` generates it's own css bundle and not plugging into the app's css.
In order to reduce confusion, this PR puts stylable inside the JS bundle by default.
This change can be reverted after `stylable` will concat / hook to the app's css.

### 🗺️ Test Plan
In god we trust :)


ranyitz:
@shlomitc `separateCss` is an option that affects also `sass`/`less`/`css`, not only `stylable`, it will change the behaviour for all the users that does not configured `separateCss: false` even if they don't use stylable/wsr at all.

You can configure the stylable plugin directly here:
https://github.com/wix/yoshi/blob/9715d4e8cc4d21b3812729d570da2b16a05ae2ae/packages/yoshi/config/webpack.config.client.js#L102-L107

Same as it configured here:
https://github.com/wix/yoshi/blob/9715d4e8cc4d21b3812729d570da2b16a05ae2ae/packages/yoshi/config/webpack.config.specs.js#L34-L38

shlomitc:
CI now passes, but I have some concern. I was pretty sure stylable will add only one bundled css file, but now I see that there are tons of style tags for each component/file:
image

@barak007 - is it because I'm in development and not production mode? 
I'm also using an `index.st.css` file which looks like doesn't do the tree shaking (I see other components which I don't even use)

shlomitc:
when separating the stylable css into css bundle, I do get tree shaking working correctly, although there are a lot of white spaces (and not minified)

image

PR #347, opened 2018-06-07
Typescript tsconfig conf, tree shaking, dynamic import

yairhaimo:
### 🔦 Summary
This PR creates a new tsconfig for yoshi as a separate package: `tsconfig-yoshi`.
Also, force webpack Typescript loader to use `module: esnext` in order to support tree shaking for user code and dynamic imports.

### 🗺️ Test Plan
Create multiple exports, import only part of them, see that the bundle doesnt contain those that were not imported.


yairhaimo:
@ranyitz @ronami - do we want to force `module: esnext` in code as it is right now or do we want to have `tsconfig.server.json` and `tsconfig.client.json` and use those for better transparency of what is being done behind the scenes?

Issue #304, opened 2018-05-23
Ability to run yoshi test and yoshi start at the same time


iaroslavshvets:
# 🙋 Feature Request

### 😯 Current Behavior

Currently we got `Error: listen EADDRINUSE 127.0.0.1:3200` while starting CDN. For example if we ran yoshi start & have prepush/precommit git hook, which runs tests

### 🤔 Expected Behavior

No error.

### 💁 Possible Solution 

1) Detect that yoshi's CDN is already running and not start another instance.
2) Setup CDN on a different empty port.

ranyitz:
As yoshi controls the "CDN" server (which runs on `127.0.0.1:3200`), we can check if the port is already caught by another "CDN" and instead of exit with an error, just continue gracefully.

PRs welcome!

PR #349, opened 2018-06-08
use the next available port when cdn port is taken (#304)

ChenRoth:
### 🔦 Summary
implement @ranyitz's suggestion in [issue #304](https://github.com/wix/yoshi/issues/304)

### 🗺️ Test Plan
* make the intended CDN port unavailable
* try to spin up a CDN server at that port
* assert the CDN is served at another port

### DIFF
--- a/packages/yoshi/src/tasks/cdn/index.js
+++ b/packages/yoshi/src/tasks/cdn/index.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const path = require('path');
 const https = require('https');
+const detect = require('detect-port');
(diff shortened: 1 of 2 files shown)


ranyitz:
@ChenRoth Thanks for sending a PR 👍 

The intention was slightly different, if there are two "cdn" servers which are the same, we would prefer to continue and not to throw an error. Running the server on an empty port is not necessary, no one configured to listen to it.

Also, you'd like to throw an error if the port is taken by a different project.

I think that someone has already started looking into that issue.

cc @dotansimha

PR #357, opened 2018-06-11
Add "hmr disabled" option

yairhaimo:
### 🔦 Summary
Even with `hmr: false` we use the `webpack-hot-client` for live reloading.
This breaks in certain cases such as web workers.
This PR adds the `hmr: "disabled"` option which skips using `webpack-hot-client` altogether.
I have left this option undocumented for now until we decide if this is a long term solution.

### 🗺️ Test Plan
Run start with `hmr: "disabled"` and test that the bundle doesnt have any HMR code in it.


ranyitz:
Closed on #358 

Added an option to remove `hot-reload`

PR #380, opened 2018-06-20
Support passing args to test runners

### LINKED ISSUE #321: Allow passing cli arguments to jest, mocha, ...
# 🙋 Feature Request

### 😯 Current Behavior

when running `yoshi test --jest`, I can not pass argument to jest.

I would like to pass a name of a specific spec file that I want to run.

### 💁 Possible Solution 

We could have something like npm does:
`yoshi test --jest -- my_spec_path.js`

So arguments after the `--` will go to jest.

I guess that if I am running an npm script that runs yoshi, I could do:
If I have in my package.json script:
```
{
  scripts: {
    "test:unit": "yoshi test --jest"
}
```
then I could run from cli:
`npm run test:unit -- -- my_spec_path.js`

Here the first `--` passes args to yoshi, and the second `--` passes args to jest.

### 🔦 Context 

My use-case is for running a specific spec file, 
I can then write a launch config for VS code that debugs the currently opened file's (editor) test.

yairhaimo:
### 🔦 Summary
Support passing cli args to test runners like this:
```
yoshi test --jest -- arg1 --arg2=2 --arg3
```

Closes #321 

### 🗺️ Test Plan
Test dynamic arguments for all test runners: `jest`, `mocha`, `jasmine`, `karma`, `protractor`


yairhaimo:
Closing until we decide that we really want this feature

PR #381, opened 2018-06-20
Reduce npm install time

ronami:
### 🔦 Summary

This PR moves `karma` (and co) and `node-sass` to `devDependencies`, and bumps several deps to prevent multiple versions of the same package from being installed.


ranyitz:
**So what's breaking for karma users?**

Yoshi used to install all of karma plugins and load them to the karma config, after this change, users need to add and specify the plugins they needs.

For example, a project that uses `mocha` and test on `Chrome`:

```js
// karma.conf.js
module.exports = {
  browsers: ['Chrome'],
  frameworks: ['mocha'],
  plugins: [
    require('karma-mocha'),	
    require('karma-chrome-launcher'),
  ],
};
```

A project that uses `jasmine` and test on `PhantomJS`
```js
// karma.conf.js

module.exports = {
  browsers: ['PhantomJS'],
  frameworks: ['jasmine'],
  plugins: [
    require('karma-jasmine'),
    require('karma-phantomjs-launcher'),	
  ],
};
```

Issue #305, opened 2018-05-23
Add info about started dev server to console on yoshi start


bodia-uz:
# 🙋 Feature Request

### 😯 Current Behavior

Currently, after `yoshi start` command, there is no info in the console about started dev server.

### 🤔 Expected Behavior

Info like `dev server started at localhost:3200` could be useful.

### 💁 Possible Solution 

Add dev server onfo to "weback-dev-server" log

### 🔦 Context 

...

ranyitz:
@bodia-uz We are going to revamp the whole start experience and a message like the one you suggest here is a great idea! In the meanwhile I would be happy to accept a PR that adds it.

ChenRoth:
what about this? https://github.com/wix/yoshi/blob/master/packages/yoshi/src/tasks/app-server/index.js#L55

ranyitz:
@ChenRoth I think that @bodia-uz means to the `webpack-dev-server` that serves the bundle. While this log represents the `app-server`.

bodia-uz:
Yes. It's about `webpack-dev-server`.

dotansimha:
Done in https://github.com/wix/yoshi/pull/391

PR #391, opened 2018-06-24
Allow `npm test` and `npm start` run on the same time

dotansimha:
### 🔦 Summary

This PR include changes to allow developers to run `npm start` and `npm test` on the same time. 
It checks if another process is using the CDN port, and verifies that the process is running for the same directory of the project, in order to make sure it's not running in another project.

Fixes https://github.com/wix/yoshi/issues/304 and also https://github.com/wix/yoshi/issues/305


packages/yoshi/src/tasks/cdn/index.js
```diff
@@ -22,7 +22,28 @@ module.exports = ({
   configuredEntry,
   defaultEntry,
 } = {}) => {
-  return new Promise((resolve, reject) => {
+  return new Promise(async (resolve, reject) => {
```
ronami: I don't think we need a promise here. Instead, we can declare the top-level function as `async` and just throw an error or return instead of resolving/rejecting.

packages/yoshi/src/utils.js
```diff
@@ -184,3 +186,38 @@ module.exports.getListOfEntries = entry => {
 module.exports.shouldTransformHMRRuntime = () => {
   return project.hmr() === 'auto' && project.isReactProject();
 };
+
+function getProcessIdOnPort(port) {
+  return childProcess
+    .execSync('lsof -i:' + port + ' -P -t -sTCP:LISTEN', { encoding: 'utf-8' })
```
ronami: template strings?

packages/yoshi/src/utils.js
```diff
@@ -184,3 +186,38 @@ module.exports.getListOfEntries = entry => {
...
+    .split('\n')[0]
+    .trim();
+}
+
+function getDirectoryOfProcessById(processId) {
+  return childProcess
+    .execSync('lsof -p ' + processId + " | grep cwd | awk '{print $9}'", {
```
ronami: template strings?

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,112 @@ const expect = require('chai').expect;
...
+            port,
+          },
+        },
+      }),
+    });
+
+    it('should throw an error when CDN port is in use by another directory', function(done) {
+      const TEST_PORT = 3335;
+      server = http.createServer();
+      server.listen(TEST_PORT, 'localhost', () => {
```
ronami: I think it will be easier to understand if this we had a function to take a port. I looked a bit an it looks like we already have one: https://github.com/wix/yoshi/blob/master/packages/yoshi/test/start.spec.js#L758, can we re-use it here?

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,112 @@ const expect = require('chai').expect;
...
+        ).to.equal(true);
+        done();
+      });
+    });
+
+    it('should skip cdn startup when yoshi is already running in the same path', async function() {
+      const TEST_PORT = 3336;
+      test.setup(executionOptions(TEST_PORT)).verbose();
+      const testPath = test.tmp;
+      const toExecute = `
```
ronami: I think having a function that takes a port in a specified path will be nice here, wdyt?

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,112 @@ const expect = require('chai').expect;
...
+              process.exit();
+            });
+          });
+        `;
+
+      fs.writeFileSync(path.join(testPath, 'use-port.js'), toExecute, {
+        encoding: 'utf-8',
+      });
+      child = childProcess.exec('node use-port.js', { cwd: testPath });
+      await wait(500);
```
ronami: why do we need to wait here? can we use `await execa`/`execSync` instead?

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,112 @@ const expect = require('chai').expect;
...
+          .setup(executionOptions(TEST_PORT))
+          .verbose()
+          .execute('test', undefined, outsideTeamCity);
+
+        expect(res.code).to.equal(1);
+        expect(
+          res.stderr.includes(
+            `port ${TEST_PORT} is already in use by another process`,
+          ),
+        ).to.equal(true);
```
ronami: I think we can make this a bit more verbose by: `expect(res.stderr).to.include(...)`

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,112 @@ const expect = require('chai').expect;
...
+        encoding: 'utf-8',
+      });
+      child = childProcess.exec('node use-port.js', { cwd: testPath });
+      await wait(500);
+      const res = test.execute('test', undefined, outsideTeamCity);
+
+      expect(res.code).to.equal(0);
+      expect(
+        res.stdout.includes(`cdn is already running on ${TEST_PORT}, skipping`),
+      ).to.equal(true);
```
ronami: I think we can make this a bit more verbose by: `expect(res.stdout).to.include(...)`

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,112 @@ const expect = require('chai').expect;
...
+      }),
+    });
+
+    it('should throw an error when CDN port is in use by another directory', function(done) {
+      const TEST_PORT = 3335;
+      server = http.createServer();
+      server.listen(TEST_PORT, 'localhost', () => {
+        const res = test
+          .setup(executionOptions(TEST_PORT))
+          .verbose()
```
ronami: let's remove `.verbse()`

ronami:
Hey @dotansimha, that looks awesome 🏆, please see the few comments.

packages/yoshi/test/test.spec.js
dotansimha: Because it takes about 100-200ms for the script to actually start

packages/yoshi/src/utils.js
dotansimha: fixed

packages/yoshi/src/utils.js
dotansimha: fixed

packages/yoshi/test/test.spec.js
dotansimha: fixed

packages/yoshi/test/test.spec.js
dotansimha: fixed

packages/yoshi/test/test.spec.js
dotansimha: fixed

packages/yoshi/test/test.spec.js
dotansimha: done :)

packages/yoshi/test/test.spec.js
dotansimha: done :)

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,94 @@ const expect = require('chai').expect;
...
+      'package.json': fx.packageJson({
+        servers: {
+          cdn: {
+            port,
+          },
+        },
+      }),
+    });
+
+    it('should throw an error when CDN port is in use by another directory', function(done) {
```
ronami: I think removing `done` and replacing it with an async function would look a bit better

packages/yoshi/test/test.spec.js
```diff
@@ -2,12 +2,94 @@ const expect = require('chai').expect;
...
+        done();
+      });
+    });
+
+    it('should skip cdn startup when yoshi is already running in the same path', async function() {
+      const TEST_PORT = 3336;
+      test.setup(executionOptions(TEST_PORT));
+      const testPath = test.tmp;
+      child = takePortFromAnotherProcess(testPath, TEST_PORT);
+      await wait(500);
```
ronami: alright, let's fork the process and send back when the server is up instead of waiting. It's not a lot of work and I'm not sure if waiting a set amount of time can cause this test to be flaky.

PR #395, opened 2018-06-25
Mocha shouldnt listen to the dist folder

yairhaimo:
### 🔦 Summary
Mocha listened to all files, including those in the dist folder, causing it to run the tests twice (once for the original files, once for the transpiled version).
Mocha's watch method is very limited as it doesnt let you configure the ignored directories.
I replaced mocha's watch method with our chokidar based watch method.

### 🗺️ Test Plan
Tested manually:
- Change files in the `src/test/root` folders, see that the changes are detected and the tests run once
- Change files in the `dist` folder, see that the change is ignored and the tests dont run


packages/yoshi/src/commands/test.js
```diff
@@ -79,8 +79,13 @@ module.exports = runner.command(
       }
 
       if (shouldWatch) {
-        mochaArgs.push('--watch');
-        mochaArgs.push('--watch-extensions=js,jsx,ts,tsx');
+        watch({ pattern: [globs.testFilesWatch()] }, () => {
+          try {
+            execa('node', mochaArgs, { stdio: 'inherit' });
```
ronami: I think `execa` returns a promise so if it fails it won't get to the `catch` clause. We can prefix it with `await` to fix it.

packages/yoshi/src/commands/test.js
```diff
@@ -79,8 +79,13 @@ module.exports = runner.command(
       }
 
       if (shouldWatch) {
-        mochaArgs.push('--watch');
-        mochaArgs.push('--watch-extensions=js,jsx,ts,tsx');
+        watch({ pattern: [globs.testFilesWatch()] }, () => {
```
ronami: The previous glob we watched for changes:  https://github.com/wix/yoshi/pull/264/files#diff-b567138ee353bf6e456f7bad8bdbc7aeL61. Should we add `globs.specs()`?

ronami:
Looks great @yairhaimo! 🏆, Please see the two comments.

packages/yoshi/src/commands/test.js
yairhaimo: `testFilesWatch` is a superset of `specs`

PR #411, opened 2018-07-02
add test linting from tslint/eslint base configs

ranyitz:
### 🔦 Summary
Add rules and globals for testing to `eslint-config-yoshi-base` & `tslint-config-yoshi-base`.

Use `eslint-plugin-jest` as it catches the cases of jasmine and mocha.


packages/eslint-config-yoshi-base/index.js
```diff
@@ -1,5 +1,7 @@
+const path = require('path');
+
 module.exports = {
-  extends: 'prettier',
+  extends: ['prettier', path.resolve(__dirname, 'test.js')],
```
ronami: require.resolve()?

PR #424, opened 2018-07-08
Do not transpile server code/restart app server for client changes

yairhaimo:
### 🔦 Summary
This PR lets you develop client code without restarting the server for every client code change.

`yoshi start` will transpile server code (`babel`) and restart the app server for all changed files **except** for files under `src/client` which are handled by webpack's `babel-loader` and do not warrant a server restart.

Unlikely Breaking Change - This PR breaks the watch behavior **only** for projects that have a `src/client` folder and are counting on an app server restart for changes under that directory.

Typescript was not touched and is still handled by the `typescript`'s task `watch` mode.


yairhaimo:
Closing. Can reopen if relevant

PR #426, opened 2018-07-08
[WIP] create yoshi app

ranyitz:
### 🔦 Summary
This PR moves the creation of new yoshi based projects into `yoshi monorepo`.

Now these projects are in [wix-js-stack](https://github.com/wix-private/wix-js-stack/tree/master/generators) repo.

We've realized that it would be easier to maintain these projects along with yoshi itself. We would also like to benifit from the feedback of yoshi runs on different projects without breaking.

We want a fast feedback loop, and for that we've created another build configuration (`e2e`) that runs in parallel with the current one.

## What's inside?

* `create-yoshi-app` - create new yoshi powered projects from templates.
* `e2e` - create each one of the projects build and test them with yoshi.

Currently supports:
* [x] - client
* [x] - client-typescript
* [x] - fullstack
* [x] - fullstack-typescript


CONTRIBUTING.md
```diff
@@ -58,6 +59,26 @@ VERBOSE_TESTS=true npm test
 ```
 This is the same as adding the `.verbose()` method to each and every test.
 
+## Running e2e Tests
+The e2e tests create each project type from the `projects/create-yoshi-app/templates` directory in a temporary directory, it will then run `npm install` & `npm test` on each project to verify it is still working.
```
ronami: I think this can be phrased a bit better, how about: "Each E2E test creates a corresponding project type from ..."?

packages/create-yoshi-app/src/generateProject.js
```diff
@@ -0,0 +1,44 @@
...
+    organization,
+  };
+  const files = {};
+
+  filesPaths.forEach(filePath => {
+    const content = fs.readFileSync(path.join(templatePath, filePath), 'utf-8');
+    files[filePath] = replaceTemplates(content, valuesMap);
+  });
+
+  for (const fileName in files) {
```
ronami: would it make sense to run just one `forEach` instead of running a 2nd `for(...)` and keeping some state?

packages/create-yoshi-app/src/getQuestions.js
```diff
@@ -0,0 +1,54 @@
+const fs = require('fs');
+const path = require('path');
+const getGitConfig = require('git-config');
+
+module.exports = async workingDir => {
+  const projectTypes = fs
```
ronami: I'm okay with maintaining a json file that would keep what project types we support. Creating a new project type would mean adding it there. I think I like it being explicit. What do you think?

PR #447, opened 2018-07-17
Improved publish script

ranyitz:
### 🔦 Summary
This PR adds the ability for the custom publish script to work with lerna monorepos.

Big thanks to @bchmn for implementing it on [rich-content](https://github.com/wix-incubator/rich-content/blob/master/scripts/customPublish.js)

This is the algorithm:

1. load all packages using `lerna.json`
2. If the package is private, skip publish
3. If the package already exists on the registry, skip publish.
4. choose a dist-tag ->
   * `old` for a release that is less than latest (semver).
   * `next` for a prerelease (beta/alpha/rc).
   * `latest` as default.
5. perform npm publish using the chosen tag.

Also, it will arrange the output in a nice pockety way:

screen shot 2018-07-17 at 17 25 07


shahata:
1. You could just do `lerna exec --concurrency 1 -- node ../../scripts/customPublish` instead of `lerna exec --concurrency 1 -- npm run customPublish`. No `customPublish` needed.

2. Making `console.log`'s clearer is obviously a welcome change, the thing that disturbs me is that we added knowledge about lerna to this script instead of simply moving to use `npm-ci-scripts` and improving the `console.log`'s there.

shahata:
BTW, I think that the fact that you replaced `glob` with `globby` to save a few characters of code and at the same time replaced `lerna exec` with your own custom implementation is pure irony :)

ranyitz:
I see your point

PR #448, opened 2018-07-18
improve user experience for yoshi initiator

ranyitz:
### 🔦 Summary
This PR should make `create-yoshi-app` fun.
Inspired by [create-react-app](https://github.com/facebook/create-react-app) to give some more details and better developer experience.

* Enable running in an empty directory or create the directory from outside
(`npx create-yoshi-app my-app`)
* Require to work with wix private registry (it won't work otherwise)
* Handle cancel event while in the prompts phase.
* Run `npm install` for the project
* Run `npm lint --fix` for prettier fixes after inserting `js` code to the templates.
* Add relevant output
* Omit the question for the project name, as it derived from the directory name.

screen shot 2018-07-18 at 18 18 31

screen shot 2018-07-18 at 18 21 51


packages/create-yoshi-app/bin/create-yoshi-app.js
```diff
@@ -4,21 +4,140 @@ process.on('unhandledRejection', error => {
...
+    process.exit(1);
+  }
 
-clearConsole();
+  const registry = getRegistry(workingDir);
 
-runPrompts(workingDir).then(results => generateProject(results, workingDir));
+  if (!registry.includes(privateRegistry)) {
+    console.log('You should be authenticated to wix private registry');
+    console.log('Run the following command and try again:\n');
```
yairhaimo: `You should be authenticated to Wix's private registry`

PR #466, opened 2018-07-27
feature: add hadron to fullstack templates

4alexandr:
### 🔦 Summary
Added Hadron support to fullstack and fullstack-typescript templates.
It will allow to use different artifacts versions without restarting the application.

### 🗺️ Test Plan
* run `npm test`
* run `npm start`

### Changes
* CHANGED `testkit.app` to `testkit.server`, to fully reload application
* CHANGED `autogenerated-fullstack[-ts]` name to use `{%projectName%}`
* CHANGED `./test/it/server.spec.js` to `./test/e2e/server.e2e.js`, cause bringing environment up and down is very slow
* REMOVED `./test/templates/autogenerated-fullstack[-ts].json`, cause it's only the duplication of original `erb` template

### Obstacles 
* Docker containers are left running in the background when a developer `npm start`s. and afterwards just quits the shell `CTRL + C`. I tried listening for the process events, but seems like `yoshi` as the master process is intercepting us and we cannot do any cleanup during the exit.
* Even with `SIGHUP` event the environment restart takes about `10s`, which is very slow.
* Yoshi prints to the output that server is ready, but it's not the truth, it still setups and brings kafka in the background.

cc: @roymiloh


packages/create-yoshi-app/templates/fullstack-typescript/index.js
```diff
@@ -1,9 +1,20 @@
...
+const greynode = require('wix-bootstrap-greynode');
+const hadron = require('wix-bootstrap-hadron');
 
 const rootDir = process.env.SRC_PATH || './dist/src';
 const getPath = filename => path.join(rootDir, filename);
 
+// list the artifacts you want hadron to manage for you
+const hadronOpts = {
+  staticArtifacts: [
+    {artifactId: 'com.wixpress.autogenerated-fullstack-ts'},
```
mastertheblaster: IMHO this should be the ID of the autogenerated project itself.

packages/create-yoshi-app/templates/fullstack-typescript/src/server.ts
```diff
@@ -1,36 +1,51 @@
 import { Router } from 'express';
 import * as wixRunMode from 'wix-run-mode';
 import * as ejs from 'ejs';
+import * as path from 'path';
 import * as wixExpressCsrf from 'wix-express-csrf';
 import * as wixExpressRequireHttps from 'wix-express-require-https';
-import { readFileSync } from 'fs';
+
+const artifactId = 'com.wixpress.autogenerated-fullstack-ts';
```
mastertheblaster: A note for myself need to check, but AFAIK it should equal to one user specified during the bootstrap.

packages/create-yoshi-app/templates/fullstack-typescript/test/e2e/app.e2e.ts
```diff
@@ -1,9 +1,9 @@
 import { expect } from 'chai';
-import { beforeAndAfter, app } from './../environment';
+import { startTestApp } from './../environment';
 import './e2e-common';
 
 describe('React application', () => {
-  beforeAndAfter();
+  const app = startTestApp();
```
mastertheblaster: `start...` refers that this method will return a promise. And actually it's an object.

roymiloh:
Closing in favor of https://github.com/wix/yoshi/pull/479.

PR #477, opened 2018-08-03
Add source-map-loader

yuriibut:
### 🔦 Summary
Include source-map-loader to yoshi

### 🗺️ Test Plan
...

### DIFF
--- a/packages/yoshi/package.json
+++ b/packages/yoshi/package.json
@@ -139,6 +139,7 @@
     "screenshot-reporter": "~1.1.3",
     "shelljs": "^0.8.2",
     "shmock": "~0.8.1",
+    "source-map-loader": "^0.2.3",
     "strip-ansi": "^4.0.0",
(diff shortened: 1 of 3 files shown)


packages/yoshi/package.json
```diff
@@ -139,6 +139,7 @@
     "screenshot-reporter": "~1.1.3",
     "shelljs": "^0.8.2",
     "shmock": "~0.8.1",
+    "source-map-loader": "^0.2.3",
```
sdoomz: I think it needs to be in `dependencies` and not in `devDeps`

sdoomz:
tests you can add here https://github.com/wix/yoshi/blob/9ffd8c083e62cf7a07df15355c4c3d4f353e0a7e/packages/yoshi/test/loaders.spec.js#L27

ranyitz:
@yuriibut Do you still need this?

PR #483, opened 2018-08-06
yoshi: use relevant externalize-realtive-module-loader

roymiloh:
I dropped the scope from the module name. Fixes https://github.com/wix/yoshi/issues/481.


roymiloh:
Because new versions are getting published under this name and @barak007 made a [change](https://github.com/wix/externalize-relative-module-loader/commit/554a05b4ea42842d2095e3f769e09917d5f132df) that adds Windows support.

PR #486, opened 2018-08-09
Upgrade stylable to new scoped package

tomrav:
Upgrade Stylable to use latest `@stylable` scoped packages.


lbelinsk:
I will be great to have this in yoshi2 as well.
branch `version_2.x`

tomrav:
This is not actually stale, starting the process with Yoshi v2.

lbelinsk:
I found a ssr project that uses yoshi 3, so the question is whether we want to release an alpha version of yoshi and ask them to use it to make sure it all works fine, or should we just merge and quickly test that it all works fine?

https://github.com/wix-private/devcenter/blob/master/dev-center-3/package.json#L83

PR #493, opened 2018-08-12
Add Stylable CLI dependency

tomrav:
Used to be a part of `stylable-integration` extracted to separate package in `@stylable/cli`


tomrav:
@ronami can we merge this and release an RC to test with? (starting with v2)

PR #502, opened 2018-08-16
Add option umdNamedDefine. It is true by default

bohdan-lytvynov:
### 🔦 Summary
Option umdNamedDefine is hardcoded as `true` right now so all modules at the output have names. Some module loaders, for example, SystemJS doesn't support loading of more than one named module [code here](https://github.com/systemjs/systemjs/blob/5ce46c325430d07ecac925441b6f04061ca258a9/src/format-helpers.js#L142). It would be nice to opt out named AMD modules.


ronami:
I'm not sure why we're setting `umdNamedDefine` to `true` by default. Optimally I would like to understand if someone needs it to be `true`, and if not use the default (`false`). Until then, making it configurable sounds good.

Can you add a test for when it's configured to `false`?

bohdan-lytvynov:
@ronami just added a test.

PR #507, opened 2018-08-20
yoshi test: forward cli args after -- to jest cli

amitdahan:
### 🔦 Summary
Forward anything after -- when using `yoshi test --jest` to jest's CLI.
_Fixes #321_

### 🗺️ Test Plan
Added a test case for 2 custom jest patterns (a subset of the default tested suites).


packages/yoshi/src/commands/test.js
```diff
@@ -162,12 +162,14 @@ module.exports = runner.command(
...
         `--config=${configPath}`,
         `--rootDir=${process.cwd()}`,
-        shouldWatch ? '--watch' : '',
-        cliArgs.coverage ? '--coverage' : '',
-        cliArgs.runInBand ? '--runInBand' : '',
-        cliArgs.forceExit ? '--forceExit' : '',
       ];
 
+      shouldWatch && jestCliOptions.push('--watch');
+      cliArgs.coverage && jestCliOptions.push('--coverage');
```
ranyitz: We don't need those anymore as these are equal to jest's options.

ranyitz:
I've opened #618 as a continue of this one

PR #518, opened 2018-08-22
open browser on app-server (re)started

### LINKED ISSUE #515: Open/Refresh browser window when application is available
# 🙋 Feature Request

### 😯 Current Behavior

When `yoshi start` finishes we tell the user that the application is available at a certain port.

### 🤔 Expected Behavior

In addition to telling the user the application is available, we should also automatically open the browser to that url.
If a browser window is already open to that url we should focus on the browser and refresh the window and not open a new window.

netanelgilad:
### 🔦 Summary
Fixes #515. Opens/refreshes the browser tab for the local app server on (re)start.
There are a few points to discuss on this initial implementation:
* This uses `openBrowser` for `react-dev-utils`. `react-dev-utils` contains a lot other utils that are not relevant, so this increases the size of the `yoshi` module, but since this is not that important for `yoshi` I thought it would be better to work with the module and get bug fixes for free instead of copy-pasting the code. Also gives the benefit of supporting the `BROWSER` env variable, same usage as in CRA.
* when app-server starts/restarts, it doesn't know when the cdn is ready. That means that it is possible to open the browser before the cdn is ready, in that case there is no HMR connection ready, and the user will need to manually refresh the browser window once. I've added a setTimeout, but I'm not super happy about that solution, as it's flaky. Would love suggestions on how to better implement this.

### 🗺️ Test Plan
Not sure how to add tests for this, and couldn't find tests for this in CRA. I could create an elaborative e2e test that checks if the correct process that opens the browser was spawned, but pragmatically, not sure it's worth the time.


netanelgilad:
Yep, Both the printing of the server ready and opening the browser should be done after the server is ready (IMO it should wait for both the app-server and the cdn to be ready).

I can fix this as part of this PR, or maybe we can fix #523 first and then update this PR with the fix.

ranyitz:
As discussed with @ronami and @netanelgilad Waiting for port would be the wanted behavior here.

Important to say though, that this would be a breaking change, as not all of our users use the `PORT` environment parameter to bootstrap their server, and for those users yoshi would wait for the port forever 😆

We've decided to merge this feature into branch `4.x` and that it'll be released as part of version 4

ranyitz:
@netanelgilad Let's open the PR to [version_4.x](https://github.com/wix/yoshi/tree/version_4.x) branch and close this one.

PR #535, opened 2018-08-26
validate yoshi config before running command

### LINKED ISSUE #500: Add field validation for Yoshi's config
# 🙋 Feature Request

### 😯 Current Behavior

If fields that are not supported are used there are no warnings. If fields are used with wrong types weird errors may be thrown.

### 🤔 Expected Behavior

By using [jest-validate](https://github.com/facebook/jest/tree/master/packages/jest-validate) we can have pretty nice errors for using wrong/unsupported types of configuration fields.

netanelgilad:
### 🔦 Summary
This PR adds configuration validation before running yoshi commands. Fixes #500. This initial implementation is using `ajv` to validate the configuration object against a JSON schema. Then the errors are printed using `YoshiOptionsValidationError` which is a copied implementation from the webpack-cli repository. (The default errors text from `ajv` are not great, so the webpack way seems a good way to go).

Please look carefully at the JSON schema. I created it based on the API docs, but maybe you'll know better. And if there are old deprecated fields, we can print deprecation warning for them, I just didn't know if there are any.

### 🗺️ Test Plan
So this part is still is open for discussion. What testing strategy do you see for this if at all? I can add unit tests for the copied webpack code for generating the error messages, but maybe it won't add too much, and a full e2e test may be an overkill. Thoughts?


ranyitz:
@netanelgilad Thanks, this is a great addition.

I think once all the configuration errors that we have in our tests are fixed we'll have enough confidence in the valid case 😆

We still need to test that the configuration validation is working in case we've used it improperly. (So something will fail in case we remove the validation) I would create 1/2 e2e tests for that.

I would unit test each case separately, but not necessarily in this PR (It shouldn't stop us from merging)

packages/yoshi/config/project.js
```diff
@@ -1,11 +1,14 @@
 const path = require('path');
 const _ = require('lodash');
+
```
ranyitz: There is an extra space here

netanelgilad:
Great, so I found that there quite a few configuration options that are not mentioned in the API doc, so I'll update the schema with those, and I'll also update the API doc (I'll need your help there, cause I don't really know what all those options are doing). These also should fix the existing tests.

netanelgilad:
> We still need to test that the configuration validation is working in case we've used it improperly. (So something will fail in case we remove the validation) I would create 1/2 e2e tests for that.

@ranyitz So I'm thinking about the e2e tests. I can create a test that makes sure that the validation runs for every cli command by creating a configuration with some generic error and asserting that the warnings appear in the stdout. But I won't check every possible configuration error cause that would be an overkill. Good?

netanelgilad:
Check out the new tests, and let me know how it looks. Ready for merge from my end.

docs/faq/YOSHI-API.md
```diff
@@ -106,6 +106,22 @@ You also use the `prod` keyword to only separate css on CI and production, this
...
+
+Set to true to build with TPA style.
+
+#### yoshi.enhancedTpaStyle ![status experimental](https://img.shields.io/badge/status-experimental-ff69b4.svg)
+
+Set to true to build with enhanced TPA style.
+
+#### yoshi.clientProjectName
+
+The name of the client project.
```
ranyitz: I don't know what this is, but maybe we should add that it is written in `nbuild.tar.gz.xml`.

__reference:__

https://github.com/wix/yoshi/blob/e7ac2320495526e5fb89404ca63db8deb441a1fe/packages/yoshi/src/tasks/maven-statics/index.js#L12-L15

Issue #491, opened 2018-08-12
Add a descriptive message in cases when the node version is below 8.7


ranyitz:
# 🙋 Feature Request

Throw a meaningful error when node version is lower than `8.7`.

### 😯 Current Behavior

When installed on a machine with `node < 8.7` `create-yoshi-app` is failing with the following error:

![image](https://user-images.githubusercontent.com/11733036/44001116-712a302c-9e34-11e8-9db5-3b43053215ff.png)

### 🤔 Expected Behavior

Check for now version, and throw a message in case node version is below 8.7

### 🔦 Context 

Some folks still use old node versions, `create-yoshi-app` is written and tested using `node > 8.7` feature. The simplest thing would be to throw a descriptive error.

netanelgilad:
How about using the [`engines`](https://docs.npmjs.com/files/package.json#engines) field on the `package.json`? Maybe not the most beautiful solution, but it would print a warning by `npm` or `yarn` about using a non-supported version of node.

ranyitz:
I think that you'll still see the following:
```
unexpected token )
```

which looks unprofessional.

`engineStrict` and other stuff that I've checked aren't supported anymore.

Manually checking the node version seems to be a pretty easy solution that does the trick.

PR #537, opened 2018-08-27
create-yoshi-app node version verification

yairhaimo:
### 🔦 Summary
`create-yoshi-app` requires a minimum node version of 8.7.
This PR will exit the `create-yoshi-app` process if the minimum node version is less than the defined minimum.

Closes https://github.com/wix/yoshi/issues/491

### 🗺️ Test Plan
Unit test with different mocked versions.

An e2e will be nice too but this PR does not contain it.


packages/create-yoshi-app/src/createApp.js
```diff
@@ -4,10 +4,12 @@ const verifyWorkingDirectory = require('./verifyWorkingDirectory');
 const runPrompt = require('./runPrompt');
 const generateProject = require('./generateProject');
 const verifyRegistry = require('./verifyRegistry');
+const verifyMinimumNodeVersion  = require('./verifyMinimumNodeVersion');
```
ranyitz: double space before `=`

packages/create-yoshi-app/src/createApp.js
```diff
@@ -4,10 +4,12 @@ const verifyWorkingDirectory = require('./verifyWorkingDirectory');
...
 
 module.exports = async (workingDir, projectDirName) => {
   verifyWorkingDirectory(workingDir);
   verifyRegistry(workingDir);
+  verifyMinimumNodeVersion(8, 7, 0);
```
ranyitz: should accept a string instead of numbers.

PR #538, opened 2018-08-27
wait for app-server port before finishing app-server task

### LINKED ISSUE #523: Wait for user server to be available before printing that its available
# 🙋 Feature Request

### 😯 Current Behavior

We print that the `Application is now available` before we even run the application server. 

### 🤔 Expected Behavior

Print the `Application is now available` message only after the server is really up and running.

### 💁 Possible Solution 

Listen to the server port and print the message only after you receive a response from it.
[wait-port](https://www.npmjs.com/package/wait-port) might be a viable solution.

### 🔦 Context 

Improve DX - If the user is quick enough and accesses the URL they will receive a "page not found" message.

netanelgilad:
### 🔦 Summary
Fixes #523. As discussed, this may be a breaking change for projects that ignore the PORT env variable, because we will wait forever on a port that is not going to be used. (Did any one say framework? 😜 ).

### 🗺️ Test Plan
Added a new test to the start command that makes sure that the message is printed only when the port is available.


packages/yoshi/src/tasks/app-server/index.js
```diff
@@ -97,6 +77,35 @@ function initializeServerStartDelegate({
...
       }
     });
+
+    console.log('');
+    console.log('Waiting for app-server to start...');
+
+    await waitPort({
+      host: 'localhost',
+      port: env.PORT,
+      output: 'silent',
```
ronami: I think we should add a (5-10s?) timeout, so if the server didn't start by that time we'll display an error: https://github.com/dwmkerr/wait-port#api. Wdyt?

packages/yoshi/test/start.spec.js
```diff
@@ -695,6 +710,40 @@ describe('Aggregator: Start', () => {
...
+
+          setTimeout(() => {
+            server.listen(port, hostname, () => {
+              console.log('Running a server...');
+            });
+          }, 1000);
+        `,
+          'package.json': fx.packageJson(),
+        })
+        .verbose()
```
ronami: we can remove the `.verbose()` call

packages/yoshi/test/start.spec.js
```diff
@@ -794,9 +843,9 @@ describe('Aggregator: Start', () => {
     );
   }
 
-  function checkStdout(str) {
+  function checkStdout(str, backoff = 100, max = 10) {
```
ronami: Wdyt about changing it to `function checkStdout(str, { backoff = 100, max = 10 }) { ...`? Just so its clearer what each value means if you're just looking at the function call.

packages/yoshi/test/start.spec.js
```diff
@@ -695,6 +710,40 @@ describe('Aggregator: Start', () => {
...
+      // Intentionally start listening after a timeout, to check that we indeed wait for the port
+      child = test
+        .setup({
+          'index.js': `
+          'use strict';
+
+          const http = require('http');
+
+          const hostname = 'localhost';
+          const port = ${port};
```
ronami: I think we can simplify it a bit: in our production code we wait for `process.env.PORT` but in the test we concatenate the port as a string. Can we just `server.listen()` to `process.env.PORT`?

packages/yoshi/src/tasks/app-server/index.js
```diff
@@ -97,6 +77,35 @@ function initializeServerStartDelegate({
         displayErrors();
       }
     });
+
+    console.log('');
+    console.log('Waiting for app-server to start...');
```
ranyitz: Just an idea, maybe we can show this message only if we've waited for a few seconds without the port to be detected. I don't think Users would like to see it printed every time, just when there is some pausing or a possible problem.

I'm not sure what the exact time should be, I would say 3 seconds, but it can be adjusted to give a better experience.

@netanelgilad @ronami What do you think?

packages/yoshi/test/start.spec.js
```diff
@@ -695,6 +709,42 @@ describe('Aggregator: Start', () => {
...
+          setTimeout(() => {
+            server.listen(port, hostname, () => {
+              console.log('Running a server...');
+            });
+          }, 1000);
+        `,
+          'package.json': fx.packageJson(),
+        })
+        .spawn('start', [], { PORT: port });
+
```
ranyitz: Maybe we can test that after `900ms` there is no `Application is now available` message printed or anything on the server port, so we'll know for sure that the wait for port feature is working.

I would try to remove these lines:

https://github.com/wix/yoshi/blob/6572b4696ae8cb04110b69638bc4a17726bd9cb1/packages/yoshi/src/tasks/app-server/index.js#L84-L88

and verify that the test fails.

netanelgilad: Yep definitely checked that :). The test works correctly because if you remove those lines the "Application is now available" gets logged right away and then it fails on the `fetch` call (cause the server is not up yet)

packages/yoshi/src/tasks/app-server/index.js
netanelgilad: yeah, makes since. I've added a more informative message after a timeout.

packages/yoshi/test/start.spec.js
ranyitz: Awesome! my bad

PR #543, opened 2018-08-28
Typescript strict checks

### LINKED ISSUE #530: Fix Typescript Types in Generators to Easily Support Strict Mode
# 🙋 Feature Request

### 😯 Current Behavior

When generating a new Typescript project and enabling `strict` mode in `tsconfig.json` there are several type errors.

### 🤔 Expected Behavior

Enabling `strict` should just work out of the box on newly generated projects.

netanelgilad:
### 🔦 Summary
Fixes #530. Fixes typescript strict errors across templates. Still waiting on a PR to the types of react-i18next to get rid of all errors. Do not merge yet.

### 🗺️ Test Plan
Added tests to make sure that generated typescript project pass strict check by typescript.


packages/create-yoshi-app/scripts/e2e.js
```diff
@@ -59,14 +59,23 @@ const testTemplate = mockedAnswers => {
...
+      prompts.inject(mockedAnswers);
       verbose && console.log(chalk.cyan(tempDir));
       await createApp(tempDir);
       await copyLocalModules(tempDir);
     });
 
+    if (mockedAnswers.transpiler === 'typescript') {
+      it('should not have errors on typescript strict check', () => {
+        console.log('checking strict typescript...');
+        execa.shellSync('./node_modules/.bin/tsc --noEmit --strict', {
```
ronami: I think we should have a lot more things to fix, can you run it with `-p tsconfig.json`? Also, looking at the `tsconfig.json` we have this option turned on: `"noImplicitAny": false`, which overrides the `--strict` flag. Removing it shows a lot more errors.

netanelgilad: That's weird, since from the typescript docs:
> Compiler options specified on the command line override those specified in the tsconfig.json file.

I thought it should pickup the `tsconfig.json` file and override it with the strict option. I'll check this out.

ronami: Passing `--noImplicitAny ` directly also shows more errors, but just with `--strict` it doesn't.

netanelgilad: So as we talked about, I've remove the last commit, so we will stay with `noImplicitAny` set to `false` to avoid having to type every dependency.

PR #549, opened 2018-08-29
Simplify server code and add basic annotations to it

ronami:
### 🔦 Summary

This PR simplifies the server code and adds some explanations on things that are not immediately clear.

I think that annotating confusing parts in the generated code can help in making it more approachable. In the end, this code is going to be part of the users' projects. Our users should understand the code that's inside their own projects.

This is something other popular boilerplates/generators do and I think that even small explanations can help.


packages/create-yoshi-app/templates/fullstack-typescript/src/server.ts
```diff
@@ -1,25 +1,40 @@
...
+  // Define a route to render our initial HTML
+  app.get('/', async (req, res) => {
+    // Extract some data from every incoming request
     const renderModel = getRenderModel(req);
-    const html = ejs.render(templateFile, renderModel, {
+
+    // Generate HTML as a string from our EJS template. See https://github.com/mde/ejs for more
+    // information on the EJS template engine.
+    const html = ejs.renderFile('./src/index.ejs', renderModel, {
       cache: isProduction,
```
ranyitz: @ronami I think that the build is failing because `renderFile` function accepts `cache` option only if `filename` is also supplied. [ejs docs](http://ejs.co/#docs)

PR #557, opened 2018-09-03
WIP: yoshi: use source property when looking for assets to copy

roymiloh:
For some reason `globby` doesn't work when project includes cyclic & symlinked deps, although pattern does match those dirs. _Somehow_, when using `source` as the base name (appended to `cwd` eventually), it's working.


ronami:
Solved in https://github.com/wix/haste/pull/197.

PR #558, opened 2018-09-03
create new packages yoshi-config and yoshi-helpers

ranyitz:
### 🔦 Summary
This PR should create two new packages

`yoshi-config`
`yoshi-helpers`

The motivation for this change is to provide a fast way to consume yoshi configuration (without consuming yoshi with all its dependencies). It will also help with creating clearer boundaries for our internal modules. Another bonus is that we can export these tests to a different phase and help reducing the time of the `yoshi` package tests. 

**usage**

#### `yoshi-config`
```js
// load, validate and return the yoshi config from `package.json` and `yoshi.config.js`
const projectConfig = require('yoshi-config');
``` 

```js
// returns yoshi's glob patterns
const globs = require('yoshi-config/globs');
``` 

```js
// load, validate and return the `jest-yoshi.config.js`
const jestYoshiConfig = require('yoshi-config/jest');
``` 

#### `yoshi-helpers`
```js
// basically everything that used to be in `yoshi/src/utils`
const yoshiHelpers = require('yoshi-helpers');
``` 

## Changes

* When we're running `test-configs` (on PR CI) we will run every other project but `yoshi`/`create-yoshi-app`
* Move `devDependencies` to yoshi monorepo root
* configure mocha for longer timeouts and to use mocha-env-reporter
* unify the way `jest-yoshi.config.js` and `yoshi.config.js` are being consumed (API'wise).
* replace the dependency in yoshi, with dependency with yoshi-config for `jest-environment-*`
* move the `globs` to be a part of `yoshi-config` and replace most functions with values.
* Change most of the function in `yoshi-config` to be values.
* There is a cyclic dependency between `yoshi-config` and `yoshi-helpers` because the `requireHooks` from the helpers in `yoshi-config/jest`, we should look into remove it in the future.
* Move `test-helpers` to be on the monorepo root as we need them also from `yoshi-config`. I know that it makes the usage in them harder (longer requires in yoshi tests). We can solve it in a different PR.
* Separate `yoshi-helpers` into 3 logical parts (queries, utils, require-hooks). all being exported through the `yoshi-helpers` module.
* Unify the `warnOnConfigValidationErrors` into the initial loading of the config.

## Afterthoughts

We may not need `yoshi-helpers`.

I've created it because I didn't want to bring all of yoshi to `yoshi-config` package just because the `require-hooks` are needed from `yoshi-config/jest`. And it makes sense to group those utils into one module.

But if we can solve the problem above without it, we could leave `yoshi-helpers` in yoshi and maybe export them in the future if needed.

@ronami @netanelgilad What do you think?


ranyitz:
BTW, the e2e build is failing because we copy `yoshi` (which has symlink to `yoshi-config` in its `node_modules`).

Yoshi does not install `yoshi-config` on the root becuase `yoshi-config` is not published yet and `yoshi` with a dependency for `yoshi-config` is something that exist only in this PR.

Maybe we can use something [like this](https://github.com/verdaccio/verdaccio/blob/master/docs/use-cases.md#override-public-packages) before our install. I'm not sure how long will it take, but it will be close to reality, and will prevent weird feedbacks like this one.

packages/yoshi/config/protractor.conf.js
```diff
@@ -35,7 +40,7 @@ const merged = ld.mergeWith(
           }).css,
       });
 
-      require('../src/require-hooks');
+      setupRequireHooks();
```
ronami: Does this work? I think that protractor might stringifies this function and can run it in a different process. I think this is why previously the `require()` call was used twice.

PR #582, opened 2018-09-20
process unprocessedModules with graphql loader

### LINKED ISSUE #581: Process graphql files of unprocessedModules
# 🐛 Bug Report

### 😯 Current Behavior

Importing graphql files from node_modules raise an exception during build time, for exapmle
```
 ERROR in ../node_modules/gql-queries/person.graphql 1:6
Module parse failed: Unexpected token (1:6)
You may need an appropriate loader to handle this file type.
> query getSomeDate () { person { id} }
```

### 🤔 Expected Behavior

No errors during build time, and able to use graphql files

### 🔦 Context 

We have a project that is split into few packages(all using Yoshi), let's call them components.
The main package import all these components, by importing the transpiled code, not the bundle(using `externalUnprocessedModules`).
Those components has graphql files and are loaded with Yoshi's graphql loader, and before publishing the component, we copy the graphql files to the transpiled files folder(dist/src) where they originally were.

When we import the component from the main project, it fails to load the grpahql files.

### 💁 Possible Solution 

include unprocessedModules in the graphql loader

### 👣 Steps to Reproduce 

Install a package with graphql files
Import them in a yoshi projects
Build

eddierl:
### 🔦 Summary
fixes #581 

### 🗺️ Test Plan
import graphql file from node_modules
build
expect to see it output


netanelgilad:
Looks good to me @eddierl. Thanks for the PR.

But a point for us internally: shouldn't we use `unproccessedModules` function for all our loaders? The name of the config is a bit confusing but a better name for it would be `modulesThatNeedToBeProccessed` (yeah went extreme to better express the intention, it doesn't need to be that name really). 

That's because this issue could rise for any type of loader, and if a user has opted-in to have a certain dependency proccessed then they probably know what they are doing. The other problem that rises is that once you opt-in to process a module, all the loaders will be used, and you can't select only a few (but that problem may never rise, so I don't think we need to solve it in advance).

Anyway, this discussion doesn't prevent merging the PR to get the current issue resolved.

eddierl:
Thanks @netanelgilad !

I totally agree about the confusion of the name.

Regrading aligning the list of loaders between source code and npm modules, I guess it also make sense (if the purpose of `unproccessedModules` is the let yoshi process npm modules as it's the source code)

I think it raise another problem, maybe it something bad in the our design... but currently, we do the copy of the graphql files while files like css/scss are copied/transpiled by Yoshi.
Maybe Yoshi should also copy the graphql files(along with other files that needs a loader)?

So can merge it?

PR #584, opened 2018-09-20
show a warning when the user tries to run e2e tests but there is no bundle built

ranyitz:
### 🔦 Summary
`e2e` tests requires a bundle to be built in `dist/statics` directory.
Sometimes there are cases when the user tries to run these tests and get an unexpected error from the tests.

We could try and predict those moments, give a better experience and spare some time. Let's start with showing a warning that could help detecting the problem.

### 🗺️ Test Plan
Verify that we encounter a warning for case of `e2e` test and no build in `dist/statics`.


packages/yoshi-helpers/queries.js
```diff
@@ -76,3 +76,8 @@ module.exports.hasE2ETests = () => {
 module.exports.hasProtractorConfigFile = () => {
   return exists(path.resolve('protractor.conf.js'));
 };
+
+module.exports.hasBundleInStaticsDir = () => {
+  const statics = glob.sync(path.resolve(globs.statics, '*'));
+  return statics.some(fileName => fileName.endsWith('bundle.js'));
```
ronami: can we filter the glob to test for `globs.statics/*.bundle.js` instead?

packages/yoshi/src/commands/test.js
```diff
@@ -53,6 +55,20 @@ module.exports = runner.command(
...
 
     function bootstrapCdn() {
+      if (!hasBundleInStaticsDir()) {
+        console.error();
+        console.error(
+          chalk.red(
+            ' ● Warning:\n\n' +
+              "   you are running e2e tests and doesn't have any bundle located in the statics directory\n" +
+              '   you probably need to run ' +
+              chalk.bold('yoshi build') +
```
ronami: maybe add `npx` too? to remove confusion.

ranyitz:
@ronami Thanks for the review, good points!

Fixed and ready for merge

PR #586, opened 2018-09-20
Server-side bundle MVP

### LINKED ISSUE #324: Changes in client side code causes the server side to refresh
# 🐛 Bug Report

### 😯 Current Behavior

Changes in client side code causes the server side to refresh. This causes to discard all in-memory changes in the server side, and makes it harder to persist state in server-side mocks/fakes.

### 🤔 Expected Behavior

Changes in client side code should cause refresh/hmr in client side.
Changes in server side code should cause refresh in server side.

### 🌍 Your Environment

```

  System:
    OS: macOS High Sierra 10.13.4
    CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
  Binaries:
    Node: 9.11.1 - ~/.nvm/versions/node/v9.11.1/bin/node
    Yarn: 1.6.0 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v9.11.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 67.0.3396.62
    Safari: 11.1
  npmPackages:
    yoshi: ^2.0.0-alpha.2 => 2.7.0 

```

ronami:
### 🔦 Summary

This PR is an attempt to change how Yoshi works for applications (client, full-stack, and universal). Instead of running different tools (babel/typescript, sass/less) on various glob patterns, it creates a dedicated Webpack bundle, specifically for the server.

#### This has a few advantages:

* When working in watch mode and a file is changed, we know if we should reload the client, server, or both. Currently, Yoshi has to restart the server on any file change (fixes #324).
* Universal projects no longer need to use require hooks to handle special (non js) assets (we can possibly remove `yoshi-runtime` in the future?).
* Makes it possible to deliver HMR experience for the server instead of restarting the entire process.
* Local development experience can be better. Yoshi runs just one tool with one watch, errors are handled in one place. Currently, the same error can show up in multiple places. Tools like `tsc` have very verbose logging and it's unclear when an operation has finished and if it finished successfully or with errors.
* Possibly faster build times (needs checking).

#### Problems I (so far) found:

* Resolving (non js) files should be done relatively to `process.cwd()` instead of `__dirname` or `require.resolve()` as the generated bundle will be in a different directory and `require.resolve` is handled by Webpack.

__The way to opt-into using this feature is by configuring `experimentalServerBundle` to `true`.__

The intent is to test this MVP on a couple of projects and get some feedback 🔙 .


packages/yoshi/src/commands/start-app.js
```diff
@@ -0,0 +1,147 @@
...
+  });
+
+  const devServer = new WebpackDevServer(clientCompiler, devServerConfig);
+
+  // Start up server compilation
+  let serverProcess;
+
+  serverCompiler.watch({ 'info-verbosity': 'none' }, (error, stats) => {
+    if (serverProcess && !error && !stats.hasErrors()) {
+      serverProcess.send({});
```
ranyitz: It worth adding a comment, regarding what happens when we send the empty object to the server process. (kill & restart ?)

packages/yoshi/src/commands/start-app.js
```diff
@@ -0,0 +1,147 @@
...
+      '0.0.0.0',
+      err => (err ? reject(err) : resolve()),
+    );
+  });
+
+  // Wait for both compilations to finish
+  try {
+    await compilationPromise;
+  } catch (error) {
+    process.exit(1);
```
ranyitz: When does this get to the catch block? Maybe we should say something to the user or add a comment for future readers?

I think it's actually `(stats)`

packages/yoshi/src/webpack-utils.js
```diff
@@ -0,0 +1,193 @@
...
+
+    const messages = formatWebpackMessages(stats.toJson({}, true));
+    const isSuccessful = !messages.errors.length && !messages.warnings.length;
+
+    if (isSuccessful) {
+      console.log(chalk.green('Compiled successfully!'));
+
+      if (isInteractive) {
+        const serverUrls = prepareUrls('http', '0.0.0.0', 3000);
+        const devServerUrls = prepareUrls('http', '0.0.0.0', 3200);
```
ranyitz: I think it worth getting those variables (3000, 3200) from `constants`/`yoshi-config`, the first due to duplication, and the second because the user might change it.

packages/yoshi/src/webpack-utils.js
```diff
@@ -0,0 +1,193 @@
...
+  const portFound = await waitPort({
+    port,
+    output: 'silent',
+    timeout: 5000,
+  });
+
+  if (!portFound) {
+    console.log(
+      chalk.red(
+        `\nCouldn't find a server running on port ${chalk.bold(3000)}.`,
```
ranyitz: `3000` port from a variable

packages/yoshi/src/webpack-utils.js
```diff
@@ -0,0 +1,193 @@
...
+      ),
+    );
+    console.log(
+      chalk.red(
+        `Please make sure "${chalk.bold(
+          'index.js',
```
ranyitz: I think that the user can configure this value (`--entry-point`?)

packages/yoshi/src/webpack-utils.js
```diff
@@ -0,0 +1,199 @@
...
+const webpack = require('webpack');
+const waitPort = require('wait-port');
+const clearConsole = require('react-dev-utils/clearConsole');
+const { prepareUrls } = require('react-dev-utils/WebpackDevServerUtils');
+const formatWebpackMessages = require('react-dev-utils/formatWebpackMessages');
+const project = require('yoshi-config');
+const { PUBLIC_DIR } = require('yoshi-config/paths');
+const { PORT } = require('./constants');
+
+const isInteractive = process.stdout.isTTY;
```
netanelgilad: I would also add a flag to disable the console clearing. Typescript also made a change to clear the console by default, got a backlash from the community and then added an opt-out flag as a compromise. Sometimes the user wants to keep his old logs.. :)

packages/yoshi/config/webpack.config.js
```diff
@@ -578,8 +611,114 @@ function createClientWebpackConfig({
...
+    },
+
+    // Do not replace node globals with polyfills
+    // https://webpack.js.org/configuration/node/
+    node: {
+      console: false,
+      global: false,
+      process: false,
+      Buffer: false,
+      __filename: false,
```
netanelgilad: I think we should use `__filename` and `__dirname`. Webpack can set them properly, so code that depends on it to read files or something like that can still work.

packages/yoshi/src/commands/build-app.js
```diff
@@ -0,0 +1,138 @@
+process.env.BABEL_ENV = 'production';
```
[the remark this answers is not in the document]
ranyitz: Basically, so it will be available for all `require`d packages. Some of them might do other stuff regarding BABEL_ENV/NODE_ENV.

Similarly to [this usecase](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/build.js#L11)

packages/yoshi/src/webpack-utils.js
ranyitz: I like [jest's `isInteractive`](https://github.com/facebook/jest/blob/6d3da9be06350bfa7414b609807ebd0cbdf0c416/packages/jest-util/src/is_interative.js#L3) function. One can just pretend to be a CI (by supplied an environment parameter or in the future `--ci`)

PR #589, opened 2018-09-25
Refactor webpack config to a single file

ronami:
### 🔦 Summary

This PR removes code duplication from different places while creating the Webpack config. It also puts most of the code for creating that config in one file, which I think makes it easier to maintain (a lot of things are related).

### 🗺️ Test Plan

There shouldn't be any breaking changes, so (hopefully) everything will be green 🍏.


packages/yoshi/config/webpack.config.js
```diff
@@ -0,0 +1,546 @@
...
+                            target: 'es2017',
+                          }
+                        : {}),
+                    },
+              },
+            },
+          ],
+        },
+
+        // Rules for JS / JSX
```
ranyitz: Only for JS

packages/yoshi/config/webpack.config.js
```diff
@@ -0,0 +1,546 @@
...
+        // Rules for assets
+        {
+          oneOf: [
+            // Inline SVG images into CSS
+            {
+              test: /\.inline\.svg$/,
+              loader: 'svg-inline-loader',
+            },
+
+            // Or return public URL to image resource
```
ranyitz: Or transform to base64 and inline?

Issue #583, opened 2018-09-20
Restore ability to generate webpack stats file


ronenst:
# 🙋 Feature Request

### 😯 Current Behavior
Since version `3.11.0` Yoshi stopped generating the `target/webpack.stats.json` file for performance reasons and will now only creates the file if the `--analyze` flag is set, since that was it's intended purpose.

In PR: https://github.com/wix/yoshi/pull/565

### 🤔 Expected Behavior
Add the ability to generate the stat files (both debug and production versions).

### 💁 Possible Solution 
Add a `--stats` or `--with-stats` or `--generate-stats` or `--profile` or `--with-profile` flag which will generate the files in the expected place.

This would also allow extending the flag in the future to take a filename template as the value of the flag. That is:
`yoshi build --stats file.json` 
Could generate `file.stats.json` and `file.min.stats.json`

### 🔦 Context 
This breaks anyone using the [business manager auditor](https://github.com/wix-private/business-manager-auditor) tool which relies on this file to check that unnecessary globally provided files are not bundled.

The `--analyze` flag can't be used in this instance since the auditor runs unattended and the analyzer doesn't halt.

In addition I have found it useful on occasion to use the non-minified version of the stats file when using [other third party tools](https://github.com/wix/yoshi/blob/master/docs/faq/WEBPACK-ANALYZE.md) to analyze the bundle, but now there is no way to generate one.

ranyitz:
Thanks @ronenst for opening this issue!

I agree that the static creation of the stats file can be useful and should be controllable.

`--stats` seems to be a good name for it, but I'm open for suggestions.

PRs welcome!

PR #591, opened 2018-09-26
yoshi: Add `--stats` flag to generate `dist/webpack-stats.json`

ronenst:
### 🔦 Summary
Fixes https://github.com/wix/yoshi/issues/583

Keeping it simple right now, `yoshi build --stats` will generate a file `dist/webpack-stats.json` this changes the previous output folder which was `target/`

### 🗺️ Test Plan
Changes are tested


packages/yoshi/config/webpack.config.js
```diff
@@ -495,14 +508,8 @@ function createClientWebpackConfig({ isAnalyze = false, isDebug = true } = {}) {
...
-      ...(isAnalyze
-        ? [
-            new BundleAnalyzerPlugin({
-              generateStatsFile: true,
-              // Path is relative to the output dir
-              statsFilename: '../../target/webpack-stats.min.json',
-            }),
-          ]
+      ...(isAnalyze || isStats
+        ? [new BundleAnalyzerPlugin(getBundleAnalyzerConfig(isStats))]
```
ranyitz: I think we can easily add support for this case: `yoshi build --analyze --stats`
Also, I think it's better to inline this code rather than export it to a different function if that's ok?

Something like:

```js
isAnalyze || isStats ? 
new BundleAnalyzerPlugin({
	analyzerMode: isAnalyze ? 'server' : 'disabled',
    generateStatsFile: isStats,
    statsFilename: STATS_FILE,
})
```

ronenst: I agree about inlining, I was on the fench but because the blocked looked very similar to the parent block I though it made sense, but when it's like this there's no need.

ronenst:
`yoshi-test` currently failing due to https://github.com/wix/yoshi/issues/595
cc: @ronami

PR #603, opened 2018-10-02
Update debugging.md > fix image links

sidoruk-sv:


ranyitz:
Thanks, @sidoruk-sv this is great!
The new path is correct but can be written in a shorter way.

Can you push a fix?

PR #612, opened 2018-10-07
Sentry error reporting

netanelgilad:
### 🔦 Summary
This PR adds error reporting to yoshi. Errors are reported and tracked by Sentry.

Currently only crashes are reported to Sentry. When `yoshi` encounters a fatal error, it sends that exception to Sentry for tracking. Changed behavior:
* Crashes no longer report the stacktrace to the user. Instead a message with possible steps is logged:

image

* Yoshi now checks a new environment variable called `DISABLE_SENTRY` and disables logging to sentry when it's set to a truthy value. Tests for yoshi have been marked with `DISABLE_SENTRY` to not pollute our logs. We need to also export that variable on our workstations, to avoid the same problem.

* The following information is collected on errors:
  * Node version
  * Operating System
  * yoshi version
  * full envinfo
  * yoshi config
  * **Name of the package yoshi was run in** (makes it non-anonymous)
  * command executed
  * NODE_ENV environment variable
  * CI true/false

* A new `error-feedback` page is added to yoshi's site. On crash errors, users can submit feedback using that page.

### 🗺️ Test Plan

As discussed, we have decided that testing reporting to Sentry would be an overkill at this stage. The case where sentry may cause problems is already when yoshi crashed, so testing that defeats the purpose. We will keep watch on the stability of this feature, and decide if tests are needed as we see fit.


packages/yoshi/src/sentry.js
```diff
@@ -0,0 +1,65 @@
...
+    beforeSend(event) {
+      if (event.level === Sentry.Severity.Fatal) {
+        console.log(
+          chalk.red(
+            '🤮  Oh, Yoshi just crashed. Care to help us with some feedback?',
+          ),
+        );
+        console.log(
+          `Go to ${chalk.green(
+            `http://localhost:8080/?eventId=${event.event_id}`,
```
yairhaimo: How does this work? Does sentry spawn a server for feedback? When does it stop it?

netanelgilad: Oh, nope. This is still for my testing. This is just a static HTML page that sends the feedback to Sentry. Once this is ready for merge I'll change it to the URL of yoshi's website where the static HTML will be hosted.

packages/yoshi/src/sentry.js
```diff
@@ -0,0 +1,103 @@
...
+const osName = require('os-name');
+const Sentry = require('@sentry/node');
+const chalk = require('chalk');
+const { getEnvInfo, loadEnvInfo } = require('../src/env-info');
+const path = require('path');
+const findPkg = require('find-pkg');
+
+module.exports = () => {
+  Sentry.init({
+    dsn: 'https://9325f661ff804c4a94c48e8c2eff9149@sentry.io/1292532',
```
netanelgilad: Do you think this should be hard-coded somewhere else?

netanelgilad:
Do you think I need to add some kind internal documentation on how errors are reported to Sentry and how to view them? So this knowledge will not only live in our brains...

packages/yoshi/src/commands/start-app.js
```diff
@@ -110,13 +110,7 @@ module.exports = async () => {
...
 
   // Wait for both compilations to finish
-  try {
-    await compilationPromise;
-  } catch (error) {
-    // We already log compilation errors in a compiler hook
-    // If there's an error, just exit(1)
-    process.exit(1);
-  }
+  await compilationPromise;
```
netanelgilad: @ronami Can you review this change? I wasn't sure exactly why you are `try/catch`ing these promises and what kind of errors they would throw. I guess they are not `UserLandError`? Any way, since the catch handling was just `process.exit(1)`, I've let the error get to the `unhandledRejection` handler. The only thing that this might create a double logging of errors.

netanelgilad:
As discussed, we have changed the way commands handle and propagate errors. Commands will now handle any known errors by themselves, and `process.exit` where appropriate. If the command function does throw an error, it is an unknown error and therefore will get to the top level error handler, and be treated as a crash. That means they'll get reported to Sentry and a crash error log will be printed.

@ronami @ranyitz Please review and see if my last change fits what we talked about.

packages/yoshi/package.json
```diff
@@ -10,10 +10,11 @@
     "url": "https://github.com/wix/yoshi"
   },
   "scripts": {
-    "test": "mocha './+(test|src)/{,!(fixtures)/**/}/*.spec.js' --retries=1",
-    "test:watch": "mocha './test/{,!(fixtures)/**/}/*.spec.js' --watch"
+    "test": "DISABLE_SENTRY=1 mocha './+(test|src)/{,!(fixtures)/**/}/*.spec.js' --retries=1",
```
ranyitz: Maybe it worth adding it in the e2e suite as well

packages/yoshi/src/commands/test.js
```diff
@@ -123,9 +124,11 @@ module.exports = runner.command(
 
         await runMocha(); // fail silently
       } else {
-        await runMocha(error => {
-          throw `mocha failed with status code "${error.code}"`;
-        });
+        await printAndExitOnErrors(() =>
+          runMocha(error => {
+            throw `mocha failed with status code "${error.code}"`;
```
ranyitz: Should move to the new way

packages/yoshi/src/env-info.js
```diff
@@ -0,0 +1,42 @@
...
+let envinfoCache = undefined;
+
+function loadEnvInfo() {
+  envinfoCache = envinfo.run(
+    {
+      System: ['OS', 'CPU'],
+      Binaries: ['Node', 'Yarn', 'npm', 'Watchman'],
+      Browsers: ['Chrome', 'Firefox', 'Safari'],
+      npmPackages: [
+        'yoshi-config',
```
ranyitz: can bring dynamically, future proof

packages/yoshi/src/sentry.js
```diff
@@ -0,0 +1,106 @@
...
+    beforeSend(event) {
+      if (event.level === Sentry.Severity.Fatal) {
+        printCrashInfo(event.event_id);
+      }
+
+      return event;
+    },
+  });
+
+  loadEnvInfo();
```
ranyitz: No need

packages/yoshi/src/error-handler.js
```diff
@@ -0,0 +1,12 @@
+async function printAndExitOnErrors(fn) {
+  try {
+    return await fn();
+  } catch (err) {
+    console.error(err);
```
ranyitz: chalk.red

packages/yoshi/src/sentry.js
```diff
@@ -0,0 +1,106 @@
...
+  console.log();
+  console.log(`    * Find us on slack on the #yoshi channel:`);
+  console.log(`      ${outputURL('https://wix.slack.com/messages/yoshi/')}`);
+  console.log();
+  console.log(
+    '    * Open a bug report on our github repository (be sure to include your event id):',
+  );
+  console.log(
+    `      ${outputURL(
+      `https://github.com/wix/yoshi/issues/new?template=BUG_REPORT.md&labels=%F0%9F%90%9B%20Bug`,
```
ranyitz: remove bug label

netanelgilad:
@ranyitz Can you review my last commit and let me know if you have any other pointers?

I think I need to add unit/integration tests, but we don't have an infrastructure for that yet. You think maybe it should be done in another PR? Maybe not worth it for now?

packages/yoshi/src/sentry.js
```diff
@@ -0,0 +1,109 @@
...
+
+    process.exit(1);
+  });
+}
+
+function outputURL(url) {
+  return chalk.cyan(url);
+}
+
+function printCrashInfo(eventId) {
```
yairhaimo: Maybe change the message to a template literal?
(I dont know how to add suggestions for multi lines)

```suggestion
function printCrashInfo(eventId) {
console.log(`
${chalk.redBright(`🤮  Oh no, Yoshi just crashed! ${chalk.yellow(`Event ID = ${eventId}`)}`)}

  Here are a few options on what you can do now:

    * Help us figure out what caused the crash by submitting feedback on the error. You can do that by going to:
      ${outputURL(
        `https://wix.github.io/yoshi/error-feedback/?eventId=${eventId}`
      )}

    * Find us on slack on the #yoshi channel:
      ${outputURL('https://wix.slack.com/messages/yoshi/')}

    * Open a bug report on our github repository (be sure to include your event id):
      ${outputURL(
        `https://github.com/wix/yoshi/issues/new?template=BUG_REPORT.md`
      )}

  `);
  }
```

PR #618, opened 2018-10-08
Forward jest options

### LINKED ISSUE #321: Allow passing cli arguments to jest, mocha, ...

ranyitz:
### 🔦 Summary
This PR fixes #321 and is the continue of @amitdahan work on #507.

This PR will enable users to pass CLI options to jest bin by placing them after `--jest` option.

* `yoshi test --jest --forceExit` would work
* `yoshi test --forceExit --jest` would __not__ work

__The debug case:__
* `yoshi test --jest --debug` would use [`jest --debug`](https://jestjs.io/docs/en/cli#debug)
* `yoshi test --debug --jest` would use `node --inspect` to run jest

__Notes:__
One option was to create a list of jest options, we could actually retrieve it from [here](https://github.com/facebook/jest/blob/master/packages/jest-cli/src/cli/args.js#L72) and dynamically create the options for the commander validation.

The other was to enable unknown options (`.allowUnknownOption()`);
I chose the second one as the validation did not seems important enough.

I rearrange some tests that were messy.


ranyitz:
After a discussion with @ronami, I've changed the implementation of this feature a bit:

* Accept command line arguments/options before `--jest`
* Use `debug` for jest debugging instead of [jest config debug](https://jestjs.io/docs/en/cli#debug)
* Add a tiny fix to add `--runInBand` only if there is no `--runInBand` which was supplied already. (to prevent multiple `--runInBand` options and mysterious bugs)

Ready for a review

PR #633, opened 2018-10-14
fix generated launch.json for vscode across templates

netanelgilad:
### 🔦 Summary
This PR fixes the generated `launch.json` for debugging with vscode for all templates. This uses jest when available in the template or just `yoshi test` otherwise.

### 🗺️ Test Plan
N/A

### DIFF
--- a/packages/create-yoshi-app/templates/client/.vscode/launch.json
+++ b/packages/create-yoshi-app/templates/client/.vscode/launch.json
@@ -1,10 +1,67 @@
 {
   "version": "0.2.0",
   "configurations": [{
-    "name": "Launch Mocha",
+    "name": "All Component Tests - Jest",
     "type": "node",
     "request": "launch",
-    "program": "${workspaceRoot}/node_modules/yoshi/debug/mocha",
-    "cwd": "${workspaceRoot}"
+    "program": "${workspaceRoot}/node_modules/.bin/jest",
+    "args": [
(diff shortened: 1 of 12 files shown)


packages/create-yoshi-app/templates/client/.vscode/launch.json
```diff
@@ -1,10 +1,55 @@
...
+    "cwd": "${workspaceRoot}",
+    "env": {
+      "MATCH_ENV": "component"
+    }
+  },
+  {
+    "name": "Current Component Tests - Jest",
+    "type": "node",
+    "request": "launch",
+    "program": "${workspaceRoot}/node_modules/jest/bin/jest",
```
hepiyellow: @netanelgilad 
Hi, 
When running jest directly , does it get the same configuration as when running jest through yoshi (`npx yoshi test --jest`)?

ranyitz: Currently when using `jest-yoshi-preset` -> https://github.com/wix/yoshi/blob/master/packages/jest-yoshi-preset/jest-preset.js#L64

ranyitz: important to say, you'll need to run `npx yoshi start --no-test` so you'll have a bundle served. But you'll get a console message for that.

PR #636, opened 2018-10-14
jest-yoshi-preset: Move stylable from devDependencies to dependencies

ronenst:
### 🔦 Summary
`stylable` and `stylable-integration` are both needed for runtime code and should be in dependencies.

### 🗺️ Test Plan
N/A


ronenst:
Sure, but still they should be in `dependencies`, right?

ronenst:
Maybe not the second one...

ranyitz:
Thanks @ronenst, This was fixed on #637

PR #638, opened 2018-10-15
Improve e2e tests by emulating production installations

ronami:
### 🔦 Summary

This PR changes how our E2E tests install the dependencies of generated projects.

Instead of installing dependencies and then copying/linking the version of Yoshi/others, it creates a local npm registry and publishes the entire monorepo to it. It then installs all dependencies from that local registry to mimic the tree as much as possible.

### DIFF
--- a/packages/create-yoshi-app/src/utils.js
+++ b/packages/create-yoshi-app/src/utils.js
@@ -13,6 +13,7 @@ module.exports.install = dir => {
   execa.shellSync('npm install', {
     cwd: dir,
     stdio: 'inherit',
+    extendEnv: false,
   });
 };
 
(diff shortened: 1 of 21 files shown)


packages/create-yoshi-app/src/utils.js
```diff
@@ -13,6 +13,10 @@ module.exports.install = dir => {
   execa.shellSync('npm install', {
     cwd: dir,
     stdio: 'inherit',
+    extendEnv: false,
```
ranyitz: Worth adding a comment that we don't want to `extendEnv` because when running with `npm run` npm is passing `npm_config_prefix` that leads to the global npm config which makes `npm install` to ignore the local `.npmrc` file and prevent it from going to the private registry.

PR #646, opened 2018-10-16
Update Babel to 7

### LINKED ISSUE #660: Upgrade to Babel 7
# 🙋 Feature Request

### 😯 Current Behavior

Yoshi currently uses babel 6.

### 🤔 Expected Behavior

Yoshi should use babel 7.

yavorsky:
### 🔦 Summary
This PR updates babel to 7.
The full list of changes could be found:
https://babeljs.io/blog/2018/08/27/7.0.0
https://babeljs.io/blog/2017/12/27/nearing-the-7.0-release

- For yoshi this means that we can take all advantages of `babel.config.js`, have development mode for react preset, customize better babel-preset-env. The other part is internal changes that shouldn't affect our API.
- Some output / error handling was changed, so I've updated some tests.
- `hmr-runtime` plugin was moved to the separate package since `babel-helper-transform-fixture-test-runner` was moved to jest and doesn't support mocha.
- `parserOptions` was also deprecated, so we need to enable syntax plugin for dynamic import.
- `polyfills` option for `@babel/register` was also removed. There is `preset-env` to transform polyfills dynamically.

There was an intention to get rid of haste-task-*, so the new babel-task was moved into yoshi package and we have a single source of truth with babel version.

`IN_WEBPACK` env variable override was also moved into `EnvirnmentMarkPlugin`.

Closes #660


lerna.json
```diff
@@ -15,14 +15,5 @@
       "PR: Internal :house:": ":house: Internal"
     }
   },
-  "command": {
```
ronami: any reason to remove that?

yavorsky: Initially, I've updated lerna to 3.x in this PR. But then I decided that it should cintain some breaking changes and worth separate PR.
This command is being applied only for lerna 3 since lerna 2 doesn't respect `allowBranch` feature.
I'll discard this change and will include it as a part of lerna 3 update in a separate PR.

Thanks! :)

packages/babel-preset-yoshi/index.js
```diff
@@ -20,63 +20,67 @@ module.exports = function(api, opts = {}) {
...
       ],
-      // Enable legacy decorators.
-      require('babel-plugin-transform-decorators'),
-      // Leave dynamic imports untranspiled for webpack
-      inWebpack
-        ? require('babel-plugin-syntax-dynamic-import')
-        : require('babel-plugin-dynamic-import-node').default, // https://github.com/airbnb/babel-plugin-dynamic-import-node/issues/27
+      // Enable stage 2 decorators.
+      [
+        require('@babel/plugin-proposal-decorators'),
```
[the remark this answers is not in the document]
yavorsky: Good catch! Thanks!

ronami:
@yavorsky, can we merge?

yavorsky:
@ronami sure, just need to figure out what's going on with tests

ronami:
Our e2e tests are having some flaky problems, @netanelgilad and I are looking into it.

PR #663, opened 2018-10-23
fix webpack bundles externals inside specs.bundle

### LINKED ISSUE #662: externals fields doesn't effect specs.bundle
# 🐛 Bug Report

### 😯 Current Behavior

source code of packages that are listed in `externals` appears in `specs.bundle.js`

### 🤔 Expected Behavior

source code of packages that are listed in `externals` **should not** appears in `specs.bundle.js`

### 🔦 Context 

we have a project that uses `yoshi@2` and has `angular`.
in that project we set `angular` as external, and load it in karma.

after upgrading to to `yoshi@3` we get an error that angular were loaded more than once
after looking at `specs.bundle.js` we saw that angular source appears there

### 💁 Possible Solution 

need to set externals also in spec's webpack config

### 👣 Steps to Reproduce 

see context

### 💻 Code Sample

...

### 🌍 Your Environment

```
// PASTE HERE
```

eddierl:
### 🔦 Summary
fix #662

add external to config
verify it doesn't appear in specs.bundle

### DIFF
--- a/docs/guides/svg.md
+++ b/docs/guides/svg.md
@@ -4,7 +4,8 @@ title: How to use SVG?
 sidebar_label: How to use SVG?
 ---
 
-There are two ways of using SVG:
+There are few ways of using SVG:
 - Just import it or use `background: url()` in your css and it will be inserted as data URI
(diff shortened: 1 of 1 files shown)


eddierl:
I changed my fork branch from master
created another PR #665

PR #670, opened 2018-10-25
add babel-runtime as a dependency and remove regenerator-runtime

ranyitz:
### 🔦 Summary
When you want to use [babel-plugin-transform-runtime](https://babeljs.io/docs/en/next/babel-plugin-transform-runtime.html#installation) you should also install `babel-runtime` in your dependencies.

In case you don't do that, and use something like `async/await` things will break.

This PR adds this to fullstack projects.

We should probably document it in the preset docs. (for users that migrate to it)


yavorsky:
@ranyitz  do we need to do the same for [universal project](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/universal/%7B%25packagejson%25%7D#L43) ?

netanelgilad:
@ranyitz Wanna add this to the universal + server templates and we'll merge?

PR #671, opened 2018-10-25
[Create Yoshi App] New Ping Emails Template and minor updates

illBeRoy:
#### New Template:
Added new template under `templates/ping-email-template`. This template is meant for people who want to create statics projects for Ping Email Notifications. The template was locally used by my dev version of `create-yoshi-app` and was found to be fully operational.

#### Changes in Create Yoshi App:
1. Changed the "transpiler type" question to be optional: if the selected `projectType` does not have a `-typescript` version, it will not ask the user to select transpiler type. This was achieved using `prompts` [type as function method](https://github.com/terkelg/prompts#-prompt-objects) (changes in [`getQuestions.js`](https://github.com/wix/yoshi/compare/master...illBeRoy:generator/ping-email-templates?expand=1#diff-4f45e314e072d4c4031be313e245fa8fR46)).
2. I've added support for a new `#generator.options.json` file in the template directory. If the file exists, `create-yoshi-app` will read it and will add \ remove stages accordingly. `create-yoshi-app` will also **not** copy the file into the working directory. At the moment, the `#generator.options.json` file only supports `doNotLint`, which makes `create-yoshi-app` skip the linting phase. This behavior is backwards compatible, and other templates are not changed because of it. (changes in [`getGeneratorsOptionsForTemplate`](https://github.com/wix/yoshi/compare/master...illBeRoy:generator/ping-email-templates?expand=1#diff-0828623616125fd841d0161a3fbe48d2), [`createApp`](https://github.com/wix/yoshi/compare/master...illBeRoy:generator/ping-email-templates?expand=1#diff-0828623616125fd841d0161a3fbe48d2) and [`getFilesInDir`](https://github.com/wix/yoshi/compare/master...illBeRoy:generator/ping-email-templates?expand=1#diff-8930264b3ec62707b9373e20000219afR6))

### 🗺️ Test Plan
I've added tests for the new `generatorOptions` unit in the `__tests__` directory.


illBeRoy:
Hey @netanelgilad  :)
I was under the impression that `create-yoshi-app` is our official company-wide go-to tool for generating front end projects, and always the first stop for every fed who starts working on one.
As such, as long as it is fine with you, we would rather integrate with `create-yoshi-app` than require people to install and use tools that are not as commonly used.
It is true that the project in question does not use yoshi, this is because it does not require building nor testing.
I hope this made sense :) and thanks again for taking the time to review

netanelgilad:
Sure, I'll let my wiser teammates drop in their opinion :). I'm not sure we'd want to maintain a template that doesn't really integrate with `yoshi`, but maybe it's time to make `create-yoshi-app` support external templates like `yo` does? could be an interesting solution.

yairhaimo:
Closing this PR as it is being dealt in a different way: extracting the core logic of `create-yoshi-app` into a generic mechanism.

PR #675, opened 2018-10-26
Initial implementation on the yoshi side where different user agents will receive best bundles for them. (IE - HUGE, Latest Chrome - small one)

yavorsky:
### 🔦 Summary
This PR implmements idea to create multiple bundles for different environments and then serve them for according to the client's useragent.
To implement it fully we need also to support on CDN, server side, but we can work on it later.

### `bundleTargets` option for  `yoshi build`.
Note: It should create multiuple bundles only for **minified** version.

`yoshi.bundleTargets`: allows to specify Array with browserslists configuration for each target that should be created.

For example, `['not dead', 'last 1 chrome version']` will create 2 bundles:
1 - fallback bundle that supports even IE (*app.bundle.min.js*)
2 - bundle for 'modern' browsers. (*app.{bundleHash}.bundle.min.js*)

`bundleHash` here is a hash from current modern browsers browserslist configuration generated by bundle utils:
`'last 1 chrome version' -> 'a460ac775beb03ed5a573b0a0f905596'`

So as we can see, first bundle name stays w/o any changes to not break anything. 2+ bundles should point to more specific targets and contains hash in file name. To get the hash you can use

```js
import bundleUtils from 'bundle-utils';

bundleUtils.getId('last 1 chrome version') // will return md5 hash for current name.
```

### `nomodule`/`type="module"` support for CYA **client** and **fullstack** versions.
PR also changes template for client's app `index.vm` and `package.json`.

*index.html* should look like:

```html
<!-- Fallback bundle for everything that doesn't support type 'module' (IE, old browsers) -->
<script nomodule src=".../app.bundle.min.js"></script>
<!-- Bundle for new browsers that support type module -->
<script type="module" src=".../app.a460ac775beb03ed5a573b0a0f905596.bundle.min.js"></script>
```

#### Next spets:
1. fullstack/universal apps implementation.
2. investigate cdn implementation
3. add more tets

I've already tested it with guineapig 2. Will share results for this and other projects a bit later.
Will add mode description soon..

TLDR:

screen shot 2018-10-15 at 11 25 14 am

Some transform stats:
screen shot 2018-10-15 at 11 23 06 am


netanelgilad:
Looks awesome! 🍻  does this PR just generate different bundles based on the `bundleTargets` config? Maybe we can start incorporating this without waiting for server/cdn changes using `<script module ...>` and `<script nomodule ...>` as described in https://philipwalton.com/articles/deploying-es2015-code-in-production-today/. That way we'll be able to break this feature into 2 parts, and gain some benefits with minimal changes to the stack.

yavorsky:
@netanelgilad Thanks! Actually, it was my plan!
Currently, we are adding the feature to create multiple bundles to yoshi, then going to update create-yoshi-app and preconfigure it + `nomodule `/`module` and then investigate cdn/backend integration.

PR #677, opened 2018-10-28
Look for a fake-server entry if the standard server entry isn't found

ronami:
### 🔦 Summary

If you're using a client project with a fake server, it doesn't make sense to put your fake server in `src`. This PR will look for a 2nd entry if the 1st one wasn't found.

__Note that this is specifically for the experimental server bundle.__

### DIFF
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -630,7 +630,7 @@ function createServerWebpackConfig({ isDebug = true } = {}) {
     target: 'node',
 
     entry: {
-      server: ['./server', '../test/fake-server'].find(exists),
+      server: ['./server', '../test/dev-server'].find(exists),
     },
 
     output: {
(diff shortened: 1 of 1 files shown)


packages/yoshi/config/webpack.config.js
```diff
@@ -619,7 +630,7 @@ function createServerWebpackConfig({ isDebug = true } = {}) {
     target: 'node',
 
     entry: {
-      server: './server',
+      server: ['./server', '../test/fake-server'].find(exists),
```
netanelgilad: Maybe a  better name instead of `fake-server`. dev or local would probably make more sense

ranyitz: `dev-server` reminds me of `webpack-dev-server` (Sorry for the bikeshedding). Och, why naming things is so hard ...

netanelgilad: IMO that's good. Why not share the same terms? less mental load. It serves the same meaning in both places, a server for development. Though it's true they are not doing exactly the same thing.

PR #683, opened 2018-10-28
print error messages and add verbose flag for full stacktrace

netanelgilad:
### 🔦 Summary
Currently errors are swallowed by yoshi and sent to Sentry. But sometimes these errors could be errors created by the user, and by viewing the error message they can know what they did wrong. This PR adds printing of the error message and a `--verbose` flag that allows printing the full stacktrace. Sending the error to Sentry wasn't changed.

### 🗺️ Test Plan
N/A


ranyitz:
@netanelgilad I think it worth adding this option to the [cli reference documentation](https://github.com/wix/yoshi/blob/master/docs/api/cli.md).

netanelgilad:
> @netanelgilad I think it worth adding this option to the cli reference documentation.

Most Def! on it!

ranyitz:
@netanelgilad Awesome, thanks :)

Issue #657, opened 2018-10-23
add `index-dev` concept to the Fullstack templates.


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior
When a new fullstack project is generated, yoshi runs `index.js` file when the user run `yoshi start`.

After the user needs to add experiments, rpc etc... He/She will need to add [`--entry-point`](https://wix.github.io/yoshi/docs/api/cli#start) to the start command with a different file, and add a mock server for those services (or the app won't work in development).

This is not easy to find, some people end up with a non functional `start` experience, and some find other conventions and weird solutions.

### 🤔 Expected Behavior

The user create a new project and already has an `index-dev.js` file configured, so adding `experiments`/`rpc` becomes a simple task.

### 💁 Possible Solution

The first step would be to add `--entry-point` & `index-dev.js` to the Fullstack templates.

The second step would be to remove the `--entry-point` configuration (A breaking change, should be in a major version) and make it a default behavior.

### 🔦 Context

Regardless of the question if we need to bring experiment/rpc by default for our users, we definitely should create a unified way of running the server on development.

While client projects use the `--entry-point` feature for a `fake-server` and fullstack project use it to bootstrap the environment, it is still an opt-in feature.

I think that we can aim towards making it the default behavior.

ronami:
I don't think we should wait to Yoshi v4 for this, I think it can be merged to master.

PR #687, opened 2018-10-29
Add index-dev to fullstack generators

yanivefraim:
Solves (part of) #657. 

1. I added `--no-test` to `npm start`. I know that this is not part of this task, I can remove it and move to a later PR.

2. I did not update universal (they are totally outdated)

3. This is a refactor of `start`. I did not add tests (I think we should have tests for generator + start)

4. After this PR is merged, I will update https://github.com/wix-private/fed-handbook/blob/master/RPC.md#add-a-new-rpc accordingly


packages/create-yoshi-app/templates/fullstack-typescript/index-dev.js
```diff
@@ -0,0 +1,12 @@
+const { bootstrapServer, emitConfigs } = require('./dist/test/environment');
+
+(async () => {
+  const app = bootstrapServer({
+    port: 3000,
+    managementPort: 3001,
+    appConfDir: './target/configs',
+  });
+
+  await emitConfigs({targetFolder: './target/configs'});
```
roymiloh: I would change it to a dedicated target folder (for `start` only), as it might collides with testing stuff.

packages/create-yoshi-app/templates/fullstack-typescript/index-dev.js
```diff
@@ -0,0 +1,12 @@
+const { bootstrapServer, emitConfigs } = require('./dist/test/environment');
```
roymiloh: Better to run (at least) babel register or smth similar. Importing from `./dist` is a bit awkward.

packages/create-yoshi-app/templates/fullstack-typescript/index-dev.js
```diff
@@ -0,0 +1,12 @@
...
+
+(async () => {
+  const app = bootstrapServer({
+    port: 3000,
```
roymiloh: Shouldn't we use smth like `process.env.PORT`? allocated by yoshi. Even better - we might even not need this (in case env vars are propagated, it should be working ootb - see [this](https://github.com/wix-platform/wix-node-platform/blob/master/bootstrap/wix-bootstrap-testkit/lib/server.js#L10)).

netanelgilad: Yep that's true. `yoshi` has features that rely on knowing the port the server starts on, like waiting for the server to start and opening the browser automatically (starting from `yoshi@4`). So we should use the env var.

packages/create-yoshi-app/templates/fullstack-typescript/index-dev.js
yanivefraim: @roymiloh - You are right. Fixed

packages/create-yoshi-app/templates/fullstack-typescript/index-dev.js
yanivefraim: done. I added `target/dev` (wasn't sure if `target/config-dev` is better. WDYT?)

packages/create-yoshi-app/templates/fullstack-typescript/index-dev.js
yanivefraim: changed to `process.env.PORT`

PR #688, opened 2018-10-29
validate @wix email on project generation

### LINKED ISSUE #684: Generators: author email should be mandatory, with `@wix.com` validation
This is an internal/private Wix feature, but it is crucial when generating a new project into our ci/cd systems. 

If we miss a `@wix.com` email, in `pom.xml`, the newly generated project will not have an owner, leading to a readonly project in Fryingpan. This cannot be changed later (only manually, by system guys).

netanelgilad:
### 🔦 Summary
Fixes #684.

### 🗺️ Test Plan
I couldn't find a way to test it easily with `prompts`. It seems that `prompts.inject` skips validation, so really testing it would be quite a hassle. IMO not worth it.


packages/create-yoshi-app/src/getQuestions.js
```diff
@@ -1,7 +1,7 @@
 const getGitConfig = require('parse-git-config');
 const getProjectTypes = require('./getProjectTypes');
 
-module.exports = async () => {
```
ronami: any reason to remove the `async` keyword? if any errors are thrown it will still return a rejected promise.

PR #696, opened 2018-10-30
change default timeout for e2e tests, in fullstack generators

### LINKED ISSUE #685: Jest default timeout (5000ms) should be changed for e2e tests
Running e2e tests with Jest's default timeout (5000ms) might result in flakiness. I think we should change the default to be at least 10000.

yanivefraim:
Fixes #685 - this will add a timeout of 10000 for e2e tests in fullstack generators


yanivefraim:
> Looks good, I like that it shows users how they can have their own setup script. Like @netanelgilad said, I think we should also add it to the client template.

done...

PR #699, opened 2018-10-31
Serve JS files on CDN with proper content type

yurynix:
### 🔦 Summary

Chrome 71 blocks JS files in WebWorkers that aren't served with the proper `Content-Type` header.
Chrome 70 displays the following message:
screen shot 2018-10-31 at 10 32 44

That makes development of platform apps locally impossible 😭 on Chrome 71.

### 🗺️ Test Plan

I've added an integration test, we can also test that on form-builder platform app.


netanelgilad:
So it seemed weird to me that we are not already returning the correct `Content-Type`, especially since we are using `webpack-dev-middleware` and I was sure they implemented it correctly. So I went to their code and saw that they are actually setting the correct `Content-Type`.
So I searched where are we causing the correct `Content-Type` to go away, and the problem was with a middleware we have called `redirectMiddleware` and it is doing a proxy request but ruining the headers. I changed the function to use `http-proxy` and now the test works without the extra middleware. Here is the new function:
```
function redirectMiddleware() {
  const proxy = httpProxy.createProxyServer({
    secure: false,
    ignorePath: true,
  });
  return (req, res, next) => {
    if (!/\.min\.(js|css)/.test(req.originalUrl)) {
      return next();
    }

    proxy.web(req, res, {
      target:
        req.protocol +
        '://' +
        req.get('host') +
        req.originalUrl.replace('.min', ''),
    });
  };
}
```
@yurynix What do you think? I think this would be the better solution.

yurynix:
@netanelgilad indeed! Great find! I'll change the code on Sunday in this PR if you won't beat me to it!

netanelgilad:
Good point. I know another dependency is bloat, but I don't know where this might bite us again, since the real use case of the redirect middleware is to proxy. But I'm probably being too cautious here. This debate is as old as time itself. Should I bring in another dependency to handle a small use case, or use small self written solution? Don't know which side I'm on. @ranyitz @ronami Share your wisdom.

PR #727, opened 2018-11-05
 Disable wix-style-react lint rule

lbelinsk:


lbelinsk:
@ranyitz I would prefer not to remove it completely as long as we don't see for sure that the tree shaking is truly working for our consumers and that there are no issues.
After we see that everything is ok, this lint rule could be removed entirely.

If you still prefer to wipe it out instead of disabling it, let me know and I will take care of it.
Thanks

lbelinsk:
I tested this on some project, and there is a guineapig project which works on an alpha version of this feature and he is not complaining about any issue.
That is why I think that if something will not work, it might be some edge case which will be hard to find.

I believe that there is a good chance that the lint configuration which the users will add in their project will never be removed by them, even after we announce that everything is ok.

It is your call, but I just think it makes more sense to "pollute" only one place (although it is in Yoshi) where we are in charge of cleaning it in time, instead of asking all of the users to add something they would probably never remove.

With that said, are we still in the same plan?
If so, we can close this pr and I will instruct our users to override the lint rule in their project.

ranyitz:
@lbelinsk 

> It is your call, but I just think it makes more sense to "pollute" only one place (although it is in Yoshi) where we are in charge of cleaning it in time, instead of asking all of the users to add something they would probably never remove.

It's not only my call, we are just discussing it.

Removing this lint rule is not a breaking change, which means we don't have to wait for a major version in order to remove it. To put it back is a breaking change so it would be a bit harder. I think we should keep that in mind before doing the change, and than only to do so after we sure that it's working.

When you are confident that wsr can be tree shaked, let's remove/adjust this rule accordingly.

______

Just a tip,

You have the full control over - https://github.com/wix/wix-ui/tree/master/packages/eslint-plugin-wix-style-react & https://github.com/wix/wix-ui/tree/master/packages/tslint-plugin-wix-style-react

PR #735, opened 2018-11-06
Start a forward proxy to intercept public cdn requests to the local cdn

### LINKED ISSUE #707: e2e tests with dynamic imports & static cdn public path will fail
# 🐛 Bug Report

### 😯 Current Behavior

Currently in CI, we are hard coding the cdn url into the public path during build. This is done so that dynamic imports and other assets would work when deployed to production. We are also running e2e in CI on the created bundle, before the assets were deployed to the CDN. Which means the browser will try to download assets that don't really exist.

### 🤔 Expected Behavior

E2E tests with dynamic imports should not fail in CI.

### 🔦 Context 

It seems that the only reason this still works for the existing projects is that they are still using the dynamic public path plugin which takes the public path from the global variable supplied by the templated HTML (which is localhost:3200 in e2e tests).

### 💁 Possible Solution 

A possible solution would be to use a forward proxy that will transfer the requests that are sent to parastorage to the local CDN that is started during e2e tests (which `yoshi` provides). That could be [possibly done with puppeteer](https://github.com/GoogleChrome/puppeteer/issues/336), and any configurable forward proxy.

netanelgilad:
### 🔦 Summary
Fixes #707. `yoshi` will start a forward proxy that will intercept requests made for the public cdn, and will proxy them to the local cdn instead. This solution fixes the problems both for puppeteer and protractor. 

For protractor - this PR configures the tests to run with chrome and start it with the proxy. If a project is configuring their own browser, they may not get this configuration and will need to do it themselves.

### 🗺️ Test Plan
Added a test for protractor, that runs a CI build and test of a project with dynamic imports, and makes sure the tests pass. also made sure that without the new changes the new test doesn't pass :)


packages/yoshi-helpers/rewrite-forward-proxy.js
```diff
@@ -0,0 +1,85 @@
...
+const path = require('path');
+const url = require('url');
+const enableDestroy = require('server-destroy');
+
+module.exports = async function startRewriteForwardProxy({
+  search,
+  rewrite,
+  port,
+}) {
+  const regularProxy = createProxyServer({ ignorePath: true });
```
Schniz: have you considered listening to `error` events? I think they can make the app die (or maybe it is just old node apps, maybe worth sending errors to sentry?)

Schniz: I'm not sure that `http-proxy` even has these listeners, but I remember I had this problem

packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -36,9 +41,27 @@ module.exports = async () => {
     // start with a few new lines
     console.log('\n\n');
 
+    const forwardProxyPort = process.env.FORWARD_PROXY_PORT || 3333;
+
+    if (shouldDeployToCDN()) {
+      global.__closeProxy = await startRewriteForwardProxy({
```
Schniz: what do you think about having an external file that exposes a "singleton" for the puppeteer tests with just 2 methods: `start` and `stop`? This way it won't be on `global`, and you could require it from both files (`globalSetup`, `globalTeardown`) without cross-requiring them (if that's even a word).
Check this out:

```js
let stopFunction = null;
module.exports.stop = async () => stopFunction && stopFunction();
module.exports.start = async () => {
  if (!stopFunction) {
    stopFunction = await startRewrite...
  }
};
```

then just call `proxyServer.start()` in this file

netanelgilad: You are definitely right, that's a much better option! I was so fixed on putting it in the `global` to share state between them (like the other objects with have on the `global` there) that I forgot they share the same module cache! Will fix.

packages/yoshi-helpers/rewrite-forward-proxy.js
netanelgilad: That's a good point but here is my consideration. Any uncaught error gets sent to Sentry anyway, so if the errors from the proxy cause the app to crash, we'll be notified. If those are errors that don't cause a crash (which seems unlikely when I think about) they will be swollen silently, and I may prefer that because I don't want to pollute the user stdout if it doesn't cause a fail. But maybe I need to log them if yoshi was run with a verbose flag? What do you think?

Schniz: Oh, we're already sending every exception and optionally logging with the verbose flag:

https://github.com/wix/yoshi/blob/119354d3faa330c1e1a1fa11211bd47a38df0a3a/packages/yoshi/bin/yoshi-cli.js#L119-L120

so everything is good
and my comment should be ignored!

Schniz: i forgot that overriding `uncaughtException` changes the default behavior which is dying 😥

packages/jest-environment-yoshi-puppeteer/cdnProxy.js
```diff
@@ -0,0 +1,17 @@
+const startRewriteForwardProxy = require('yoshi-helpers/rewrite-forward-proxy');
```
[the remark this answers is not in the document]
netanelgilad: I thought start was better since it actually starts it and not only creates a instance which could later be started. WDYT?

packages/yoshi-helpers/rewrite-forward-proxy.js
```diff
@@ -0,0 +1,85 @@
...
+        protocol + '://' + req.headers.host + url.parse(req.url).path;
+      if (target.startsWith(search)) {
+        target = target.replace(search, rewrite);
+      }
+
+      regularProxy.web(req, res, { target });
+    };
+  }
+
+  const httpsReverProxyServer = https.createServer(
```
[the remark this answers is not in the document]
netanelgilad: Good catch!

packages/yoshi/test/test.spec.js
```diff
@@ -214,6 +216,57 @@ describe('Aggregator: Test', () => {
...
+              browser.ignoreSynchronization = true;
+              browser.get("http://localhost:1337");
+              const until = protractor.ExpectedConditions;
+              browser.wait(until.presenceOf(\$('h1')), 8000, 'Element taking too long to appear in the DOM');
+              expect(element(by.css("body")).getText()).toEqual("Dynamic");
+            });
+          `,
+      });
+
+      const buildResponse = project.verbose().execute('build', [], {
```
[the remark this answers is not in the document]
netanelgilad: Super catch!

netanelgilad:
@ranyitz You are super right about `puppeteer`. Wanted to add a test, but then I saw we have no tests at all for `puppeteer` and it seems harder to test, because our current test infra doesn't copy all the monorepo to the create temp folders, only `yoshi` and `yoshi` doesn't depend on the `puppeteer` jest env directly. I'd love to add a test for `puppeteer` but we thought maybe it's too much overhead. WDYT?

ranyitz:
@netanelgilad The build is failing because we don't have puppeteer in node modules.

I think that we can add puppeteer as a dev dependency of yoshi.

packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -36,9 +37,25 @@ module.exports = async () => {
     // start with a few new lines
     console.log('\n\n');
 
+    const forwardProxyPort = process.env.FORWARD_PROXY_PORT || 3333;
```
ronami: I didn't saw it being used by we might want to wrap it with `parseInt(xxx, 10)`

PR #740, opened 2018-11-06
improve index-dev.js and environment files

ranyitz:
### 🔦 Improvements summary
* Change management port to be `process.env.port + 1` instead of hard-coded `3001`.
* Add links to `wix-bootstrap-testkit` and `wix-config-emitter`
* explain the functions in `environment.js` file
* change `fn()` to `val()`, both work, but `val` should be used according to documentation.


ranyitz:
@netanelgilad Interesting idea regarding the JSDoc, I'll merge it for now and we could create another PR improving it (also in many other places in the codebase)

PR #741, opened 2018-11-07
Finish initial copying of static files before starting cdn

netanelgilad:
### 🔦 Summary
This is (hopefully) a solution to the flakiness of some of the tests for `yoshi start`. This makes sure we finish copying the static files before starting the CDN, thus preventing a race condition of calling the CDN before the files have finished copying and getting a `404`.

### 🗺️ Test Plan
Should help with the flakiness of current tests.


ranyitz:
@netanelgilad I think that if you've found a solution for the flakiness we can close this PR.

copy task should be much faster than webpack in 99% of cases.

PR #742, opened 2018-11-07
Presence of README.md allows to create a yoshi app

### LINKED ISSUE #541: Presence of `README.md` should allow to create a yoshi app
# 🙋 Feature Request

Presence of `README.me` should still allow me to create a yoshi app.

### 🤔 Expected Behavior

Filter out `README.md` from files check and **not override it**.

### 🔦 Context 

It's pretty common to create a repo on GitHub with a README.md file included / write a README.md file beforehand (... before creating the actual app, API first etc.). I don't see any point to throw an error in case of its existence.

Schniz:
### 🔦 Summary
Fixes #541.

Now, when a `README.md` is in the working dir, `create-yoshi-app`
wouldn't throw an error, and would ignore it - and *won't override
its contents.*

### 🗺️ Test Plan
Generate a project on a tempdir with README.md file in it, see that
content doesn't change :clap:

### 🤔 Open Questions
* Should we allow `README.md` files across all directories? `docs/README.md` for instance?
* Should we allow any markdown file? what about images? how can people know that README is allowed, but CONTRIBUTING isn't?
* Right now, It's implemented so every "allowed file", like any hidden file (`.gitignore`, for instance) or `README.md` is being *skipped*, for the sake of "one source of truth" - there's only one function that decides on which files are allowed to stay, and that's the `verifyWorkingDir` function. That means we don't override `.gitignore` on projects that already have `.gitignore`. Do you think it would cause an issue?

cc @roymiloh


packages/create-yoshi-app/__tests__/generateProject.spec.js
```diff
@@ -26,3 +27,32 @@ test('verify generation works as expected', () => {
...
+      return path.join(__dirname, './__fixtures__/fake-template/');
+    },
+  });
+
+  generateProject(answers, tempDir, ['README.md']);
+
+  expect(fs.readFileSync(readmeFilePath, 'utf-8')).toEqual(originalContent);
+
+  const files = getFilesInDir(tempDir);
+  expect(files).toMatchSnapshot();
```
Schniz: I wonder if I should leave this snapshot at all - because I don't need to test it (the other test already verifying we have the correct results),
what do you think? should I remove that?

packages/create-yoshi-app/src/verifyWorkingDirectory.js
```diff
@@ -1,16 +1,24 @@
...
-  if (!emptyDirectory) {
+  if (unallowed && unallowed.length > 0) {
     console.log(`The directory "${workingDir}" is not an empty directory\n`);
+    console.log(`Conflicting files: ${unallowed.join(', ')}`);
     console.log('Aborting...');
 
     process.exit(1);
   }
+
+  return allowed;
```
Schniz: Returning the "allowed" files is kinda messy, but I believe it works. If it was a pure function, it could return something like `Either error [allowedPath]`

ranyitz:
@Schniz Thanks for working on this!

> It's pretty common to create a repo on GitHub with a README.md file included / write a README.md file beforehand (... before creating the actual app, API first etc.). I don't see any point to throw an error in case of its existence.

IMHO the issue here is that sometimes people create a repository in github (before they call `create-yoshi-app`) which means they will have `README.md` + `.git` directory.

image

I don't mind overriding the readme file, it as it doesn't contain anything.

> Should we allow README.md files across all directories? docs/README.md for instance?

> Should we allow any markdown file? what about images? how can people know that README is allowed, but CONTRIBUTING isn't?

I don't think so.

> Right now, It's implemented so every "allowed file", like any hidden file (.gitignore, for instance) or README.md is being skipped, for the sake of "one source of truth" - there's only one function that decides on which files are allowed to stay, and that's the verifyWorkingDir function. That means we don't override .gitignore on projects that already have .gitignore. Do you think it would cause an issue?

These are edge cases, most projects will fall into one of those 3:
* `create-yoshi-app` is with `<name>` argument (will create a directory for them).
* Create an empty directory and generate inside.
* Create a github project, clone it and only then run `create-yoshi-app` inside (will maybe have a readme)

What do you think?

Schniz:
Yup, These 3 cases are the most relevant, I agree. I totally understand why this PR is important and sort of a "bug", but I think there's a potential bug/feature here:

https://github.com/wix/yoshi/blob/a988ea9e5678829108ec8c07966b65fa3cae0bac/packages/create-yoshi-app/src/verifyWorkingDirectory.js#L3

If we'll take a closer look on this regex,

```js
const hiddenFilesRegex = /(^|\/)\.[^/.]/;
hiddenFilesRegex.test("some/directory/but/then/.hidden_file") // => true
// so...
hiddenFilesRegex.test("src/.hidden_file") // => true. whooops
```

I guess we should change the regex as well, to only seek for hidden files/directories on the working dir root. Otherwise, it's weird to have two different behaviors (checking for an inner-most hidden files, but allowing only top-most README file)

ranyitz:
@Schniz I agree about the regex.

It used to be only for `.git` on the root, and it changed [here](https://github.com/wix/yoshi/pull/638/files#diff-973cfe15d8d4a8dc8a9eecd169953aab) I don't remember why.

We should ask @ronami

ronami:
Looks like it's a mistake, let's fix it 👍

Schniz:
Yeah so there's no actual bug because I forgot that `dir/.hidden` would have 2 entries in glob search: `dir` and `dir/.hidden`. Also, I think that we do `readdirSync`, so it's not a problem at all.

I'll make it so `.git` and `README.md` are the only files allowed with an easy equality check, if there are no other considerations

Schniz:
We've decided to close this PR for now:

* I have never had this problem with other tools (CRA, rails, etc)
* it makes the code messier
* it has open questions:
  * should we overwrite the README file? wouldn't it be surprising?
  * should we rename it? (like create-react-app does?)
  * should we just ignore and don't overwrite the file? what about when we'll have better README files with explanations about the tasks (`npm {start,test}`, etc)
* it's not hard to rename/delete the file yourself

Issue #730, opened 2018-11-05
Sometimes there is a detached jest process running a server which causes next runs to fail [jest-yoshi-preset] 


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

Sometimes, after running tests (using `jest-yoshi-preset`) there is an open server (bootstrap server) left open on a detached jest process. The next time we'll run the tests they will fail with an error.

![image](https://user-images.githubusercontent.com/11733036/47995292-7c350f80-e0fd-11e8-9728-e2acb3a3ba81.png)

The error indicates that there is a server running on port `1301` and that we are trying to run another server on the same port.

### 🤔 Expected Behavior

1. There should be no detached jest processes after running tests.
2. Our tests should work normally even if there is a detached process with a running server.

### 🔦 Context

It's really annoying to work with it as the user needs to kill the detached node processes from time to time (If the user understands that this is the problem).

### 💁 Possible Solution
We can separate this bug into three different problems:
1. The environment teardown function (that was supposed to close the server isn't called in some cases).

> We should find all the cases that it happens and solve them, for example, add a feature in jest that can listen to `SIGINT` and wait for environment teardown etc...
> There is a [related issue in jest](https://github.com/facebook/jest/issues/6029)

2. Sometimes there is a detached child process (with a running server) after running tests.

> We can open an issue in jest about the detached processes problem, jest should be able to close its workers.
> We may be able to add a `SIGINT`/`exit` listener from `jest-yoshi-preset`'s code and exit all the child processes. 

3. If we have a server running on a certain port (with is the same one that the use intend to run a server on), our tests will fail with an unclear message.

> The user runs the server, so it would be hard to build a fault-tolerant strategy from there.

> We provide the user with a port using the `getPort()` function. We can Verify that this port is available and provide the user only with available ports.

### 👣 Steps to Reproduce

Generate a new project using `create-yoshi-app`
```sh
npx create-yoshi-app test
```

Run e2e tests and press `ctrl+c` after the initial setup.

```sh
npx yoshi test --jest --watch e2e
```

![jest-yoshi-bug](https://user-images.githubusercontent.com/11733036/47995826-4d1f9d80-e0ff-11e8-9647-07fdce8a6973.gif)

The next time you'll run your tests, you'll see the bug

## Summary

I think that we should start by changing the `getPort()` function to provide with a free port, because this would be the fastest + easiest solution.

We should, of course, look into solving the other bugs, so the user won't have detached processes on its machine.

[...]

PR #743, opened 2018-11-07
Kill process from `getPort()` only if it belongs jest

yavorsky:
### 🔦 Summary
This PR is an follow-up for https://github.com/wix/yoshi/pull/731 that should resolve https://github.com/wix/yoshi/issues/730 issue
For now, we are checking port for availability and then just killing it. But it could be possibly everything. So after this PR we are going to check that process we want to kill is a `jest` process.
Again, it's not the best soltuion for this issue and we need to send PR directly to jest ideally.

But for now, we at least have less chances to kill NASA's flight control system served on some port. 😨

Also, utils in jest-environment-yoshi-puppeteer was refactored, since it used most of already implemented in yoshi-helpers methods.


packages/yoshi-helpers/utils.js
```diff
@@ -111,21 +111,35 @@ module.exports.getProcessIdOnPort = port => {
...
+    })
+    .trim();
+};
 
-  if (port === portTestResult) {
-    return null;
-  }
+module.exports.processIsJest = pid => {
+  const commandArg = getCommandArgByPid(pid, 1);
+  return commandArg.split('/').pop() === 'jest';
```
Schniz: this is a wild trick! 👏

PR #745, opened 2018-11-07
Change prettier errors to warnings instead of errors

### LINKED ISSUE #744: New versions of prettier break project when linting
# 🐛 Bug Report

### 😯 Current Behavior

When `prettier` releases a new version and change the way the format some code, our projects get broken in CI because lint not fails with errors from `prettier`. We depend on `prettier` with `^` so we get new minors from `prettier` automatically, and minors in `prettier` can change the formatting rules (this isn't a breaking change for `prettier` because they are just a formatter, but it is a breaking change for us because we use eslint, and emit an errors on prettier issues).

### 🤔 Expected Behavior

Projects builds should not fail because of new `prettier` versions.

### 🔦 Context 

Release 1.15 of `prettier` caused a lot of project build to fail today.

### 💁 Possible Solution 

Two possible solutions:
1. Make prettier errors into warning instead of errors. That way we can keep getting minor updates from `prettier` without failing.
2. Pin `prettier` version to only patches with a `~`.
@ronami @ranyitz @yanivefraim @roymiloh Thoughts?

netanelgilad:
### 🔦 Summary
Fixes #744.


ronami:
Let's add this to our tslint config too.

PR #750, opened 2018-11-08
Lint Stylesheets by default

### LINKED ISSUE #661: Lint Stylesheets by Default
# 🙋 Feature Request

### 😯 Current Behavior

We do not currently have linting for stylesheets.

### 🤔 Expected Behavior

We should have linting for stylesheets!

### 💁 Possible Solution 

https://stylelint.io/ is valid solution.

Schniz:
### 🔦 Summary
Fixes #661.

I've added `stylelint-config-yoshi`, a default configuration to `stylelint` for yoshi projects. It has `prettier` already baked in to make code look nicer. `yoshi lint --fix` now also run, lints and fixes against prettier, making stylesheets more maintainable (I hope).

I have left all the [possible-errors rules](https://stylelint.io/user-guide/rules/#possible-errors) and commented out everything I thought would just be annoying and not buggy, like `font-family-no-missing-generic-family-keyword`: would throw an error if we use a custom web font. I guess that it's a good practice to have `Arial, serif`, but it's not auto-fixable, so it's a problem.

The only two rules that are auto-fixable are `no-extra-semicolons` and `prettier/prettier`. So we need to beware!

### 🗺️ Test Plan
I have added a test to lint a file with the new stylelint configuration (`stylelint-config-yoshi`)

### 🤔 Open Questions
* Please take a look at `stylelint-config-yoshi`. I think we should be only adding rules that can cause bugs and that aren't annoying. I would love to get feedback.
* Just like #745, prettier is configured as `warning`. This throws a warning in `stylelint-prettier`, and I have sent a PR to fix this (https://github.com/prettier/stylelint-prettier/pull/10)
* I believe [we should have a wrapper around the three linters](https://github.com/wix/yoshi/issues/661#issuecomment-436909858), to have a unified experience
* `yoshi lint --fix` now fixes stylesheets, I dropped the usage of `haste-stylelint`.
* Templates now have `stylelint` configuration in their `package.json`. It is needed by IDE plugins (just like `eslintConfig` in CRA). Do we need to write it down somewhere? Docs? Migration path?


Schniz:
image


pretty

Schniz:
@ranyitz my PR to `prettier-stylelint` has been merged. can you please review again and hopefully merge? 😸

PR #752, opened 2018-11-11
Increase max retries for tests, so they'll pass locally

yurynix:
### 🔦 Summary
Currently the tests rely on retries with a backoff to test CDN ( and other stuff, but lets focus for now just on CDN ).
To account for that, this PR increases the retries.

This PR also makes `insideTeamCity` the default environment 

| Before  | After  |
|---|---|
| screen shot 2018-11-11 at 11 30 00  | screen shot 2018-11-11 at 11 37 01  |

### 🗺️ Test Plan
Do we test the tests? ¯\\_(ツ)_/¯


ranyitz:
@yurynix What about [`waitPort`](https://github.com/dwmkerr/wait-port#readme)? we already use it in yoshi - [here](https://github.com/wix/yoshi/blob/8ff47958e1a22fa797f39777cffdeb080b8891f0/packages/jest-environment-yoshi-puppeteer/globalSetup.js#L97) and [here](https://github.com/wix/yoshi/blob/971634d6d427cb7a84a21bc3ab6193cca24e926d/packages/yoshi/src/webpack-utils.js#L170-L176)

yurynix:
Thank you for the suggestion @ranyitz , I've modified the code to use waitPort when appropriate.
IMO, There is still a use-case for `waitForOutput` though, here and in https://github.com/wix/yoshi/pull/762

test-helpers/test-phases.js
```diff
@@ -45,6 +46,37 @@ class Test {
...
+      reject = pReject;
+    });
+
+    this.outputWaiters.push({
+      outputString,
+      resolve,
+      reject,
+    });
+
+    setImmediate(() => this._checkOutputForWaiters());
```
ranyitz: remove setImmediate

test-helpers/test-phases.js
```diff
@@ -45,6 +46,37 @@ class Test {
...
+          if (this[stream].indexOf(waiter.outputString) > -1) {
+            waiter.resolve(stream);
+            return true;
+          }
+          return false;
+        }),
+    );
+  }
+
+  waitForOutput(outputString) {
```
ranyitz: If we only call it once per TestPhase, maybe we can drop the outputWaiters array,

packages/yoshi/test/start.spec.js
```diff
@@ -815,45 +827,60 @@ describe('Aggregator: Start', () => {
   }
 
   function checkStdout(str) {
-    return retryPromise(
-      { backoff: 100 },
-      () =>
-        test.stdout.indexOf(str) > -1 ? Promise.resolve() : Promise.reject(),
-    );
+    return test.waitForOutput(str);
```
ranyitz: just call `test.waitForOutput(str)`

test-helpers/test-phases.js
```diff
@@ -45,6 +46,37 @@ class Test {
     return this;
   }
 
+  _checkOutputForWaiters() {
```
ranyitz: Maybe it's better to show each case (`stdout`/`stderr`)

PR #756, opened 2018-11-11
Add tests for generated project local environment (`npm start`)

### LINKED ISSUE #546: Add e2e tests for generators local development
# 🐛 Bug Report

### 😯 Current Behavior

Currently, on our e2e tests, we only test that running `npm build`/`npm test` work correctly. If something breaks local development we don't know about it.

### 🤔 Expected Behavior

We're going to make changes to local development and we should be able to know (when we make a PR) that we didn't break local development.

### 💁 Possible Solution 

We can add an e2e test (`create-yoshi-app`) that will run `npm start` and poll requests to `http://localhost:3000`. Once it returns a 200 request, match it against the previous snapshot.

yanivefraim:
Fixes #546

### 🔦 Summary
Added tests for generator's `npm start` command.

For now, I only test two things:
1. A call to the server (port 3000) returns 200.
2. A call to cdn server (port 3200) returns 200.

Concerns: 
1. Ports are hardcoded (tests will fail in case ports are taken and we use different ports)
2. Not sure if we want to test stuff on the actual request (is checking for 200 enough?) 

**cdn tests are failing for generators with no static files to serve (server generator, for example). My opinion about this:**

Step 1: Remove CDN test and merge this PR.
Step 2: Add a Puppeteer test which renders `http://localhost:3000`. This is nice because we can actually assert page title and know for sure that React works, including i18n, etc. We can also check for console errors. The problem: different generators produce different outputs, so we should still somehow run a different test for the different generators.
We can also combine this with moving e2e tests to Jest, and compare the output of `http://localhost:3000` with snapshot. This is nice because you don't care about generator type

### DIFF
--- a/packages/create-yoshi-app/package.json
+++ b/packages/create-yoshi-app/package.json
@@ -53,6 +53,7 @@
     "mocha-env-reporter": "^3.0.0",
     "puppeteer": "^1.0.0",
     "tempy": "^0.2.1",
+    "wait-port": "^0.2.2",
     "yoshi-helpers": "3.19.0"
   }
 }
--- a/packages/create-yoshi-app/package.json
+++ b/packages/create-yoshi-app/package.json
@@ -51,6 +51,7 @@
     "minimist": "^1.2.0",
     "mocha": "^5.2.0",
     "mocha-env-reporter": "^3.0.0",
+    "puppeteer": "^1.0.0",
     "tempy": "^0.2.1"
   }
 }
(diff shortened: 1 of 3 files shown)


yanivefraim:
After discussing this with @ranyitz and @ronami, we decided to split this PR:

1. Run Puppeteer test, which checks for console errors/warnings (instead of the current request for port 3000 which checks for 200 response)
2. Run `npm start` and then `jest`, for the generated project. This will run the generated project tests using start's bundles. We will run it only on full-stack/client projects, for now.

ronami:
@yanivefraim can this be merged? is there still work on it?

ronami:
@yanivefraim?

PR #768, opened 2018-11-12
Add unit testing infra to `yoshi` package

### LINKED ISSUE #767: Unit testing in yoshi package
# 🙋 Feature Request

### 😯 Current Behavior

We don't have an infrastructure of unit testing.

### 🤔 Expected Behavior

Add `jest` for adding unit tests to yoshi

### 💁 Possible Solution 

1. Add `jest` to the party for running unit tests, and use `mocha` for running our integration tests, until https://github.com/facebook/jest/issues/6616 is fixed.
1. Change mocha's glob pattern to `**/*.mocha.js`, and rename all existing files. new tests can be written in `**/*.jest.js` files. Choosing `*.spec.js` and `*.test.js` can confuse newcomers - using the test driver in the file name will make it easier to use (well, that's my opinion)
1. Make a script (`bin/test.js`) that accepts files and matches them against the two globs (both `mocha`'s and `jest`'s), then runs the tests:
   * `bin/test.js` will run all tests
   * `bin/test.js file.mocha.js` would run mocha for `file.mocha.js`
   * `bin/test.js file.jest.js` would run jest for `file.jest.js`
   * `bin/test.js file.mocha.js file.jest.js` would run jest for `file.jest.js`, then mocha for `file.mocha.js`

### 🔦 Context 

We currently use `mocha` to drive our integration tests, and hopefully we'll move towards `jest` in the near future. We don't have any unit tests in our `yoshi` package, causing us to make slow integration tests and skipping edge cases due to permutation complexity

Schniz:
### 🔦 Summary
Fixes #767 

* I have added `jest` as a (unit) test runner
* I renamed `*.spec.js` files to `*.mocha.js` files, so it would be easier to know which runner runs which file
* I have created a script (`bin/test.js`, which is also `npm test`), that handles running the tests:
  * `bin/test.js` will run all tests
  * `bin/test.js weird_file.js` will do nothing but complaining about an ignored file (it doesn't match any test runner file convention!)
  * `bin/test.js file.mocha.js` would run mocha for `file.mocha.js`
  * `bin/test.js file.jest.js` would run jest for `file.jest.js`
  * `bin/test.js file.mocha.js file.jest.js` would run jest for `file.jest.js`, then mocha for `file.mocha.js`
  * `bin/test.js --watch` throws an error, you can't watch 2 test runners!
  * `bin/test.js --only jest --watch` will watch jest files (similar to `jest --watch`)
  * `bin/test.js --only mocha --watch` watches on mocha (but these tests are sllllow)

This `bin/test.js` file and naming conventions (`*.jest.js`), can sound weird, but should be easy as long as we have 2 test runners. This makes tests easier to approach (it is almost like a shebang).

What do you think?

### 🗺️ Test Plan
It is all about testing 💃 
Maybe we can split it to be parallel in CI?


Schniz:
Maybe the script will be beneficial in the future

PR #776, opened 2018-11-13
added source map loader

amiryonatan:
### 🔦 Summary
Added source map loader under configuration


amiryonatan:
@ranyitz

amiryonatan:
@ranyitz Added loader for ts. 
I would really appreciate some help with the specs

PR #777, opened 2018-11-13
Testing infra for experimental server bundle

ronami:
### 🔦 Summary

Still a work in progress. Will update. 👷


test/utils.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const content = (await request(url)).replace(/\s/g, '');
+
+  for (const regex of regexes) {
+    expect(content).toMatch(regex);
+  }
+};
+
+const initTest = async feature => {
+  await page.goto(`http://localhost:3000/${feature}`);
```
ranyitz: This should probably be in constants.

PR #778, opened 2018-11-13
wrong file name format in the md template of business-manager-module

liorMar:
### 🔦 Summary
...

### 🗺️ Test Plan
...


ranyitz:
@liorMar Can you provide some explanation?

liorMar:
in the create-yoshi-app templates of "business-manager-module-typescript" and "business-manager-module" there is an auto generated md file with a link to "templates/module_{%projectName%}.json.erb"
but the real file is created as project name with capital letters {%PROJECT_NAME%} so the link is broken.

Issue #672, opened 2018-10-25
Values in the cli reference are not searchable


ranyitz:
# 🐛 Bug Report
Currently you can't search for cli options, for example, type `--ssl` or `ssl` and you won't find relevant entries, even though we do have doc which is placed in the [CLI options section](https://wix.github.io/yoshi/docs/api/cli#start).

A good solution would be to use headers (h1/h2/h3) for those cli options, so they would be searchable.

a good example can be found on [jest's site](https://jestjs.io/docs/en/cli#options)

yurynix:
I actually haven't noticed that issue has a PR already...
Anyway, I think it's maybe better to keep to markdown semantics and use `###` style headlines instead of explicit `Hn` tags, what you think?

PR #783, opened 2018-11-14
Use headlines instead of table for start options

yurynix:
### 🔦 Summary
Attempt at #672

### 🗺️ Test Plan
Read the docs, see the cli start option, such as `--ssl` is linked and you can get direct link


yurynix:
Thank you for the review @ranyitz , that's a great idea, i'll update the PR.

PR #784, opened 2018-11-14
Add source map loader yoshi2

amiryonatan:
### 🔦 Summary
Added source map loader to yoshi 2


amiryonatan:
@ranyitz tested locally and it works

PR #790, opened 2018-11-15
Remove server-handler, use built-in webpack-dev-server instead

yurynix:
### 🔦 Summary
Remove server-handler, use built-in webpack-dev-server instead with `contentBase` `watchContentBase`

screen shot 2018-11-15 at 9 58 49

### 🗺️ Test Plan
Create a new app via `npm run create-yoshi-app:dev`
Modify it's `package.json` to include `"experimentalServerBundle": true,` for using the new `start-app` task that uses the `webpack-dev-server`

Make sure that you see a file's index on the CDN http://localhost:3200


packages/yoshi/package.json
```diff
@@ -96,7 +96,7 @@
     "react-hot-loader": "^4.3.3",
     "rtlcss-webpack-plugin": "^4.0.0",
     "ruby-haml-loader": "~1.0.0",
-    "serve-handler": "^5.0.5",
+    "serve-index": "^1.9.1",
```
yurynix: This is a dependency of [WDS](https://github.com/webpack/webpack-dev-server/blob/master/package.json#L48)

test/kitchensink/jest.build.config.js
```diff
@@ -7,7 +7,6 @@ module.exports = {
     '**/loaders.test.js',
     '**/moment.test.js',
     '**/optimize.test.js',
-    '**/statics.test.js',
```
yurynix: Not sure why build step needs to check static assets serving? @ronami what am I missing?

yurynix:
After discussing that with @ronami , we decided to drop that PR.

PR #805, opened 2018-11-18
HtmlWebpackPlugin integration

netanelgilad:
### 🔦 Summary
Still WIP, a place for me to start discussions on implementation. I'll update this comment when ready.

To complete:
- [x] support for index.ejs
- [x] support for index.vm
- [x] support for zero conf - no template file at all
- [x] support for configuring which entries go into which templates
- [ ] opt in support for rtl css
- [ ] works with split chunks
- [ ] ? support for disabling injection and full template control to user
- [ ] works in local development
- [ ] works in production mode
- [ ] ? support for favicon and manifest.json out of the box

### 🗺️ Test Plan
...


netanelgilad:
I'm thinking about how to implement a solution for choosing which chunks go into which templates.
There are a few questions to answer about it:
* How should it work with zero-conf? By matching template names to entry names? That sounds like the simplest solution.
  * If they should match by name, how should we handle no template at all? should we output it as `app.ejs` instead of `index.ejs`? Since the default entry is `app`. 
* How do you specify it manually in `yoshi` config? I'm thinking of adding a `templates` section to the config that matches a template path to the entry that should go in there. Example:
```json
{
  "templates": {
    "first.ejs": "first",
    "second.ejs": "second"
  }
}
```
But maybe putting file paths in the key is annoying? Any thoughts?

ronami:
Closing in favor of #1090. Please re-open if still relevant.

PR #807, opened 2018-11-18
Add documentation for stylesheet linting

Schniz:
### 🔦 Summary

Adds docs for #661 and #750 

### 🤔 Other Considerations

I haven't added `stylelint-yoshi-config` as a dev dependency to the templates in #750. I'm running `stylelint` from `yoshi lint`, and yoshi has this preset as a dependency, so it isn't a problem running from yoshi. We will only face problems in the future with something like `tink` or `yarn plug n' play`, and it will happen in IDE-land only. wdyt?


Schniz:
trying out with a fresh branch

docs/api/cli.md
```diff
@@ -236,7 +236,7 @@ Default: `false`
 
 #### `--format`
 
-Use a specific formatter for eslint/tslint
+Use a specific formatter for eslint/tslint/stylelint
```
ronami: I'm not sure how well this option is when it's being passed to multiple tools. Only values that exist in both tools (eslint and stylelint or tslint and stylelint) will work. I see very little usecase for this option, maybe remove it? wdyt @ranyitz @netanelgilad @Schniz?

docs/api/cli.md
```diff
@@ -246,7 +246,11 @@ Optional list of files (space delimited) to run lint on
 
 Default: empty
 
-Executes `TSLint` or `ESLint` (depending on the type of the project) over all matched files. An '.eslintrc' / `tslint.json` file with proper configurations is required.
+Executes linters based on the project type:
+
+- `TSLint` for TypeScript projects (a `tslint.json` configuration file is required)
```
ronami: It's not related to this PR but I think it would be nice to have links to ESLint/TSLint/Stylelint as some people might now know what Stylelint is.

docs/guides/stylelint-setup.md
```diff
@@ -0,0 +1,32 @@
...
+id: stylelint-setup
+title: Stylelint Setup
+sidebar_label: Stylelint Setup
+---
+
+## Introduction
+
+Yoshi defines [`stylelint-config-yoshi`](https://github.com/wix/yoshi/tree/master/packages/stylelint-config-yoshi), which is a custom [Stylelint configuration](https://stylelint.io/user-guide/configuration/) to enable zero-configuration stylesheet linting for any application. It follows the following mindset:
+
+- Focus on possible errors rather than stylistic opinions
```
ronami: I would emphasis on our intent here, how about something like: `Only fail on errors that can cause production bugs rather than stylistic opinions`?

docs/guides/stylelint-setup.md
```diff
@@ -0,0 +1,32 @@
...
+- Auto-fixable code-style with prettier (`yoshi lint --fix`)
```
ronami: Here too, how about: `If it's not a critical bug, we try to only add auto-fixable rules`?

docs/guides/stylelint-setup.md
```diff
@@ -0,0 +1,32 @@
+---
+id: stylelint-setup
+title: Stylelint Setup
+sidebar_label: Stylelint Setup
+---
+
+## Introduction
+
+Yoshi defines [`stylelint-config-yoshi`](https://github.com/wix/yoshi/tree/master/packages/stylelint-config-yoshi), which is a custom [Stylelint configuration](https://stylelint.io/user-guide/configuration/) to enable zero-configuration stylesheet linting for any application. It follows the following mindset:
```
ronami: I would start by explaining why users would want to configure it: How it can help that catch bugs and keep the styles consistent.

docs/api/cli.md
Schniz: Are you thinking about removing the `format` option or to remove the fact it formats stylelint as well?

PR #808, opened 2018-11-18
Update test globs

### LINKED ISSUE #717: Change test globs to work according to file extension, instead of file location
# 🙋 Feature Request

### 😯 Current Behavior
Different test types are triggered according to their location. For example, all spec files inside `test/e2e` folder are initialized using puppeteer env. Same `src/components` - JSDOM env. 

### 🤔 Expected Behavior

I would like to write my e2e tests in the same directory as the component (for example), without being dependent on its location.

### 💁 Possible Solution

Change globs to be according to file extention:

test.e2e.js - puppeteer
test.spec.js - JSDOM
test.server.js - bootstrap env

### 🔦 Context

- I think that it is better to have all component files (tests / css etc.) in the same directory. I also think that Yoshi shouldn't interfere with where I want to put my test files.

- Searching for a test file (app.spec.js) will result in more than one file (both e2e / component test files have the same name), which makes it harder to navigate.

- Currently, we do not have a good place to put our client unit test files. The only available location is inside `src/components` folder. This makes me move unit tests and their associated production code into `src/components` folder. Instead of creating a new environment for unit tests, we can take advantage of JSDOM environment and use `src/service.spec.js`, which will just work.

ranyitz:
### 🔦 Summary

This PR fixes #717.

We've talked about the current tests glob patterns (@ronami @netanelgilad @Schniz @yanivefraim) And we all agreed that we should not interfere with the way the user chooses to structure his tests.

We've decided on two testing environments:

* `e2e` - An environment that will bootstrap the server and a puppeteer page.
* `spec` - Regular unit and component test which will run using JSDOM environment.

We'll remove the `server` test environment as it adds complexity while server tests are actually e2e tests and can be tested using the e2e environment.

It means that migration would be relatively easy:

* Change all tests which ends with `*.spec.js` in `e2e` and `server` directories into `*.e2e.js` files.

_____________

### jest-yoshi-preset setup files change [breaking]

We've realized that the setup file of yoshi was `setup.e2e.js` and now collides with the new `e2e` tests glob pattern, so we've decided on changing the pattern of the setup files to `[envirnoment]-setup.js/ts` (`e2e-setup.js`)

### Changes in templates

* Change the directory `test` -> `__tests__` which places it at the top and it's easy to identify because of the `__` prefix and suffix.

* Remove the `test/helpers` directory, and move the `i18n.mock.js` into a `__mocks__` directory, we believe that `helpers` is to abstract and we prefer something like `__mocks__` if we know in advance that it's the only usecase, also, it didn't make sense to go from `src/components` directory all the way to `test/helpers`.

* `environment.js` file is a module that is being reused from both `index-dev.js` and `jest-yoshi-preset.js` files, both on the root, so we've moved it from `test` directory to the root of the project. We believe that it will make the `__tests__` directory easier to grasp when getting to the project (because it has less files), also, the environment is not just for testing, also for dev. Maybe it makes sense to create a dev directory, and put `index-dev.js` `envirnoment`  and more files there, but for now, this PR moves it to the root of the project.


docs/guides/jest-setup.md
```diff
@@ -80,27 +80,19 @@ Yoshi serves the files from `dist/statics`. Make sure to run `npx yoshi build` b
...
+### JSDOM (spec)
 
-Sets up a standard [JSDOM](https://github.com/jsdom/jsdom) environment for component tests.
+Sets up a standard [JSDOM](https://github.com/jsdom/jsdom) environment for component and unit tests.
 
-It's configured for every file under `<rootDir>/src/**/*.spec.js`.
-
-### Bootstrap (server)
-
-An environment for testing your server (API) code. It starts up a different instance of your server ([wix-ng-bootstarp based](https://github.com/wix-platform/wix-node-platform)) for every test file.
```
ranyitz: should probably move this into the Puppeteer environment explanation

ranyitz:
# What's left to be done?

__2 tests from `test:integration` seem to be failing:__

* `should pass with passing e2e tests`
* `should support dynamic imports when running e2e tests in a CI build`

__Changes in `create-yoshi-app's` templates__

* All templates needs to use the new globs.

docs/guides/jest-setup.md
yanivefraim: @ranyitz - I agree. I also think that it might be confusing to call it "Puppeteer" environment since this is the place to write server e2e tests. 
Maybe we could think of different terminology, something like:

### '.spec' - component and unit tests. Sets up JSDOM.
### '.e2e' - Browser and Server e2e tests. Starts up a server instance (bootstrap) and creates a Puppeteer page for that test.

WDYT?

docs/guides/jest-setup.md
```diff
@@ -135,10 +127,9 @@ module.exports = {
 
 ### Setup Files
 
-If you want to run some code before your tests you can use one of the 3 following setup files (1 for each environment):
+If you want to run some code before your tests you can use one of the 2 following setup files (1 for each environment):
 
-- `<rootDir>/test/setup.component.(j|t)s`: JSDOM (component)
-- `<rootDir>/test/setup.server.(j|t)s`: Bootstrap (server)
+- `<rootDir>/test/setup.spec.(j|t)s`: JSDOM (spec)
 - `<rootDir>/test/setup.e2e.(j|t)s`: Puppeteer (e2e)
```
yanivefraim: Should be:

`<rootDir>/test/spec-setup.(j|t)s`: JSDOM (spec)
`<rootDir>/test/e2e-setup.(j|t)s`: Puppeteer (e2e)

I also think we should describe it a little different. Maybe:

`<rootDir>/test/spec-setup.(j|t)s`: Setup for '.spec.' tests (Component and Unit tests)
`<rootDir>/test/e2e-setup.(j|t)s`: Setup for '.e2e' tests (Browser an Server e2e tests)

docs/guides/jest-setup.md
ranyitz: Yes, I agree.
Moving from the specific term puppeteer will help us transferring the message of this PR.

docs/guides/jest-setup.md
ranyitz: Sound good!

yanivefraim:
@ranyitz @ronami - I moved all (jest) generators to the new globs pattern. I also changed mocha generators to use `--mocha` (just not to break them). Notice that I did not manually test all generators (maybe we should, for "old" untested generators)

packages/jest-environment-yoshi-puppeteer/utils.js
```diff
@@ -1,9 +1,10 @@
 const globby = require('globby');
+const { e2eTests } = require('yoshi-config');
 
 const { MATCH_ENV } = process.env;
 
 module.exports.shouldRunE2Es = async () => {
-  const filesPaths = await globby('**/*.e2e.(ts|js){,x}');
+  const filesPaths = await globby(e2eTests);
```
ranyitz: 👍

docs/guides/jest-setup.md
```diff
@@ -78,29 +78,17 @@ npx yoshi test --jest
 
 Yoshi serves the files from `dist/statics`. Make sure to run `npx yoshi build` before you run the tests using this mode.
 
-## Environments
+## Test types
 
-### JSDOM (component)
+### `*.spec.js` files
```
ronami: I think the title should tell the type of tests instead of how they're matched.

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -1,7 +1,11 @@
 const fs = require('fs');
 const globby = require('globby');
 const { envs } = require('./constants');
-const globs = require('../yoshi-config/globs');
+const globs = require('yoshi-config/globs');
+
+const modulePathIgnorePatterns = [
```
ronami: We'll rebase that from master

PR #820, opened 2018-11-19
fix downgrading node to version 8 for version that are grater then 8

liorMar:
there is a bug when every time i want to use node version greater then 8
the update-node-version change my version to 8 because it is comparing unformatted string (line 6 not using 'utf8') '8' with another string instead of numbers compare for example i want to use node 10
and '8' is not greater than '10'
...

there is no tests for tasks
...


ronami:
Thanks for the PR, it's already resolved by #836 though.

PR #837, opened 2018-11-20
Migrate templates to scoped packages

Schniz:
### 🔦 Summary
Migrating templates to scoped packages (`@wix/...`)

### 🗺️ Test Plan
Making sure e2e tests run with the public npm registry


roymiloh:
@Schniz better to order deps alphabetically, so next `npm install` won't reorder them (just redundant noise). Maybe you can automate this process somehow.

PR #840, opened 2018-11-20
fix killing of server process and its children, for travis-ci

yanivefraim:
### 🔦 Summary
On Linux, child processes of child processes will not be terminated when attempting to kill their parent (see [here](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal))

In [globalTeardown](https://github.com/wix/yoshi/blob/master/packages/jest-environment-yoshi-puppeteer/globalTeardown.js#L13) we kill spawned server - this will not kill its child processes, causeing some tests to hang on travis-ci...

This PR tries to solve the problem using what was suggested [here](https://azimi.me/2014/12/31/kill-child_process-node-js.html)


.travis.yml
```diff
@@ -14,6 +14,7 @@ branches:
   only:
     - master
     - /^version_\d+\.x$/
+    - 796-j
```
netanelgilad: need to remove this before merge.

yanivefraim:
plz do not merge, yet... I'm trying to refactor something...

yanivefraim:
I ended up using `killSpawnProcessAndHisChildren` for killing child process and its children

packages/jest-environment-yoshi-puppeteer/globalTeardown.js
```diff
@@ -2,6 +2,7 @@ const fs = require('fs-extra');
 const { WS_ENDPOINT_PATH } = require('./constants');
 const { shouldRunE2Es } = require('./utils');
 const cdnProxy = require('./cdnProxy');
+const { killSpawnProcessAndHisChildren } = require('yoshi-helpers');
```
ronami: how about removing `killSpawnProcessAndHisChildren ` and using https://github.com/dwyl/terminate instead?

yanivefraim: done...

PR #843, opened 2018-11-21
Puppeteer: tearing down Parent Environment and no throw on `pageerror`

liorMar:
### 🔦 Summary
added missing parent environment teardown invoke
Page errors: instead of error throw, changed to console.warn (for business-manager template)
e2e test are failing in business-manager because of unrelated errors.
not sure if this is the best solution for the page errors i'll be happy for suggestions. 

### 🗺️ Test Plan
...


packages/jest-environment-yoshi-puppeteer/index.js
```diff
@@ -30,11 +30,12 @@ module.exports = class PuppeteerEnvironment extends ParentEnvironment {
     this.global.page = await this.global.browser.newPage();
 
     this.global.page.on('pageerror', error => {
-      throw error;
+      console.warn(`Puppeteer page error: ${error.message}`);
```
yanivefraim: +1 for changing the failure with a warning. @ronami - wdyt? BTW - is it considered a breaking change?

I would add `error.stack` as well. I think that it is empty most of the time, but just for the chance that it'll have some data.

liorMar: @yanivefraim done

yurynix:
IMO documenting that well in the changelog is enough.

PR #762, opened 2018-11-12
[WiP] Use webpack-dev-server for CDN

yurynix:
### 🔦 Summary
Attempt at #753 

This PR:
* Uses code from https://github.com/wix/yoshi/pull/752 to make the tests stable.
* Modifies the tests that are testing `"hmr": true` (of `webpack-hot-client`) to test `/webpack/hot/` instead.
* Moves code from `start-app` to `webpack-util` to be later used in both the CDN and start-app
* Removes `serve-handler` dependency in favour of `webpack-dev-server` `contentBase`

screen shot 2018-11-14 at 15 26 59

### 🗺️ Test Plan
* Verify all tests are passing.
* Use with existing project (I've tested with `form-builder` and client only project with `create-yoshi-app`)

PR #844, opened 2018-11-21
Move hmr to webpack config

yurynix:
### 🔦 Summary
This PR moves configuration to webpack.config, and creates a function to create the WDS, to be used later at #762

### 🗺️ Test Plan
Run `npm run test:app-flow`


packages/yoshi/config/webpack.config.js
```diff
@@ -116,6 +115,22 @@ function overrideRules(rules, patch) {
   });
 }
 
+function addEntry(config, hotEntries) {
```
ronami: Maybe put that in `webpack-utils.js`? That way this code will mainly be webpack config logic.

packages/yoshi/src/commands/start-app.js
```diff
@@ -142,10 +121,10 @@ module.exports = async () => {
   console.log(chalk.cyan('Starting development environment...\n'));
 
   // Start up webpack dev server
-  await new Promise((resolve, reject) => {
-    devServer.listen(project.servers.cdn.port, '0.0.0.0', err =>
-      err ? reject(err) : resolve(),
-    );
+  const devServer = await createDevServer(clientCompiler, {
```
ronami: I think it may be better having one function to create the webpack dev server and then start it (call `listen()`) separately. I think it will make it easier to understand as calling `createDevServer()` and having it listen on a port is unexpected. Wdyt?

packages/yoshi/config/webpack.config.js
yurynix: You're right - I'm moving it back.
For some reason I thought they're in different packages.

yurynix:
Thank you for reviewing @ronami ! 🙇

packages/yoshi/src/commands/start-app.js
```diff
@@ -92,39 +90,20 @@ module.exports = async () => {
   const clientConfig = createClientWebpackConfig({
     isDebug: true,
     isAnalyze: false,
+    hmr: true,
```
ronami: I didn't notice it the first time but I think we should call it `isHmr`. It makes it easier to understand that it's a boolean (same as the other flags).

packages/yoshi/src/commands/start-app.js
ronami: I understand that we'll duplicate a bit of logic (calling listen both in `start` and `start-app`) but I think that it's worth it as it makes the code much easier to understand. It makes sense to me to create everything needed to run a server and being able to run it in a later time.

I prefer that we keep it as two separate functions, even if we have some duplication. If we'll have another usage or that this logic grows we can look into refactoring.

yurynix:
@ronami yep, thanks. I were just waiting for the CI to pass

PR #849, opened 2018-11-22
Generate port by checking it's actual availibility for listen()

yurynix:
### 🔦 Summary
We shouldn't kill random processes that guilty of occupying a port,
it can lead to unexpected effects on the user. it might be a legitimate program listening on that port. 

[get-port](https://github.com/sindresorhus/get-port/blob/master/index.js) is really tiny, we might want to just copy it inside as-is, wdyt?

### 🗺️ Test Plan
...


ranyitz:
@yurynix I agree with the general idea. Notice that this would be a breaking change for users who use the sync version of `getPort`.

Please see #743. It seems to do the job.

PR #850, opened 2018-11-22
Migrate business manager template to puppeteer

liorMar:
### 🔦 Summary
migrate e2e test runner from protractor to puppeteer.
before this PR could be merged. resolve this [PR](https://github.com/wix/yoshi/pull/843)

### 🗺️ Test Plan
no further changes to the original test was made except increasing of the timeout from 180000 to 200000


packages/create-yoshi-app/templates/business-manager-module-typescript/jest-yoshi.config.js
```diff
@@ -0,0 +1,56 @@
+const { environment } = require('./test/environment');
+const getMyPort = require('get-port');
+
+const getMultiplePorts = numOfPorts =>
+  Promise.all(new Array(numOfPorts).fill(undefined).map(() => getMyPort()));
+
+const getTwoConsecutiveAvailablePorts = async () => {
```
yurynix: This can be removed after https://github.com/wix-private/business-manager/pull/980

packages/create-yoshi-app/templates/business-manager-module-typescript/{%packagejson%}
```diff
@@ -8,67 +8,70 @@
...
+      "@wix/business-manager": "latest",
+      "@wix/business-manager-api": "latest",
+      "enzyme": "~2.9.0",
+      "get-port": "^4.0.0",
+      "husky": "~0.14.0",
+      "jest-yoshi-preset": "^3.20.0",
+      "lint-staged": "^7.2.2",
+      "prop-types": "~15.5.0",
+      "puppeteer": "^1.10.0",
+      "react": "15.5.4",
```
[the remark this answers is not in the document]
liorMar: @yurynix done

packages/create-yoshi-app/templates/business-manager-module-typescript/test/dev/server.ts
```diff
@@ -1,3 +1,33 @@
-require('../environment')
-  .environment()
-  .then(env => env.start());
+const getPort = require('get-port');
+
+const isServerRunning = async () => {
```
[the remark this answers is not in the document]
liorMar: @yurynix changed to the suggestion

packages/create-yoshi-app/templates/business-manager-module-typescript/test/dev/server.ts
```diff
@@ -1,3 +1,26 @@
+let testKitEnv;
```
yurynix: let's use the same case as in https://github.com/wix/yoshi/pull/850/files#diff-d048400909ba6afd8551925810853883R42

packages/jest-yoshi-preset/types.d.ts
```diff
@@ -3,6 +3,7 @@ import { Browser, Page } from 'puppeteer';
 declare global {
   const browser: Browser;
   const page: Page;
+  const testkitEnv;
```
yurynix: It should be somehow in the template itself, not `jest-yoshi-preset`, since only this template uses it.

packages/create-yoshi-app/templates/business-manager-module-typescript/src/external-global-types.d.ts
```diff
@@ -0,0 +1,15 @@
+declare global {
+  let testKitEnv: {
+    start: Function;
```
yurynix: `start` here is the same as `businessManager.start`
`stop` here is the same as `businessManager.stop`

We don't really need both, or I'm missing something?

We can have something like:
```es6
export const environment = async envConfig =>
    createTestkit(await getTestKitConfig(envConfig));
```


or 

```es6
export const environment = async envConfig => {
  const testKit = createTestkit(await getTestKitConfig(envConfig));
  return {
    businessManager: testkit,
  };
}
```

What do you think?

yurynix:
Thank you for working on that @liorMar 👍

packages/create-yoshi-app/templates/business-manager-module-typescript/test/dev/server.ts
```diff
@@ -1,3 +1,22 @@
 require('../environment')
   .environment()
-  .then(env => env.start());
+  .then(env => {
+    testKitEnv = env;
```
Schniz: How does that pass the tests? it looks like you haven't defined `testKitEnv` 🤔

sidoruk-sv: @Schniz it is defined globally here
https://github.com/wix/yoshi/pull/850/files#diff-d048400909ba6afd8551925810853883R42

yurynix: It has typedef here:
https://github.com/wix/yoshi/pull/850/files#diff-4c4b03fd3e6501f2b105f7a66b5f2d1fR2

❤️ the suggestion!

yurynix: We already have magic test related [globals](https://github.com/wix/yoshi/blob/master/packages/jest-yoshi-preset/types.d.ts#L3), such as `page` and `browser`, adding a global testkit looks to me in-convention with the rest of jest setup.

Schniz: I think we should consider [the boy scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html), and constantly rethink the way we code and make it better than before. If we think this is a better way (and that's _a big if_), I believe we should write new code this way.

There's also the possibility that I'm completely wrong, and this is a good way of doing stuff in tests. For example, `it` and `expect` are globals, so why not have `testKit` which is another test "primitive"?

This is may also be out of context, and can be treated in another PR, but I believe this is a good way of avoiding "legacy apps". Constantly touching older parts of the system 😸

Schniz:
btw I think I would consider it as a new feature? @ronami

packages/create-yoshi-app/templates/business-manager-module-typescript/src/external-global-types.d.ts
liorMar: done

packages/create-yoshi-app/templates/business-manager-module-typescript/test/dev/server.ts
liorMar: done

packages/create-yoshi-app/templates/business-manager-module-typescript/test/environment.ts
```diff
@@ -4,7 +4,15 @@ import {
   ModuleConfigFileEmitter,
 } from '@wix/business-manager/dist/testkit';
 
-const getTestkitConfig = async () => {
+const getTestKitConfig = async ({
+                                  dispatcherPort = 5000,
```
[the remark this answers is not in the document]
liorMar: done prettier

packages/create-yoshi-app/scripts/runE2E.js
```diff
@@ -23,7 +23,7 @@ if (focusProjects) {
 
 // Instantiate a Mocha instance.
 const mocha = new Mocha({
```
yairhaimo: Do you still use mocha in this template?

yurynix: Nope, the tests of the templates are in mocha, and they weren't passing on the CI, so we increased the timeout.

yurynix:
Thank you for the review @yairhaimo !

PR #858, opened 2018-11-26
Test Yoshi opposite several node versions (travis config)

sidoruk-sv:
`8` - past LTS
`10` - current LTS
`node` - refer to latest Node.js version


sidoruk-sv:
Can I somehow help this PR?)

ronami:
Hey @sidoruk-sv, thanks for opening the PR and sorry for the long time it took me to respond.

I like the idea of running our tests suites on different node versions (and possibly on different operating systems, see what Parcel is doing https://dev.azure.com/devongovett/devongovett/_build/results?buildId=422).

With that said, I think it's too early and can make our work (currently) harder. We still have flaky tests, and since Yoshi is used internally in Wix, we can count that pretty much everyone uses the same (LTS) node version. Also, it can be confusing for users distinguishing between 15 or more status checks.

I don't see a big benefit of testing Yoshi on node versions other than LTS (at least not currently). Being able to test Yoshi on MacOS would be great as most of our users run it locally.

PR #867, opened 2018-12-02
change redex to exclude stylable css files

jonathanadler:
### 🔦 Summary
My previous PR changed the regex to include css files, however this collided with stylable css file configuration for tests

### 🗺️ Test Plan
Is there a way we can add tests to this?


yairhaimo:
Duplicate with #866.
Thanks!

PR #871, opened 2018-12-04
Add a test that verifies directory listing in assets directory

yurynix:
### 🔦 Summary
Adds a test from https://github.com/wix/yoshi/pull/790 that verifies that dev server serving directory index for `/assets` is index as well.

### 🗺️ Test Plan
Assert tests are passing


test/kitchensink/integration/dev-server.test.js
```diff
@@ -10,4 +10,14 @@ describe('dev-server', () => {
       expect.arrayContaining(['components/', 'app.bundle.js', 'assets/']),
     );
   });
+
+  it('shows the contents of assets dir', async () => {
+    await page.goto('http://localhost:3200/assets');
+
+    const list = await page.$$eval('#files li a', a => {
```
Schniz: `#files li a` is very specific to the implementation (which isn't ours anyway): what about testing that the response just contains the string `hello.txt`?
or at least checking all the anchor tags and not scope it to `#files li` like `$$eval('a', a => a.map(li => li.textContent))`?

test/kitchensink/integration/dev-server.test.js
```diff
@@ -10,4 +10,14 @@ describe('dev-server', () => {
...
+      return a.map(li => li.textContent);
+    });
+
+    expect(list).toEqual(expect.arrayContaining(['hello.txt']));
```
Schniz: this can be written also as:

```suggestion
    expect(list).toContain('hello.txt');
```

which is less code and more reasonable IMO. what do you think?

Schniz:
```node
> let expect = require('expect')
undefined
> expect([1,2,3]).toContain(1);
undefined
> expect([1,2,3]).toContain(10);
Error: expect(array).toContain(value)

Expected array:
  [1, 2, 3]
To contain value:
  10
```

test/kitchensink/integration/dev-server.test.js
yurynix: Sure!

Schniz:
one thing to consider: tests are fixed on `3200`, is it good?

yurynix:
Thank you for the review @Schniz !

> one thing to consider: tests are fixed on 3200, is it good?

It's not just that one test, we should handle that systematically for all tests together IMO.

Issue #774, opened 2018-11-12
Use add-sentry and add-fedops when generating a new project


technotronic12:
# 🙋 Feature Request
Add sentry and fedops to newly generated projects

### 😯 Current Behavior
It is not adding those two tools

### 🤔 Expected Behavior
New projects should come with this tool by default

### 💁 Possible Solution 
Use the two libraries:
add-fedops: https://github.com/wix-private/fed-infra/tree/master/add-fedops
add-sentry: https://github.com/wix-private/fed-infra/tree/master/add-sentry

### 🔦 Context 
**Add fedops** 
should work without passing any params, it uses the data from package.json

**Add Sentry**
It receives two params, teamName and ProjectName. We can either read them from package.json  (we can use name as projectName and the organization as teamName) or we can ask people if they would like to have a new sentry project created and ask if the name and the team name should be taken from package.json.

technotronic12:
Yo, only saw this message only now.
When a user creates a new client / fullstack project he should have sentry and fedops already pre-configured in the project. Similarly to what we have with new relic both in the scala and node framework.

When speaking about it with @yairhaimo we decided that the best way to do it is to provide a module that does exactly that, adds all of the needed config to a project and this way it can be used both in new and existing projects.

The flow should be that when the user is creating a new project it should come with fedops in it. Sentry, when he is getting asked for the different fields in the generator it should also ask him if he wants sentry to be added to his project. Sentry needs the team name and project name to be added to Sentry website (add-sentry module does it), we should ask him to provide a team name (we can probably use the organization name that you ask for in the generator questions). Project name should be the project name from package.json i guess.

Schniz:
I'm opening a discussion here (after speaking with @technotronic12, @cooperd, @yairhaimo and @ranyitz):

Using `add-fedops` and `add-sentry` inside `create-yoshi-app` is problematic in terms of maintenance, testing, reliability and speed.

In addition to add documentation (which is done in #873), we totally understand how important is to make it easier to integrate error tracking/monitoring. Let's talk about each feature individualy:

## `add-fedops`

This executable just adds a couple of libraries, adds a new file called `fedops.json`, updates the main component definition and the `index.ejs`. basically, modifies static files.

Although it sounds logical to run it on every new template for "code reuse", we won't be testing it in our e2e. And if we would have tested that, it just makes no sense to have it in a different repo.

[...]

PR #872, opened 2018-12-04
Support sentry and fedops for client projects

Schniz:
WIP.

### 🔦 Summary
Trying to integrate `@wix/add-sentry` and `@wix/add-fedops` into the workflow, for fixing #774.

### 🗺️ Test Plan
... wip ...
add a test maybe? make this the default maybe?


packages/create-yoshi-app/src/getQuestions.js
```diff
@@ -39,6 +39,30 @@ module.exports = () => {
...
+        },
+        {
+          title: 'FedOps',
+          value: 'fedops',
+        },
+      ],
+    },
+    {
+      type: (_, { extraCapabilities }) =>
+        extraCapabilities && 'sentry' in extraCapabilities ? 'text' : null,
```
Schniz: this code looks bad

Schniz: also, i'm not sure about the `in` operator. looks better than `[].indexOf('x') !== -1` and `[].find(x => x === 'sentry')` but can be confusing

Schniz: of course it is confusing, because it doesn't work

Schniz:
After a conversation with @ranyitz and @yairhaimo we have decided that it'd be better to just add a guide to the documentation, instead of adding these options to `create-yoshi-app`

PR #873, opened 2018-12-04
Add "Monitoring" and "Error Tracking" guides

### LINKED ISSUE #774: Use add-sentry and add-fedops when generating a new project

Schniz:
### 🔦 Summary
Adding the knowledge of "production ready app" to Yoshi's documentation, instead of adding it as command (as suggested in #872).

Fixes #774


packages/create-yoshi-app/src/createApp.js
```diff
@@ -69,6 +69,6 @@ module.exports = async (workingDir, projectDirName) => {
 
   console.log(chalk.cyan('npm start\n'));
 
-  console.log('For more information visit https://github.com/wix/yoshi');
+  console.log('For more information visit https://wix.github.io/yoshi');
```
Schniz: I believe this is a good time to replace the repo url with Yoshi's website!

Schniz:
@ranyitz I have changed things 😸

ronami:
@Schniz @ranyitz is this still relevant? Can it be merged?

PR #877, opened 2018-12-05
Add FedOps to create-yoshi-app templates

Schniz:
### 🔦 Summary

Integrating `fedops` to Yoshi templates by default, while not hurting dev experience 😸 

### 🗺️ Test Plan

...

### DIFF
--- a/packages/create-yoshi-app/templates/client/src/components/App/App.js
+++ b/packages/create-yoshi-app/templates/client/src/components/App/App.js
@@ -1,3 +1,4 @@
+import * as fedops from 'fedops-logger';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { translate } from 'react-i18next';
@@ -7,6 +8,15 @@ class App extends React.Component {
   static propTypes = {
     t: PropTypes.func,
   };
+
+  // The app load event is fired from the browser before the app is loaded.
+  // After the app is loaded, using the appLoaded function on the componentDidMount
+  // function to notify Grafana that the app was fully loaded.
+  componentDidMount() {
+    const logger = fedops.logger || fedops;
+    const fedopsLogger = logger.create('{%projectName%}');
+    fedopsLogger.appLoaded();
+  }
 
   render() {
     const { t } = this.props;
--- a/packages/create-yoshi-app/templates/client/{%packagejson%}
+++ b/packages/create-yoshi-app/templates/client/{%packagejson%}
@@ -16,6 +16,7 @@
     "release": "yoshi release"
   },
   "devDependencies": {
+    "fedops-grafana-api": "^2.0.6",
     "enzyme": "~2.9.0",
     "express": "~4.15.0",
     "express-session": "~1.15.0",
@@ -29,6 +30,7 @@
     "yoshi-style-dependencies": "^3.0.0"
   },
   "dependencies": {
+    "fedops-logger": "^3.0.28",
     "axios": "~0.16.0",
     "i18next": "^11.6.0",
     "prop-types": "~15.6.0",
@@ -42,6 +44,7 @@
   },
   "yoshi": {
     "externals": {
+      "fedops-logger": "^3.0.28",
       "react": "React",
       "react-dom": "ReactDOM"
     },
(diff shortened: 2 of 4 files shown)


packages/create-yoshi-app/templates/client/{%packagejson%}
```diff
@@ -56,6 +59,9 @@
     ]
   },
   "eslintConfig": {
+    "globals": {
+      "fedops": true
```
Schniz: ugly stuff (for defaults)

packages/create-yoshi-app/templates/client/src/components/App/App.js
```diff
@@ -8,6 +9,15 @@ class App extends React.Component {
     t: PropTypes.func,
   };
 
+  // The app load event is fired from the browser before the app is loaded.
```
Schniz: I'm not sure I like the bloat it adds to the `App` component. Feels like it needs to be a HoC or something

packages/create-yoshi-app/templates/client/{%packagejson%}
Schniz: I mean, it's a new project. no need to add it 🤷‍♂️

packages/create-yoshi-app/templates/client/src/components/App/App.js
Schniz: Integrating it just by adding these lines feels half baked. We can have providers, we can really think about the use cases (how to log important events?) and make it nice

should we just keep it here?
also, we can minimize it to a one-liner, because we have no reason not to:

```diff
- const logger = fedops.logger || fedops;
- const fedopsLogger = logger.create('{%projectName%}');
- fedopsLogger.appLoaded();
+ fedops.logger.create('{%projectName%}').appLoaded();
```

wdyt? @technotronic12

Schniz: a cool solution can be a `@wix/react-fedops-logger` package that have a very minimal API surface:

```jsx
import * as fedops from '@wix/fedops-logger';

const Logger = fedops.logger || fedops;
export const Context = React.createContext(null);

export default class LoggerProvider extends React.Component {
  constructor(props) {
    super(props);
    this._logger = Logger.create(this.props.name);
  }

  componentDidMount() {
    this._logger.appLoaded();
  }

  render() {
    return <Context.Provider>{this.props.children}</Context.Provider>;
  }
}
```

this would give your team the ability to change the behavior, like we talked in person:

```diff
  componentDidMount() {
-   this._logger.appLoaded();
+   setTimeout(() => this._logger.appLoaded(), 0);
  }
```

technotronic12: btw, this line `const logger = fedops.logger || fedops;` is used because if you use fedops from our cdn (unpkg) you will get it differently than if you use it as an import. 

The LoggerProvider looks cool!

Schniz: well - that's fine, and can be isolated in your provider instead of spreading that logic to every app 😸

Schniz: well - that's fine, and can be isolated in your provider instead of spreading that logic to every app 😸

packages/create-yoshi-app/templates/client/{%packagejson%}
Schniz: @technotronic12 do we need it?

Schniz: do we need it? @technotronic12

technotronic12: yes when we load it from unpkg (you don't have an import in code)

Schniz: this is only in `index.vm`. I'm not sure this code is being checked with eslint

Schniz: also, I'd rather not add it globally anyway.

```html
<script>
/* global fedops */
fedops // should work
/* global fedops:false */
</script>
```

in `index.vm` you use `window.fedops` which should work without configurations

PR #890, opened 2018-12-23
PR CI have BUID_VCS_NUMBER, and don't have ARTIFACT_VERSION

yurynix:
### 🔦 Summary
In PR CI we have `BUID_VCS_NUMBER`, and don't have `ARTIFACT_VERSION`, we should be able to build for this scenario ("mockRC") as well.

### 🗺️ Test Plan
Unit test attached


yurynix:
`BUID_VCS_NUMBER` is a hash, like that: `c38acd438041e8fb371b9f6b9121f07e71ddc445`

screen shot 2018-12-23 at 17 02 38

ranyitz:
@yurynix ok, but how the url looks on parastorage? It has this hash in the url?

packages/yoshi/config/webpack.config.js
```diff
@@ -60,13 +60,8 @@ const computedSeparateCss =
...
 
-// default public path
-let publicPath = '/';
-
-if (isDevelopment) {
-  // Set the local dev-server url as a path
-  publicPath = project.servers.cdn.url;
-}
+// default public path set the local dev-server url as a path
+let publicPath = project.servers.cdn.url;
```
ranyitz: Please note that there are sites cases when our bundles get to production (without `pom.xml`) which means the public path will be something with `localhost:3200` (for library for example), so I rather leave the default as `/`.

yurynix: Could you elaborate more on that use case:
> when our bundles get to production (without pom.xml) 

The use case I'm trying to solve here is a project with dynamic imports that is running from a business manager testkit (`http://localhost:5000`).

If we'll use `/` as the default and this project does e2e tests, it'll make a build that is not `isDevelopment`,  it'll try to look for chunks on `/` => `http://localhost:5000` while they're actually on yoshi's CDN `http://localhost:3200/` so the dynamic `import()` will fail.
(example: https://github.com/wix-private/business-manager-mini-home/pull/105)
screen shot 2018-12-23 at 19 13 07

netanelgilad: I also think the `/` case is invalid. If something is deployed to the cdn, it should the public path of the cdn. And since all our bundles/assets get to the cdn, having their cdn url hardcoded makes sense.

ranyitz: @netanelgilad @yurynix 
Just to make sure we are all on the same page.

When we know the public path we **do** configure it with the project path on the cdn.

We are talking here about the other cases (when we don't know what is the public path):

* When building a browser library.
* When publishing an artifact to `unpkg`.
* Running a project through the business manager testkit.

__________________________________________

The default case in webpack (when not configuring anything) is:

```js
/******/ 	// __webpack_public_path__
/******/ 	__webpack_require__.p = "";
/******/
```

So it makes sense to leave it like that when we don't know what should be the public path for a certain situation. (So the `'/'` case is invalid, but we should change it to `''`)

__________________________________________

We should try to evaluate your case with the business manager. I don't think that it's the general case, it means that people will have `localhost:3200` in production bundles, just because we may want to take advantage of it in some other unrelated case.

The possible options in my opinion are:

1. Making public path configurable and let specific cases configure it however they want. (It's still possible dynamically through [__webpack_public_path__](https://webpack.js.org/guides/public-path/#on-the-fly))
2. See what is the case for a non-development bundle served under `localhost:3200` and use those indicators to specifically change the `publicPath` to `localhost:3200` while not doing so for other non-development bundles created for production use.

ranyitz:
```js
const isDevelopment = process.env.NODE_ENV === 'development';
const isTest = process.env.NODE_ENV === 'development';
const inTeamCity = checkInTeamCity();
...
if (isDevelopment || (isTest && !inTeamCity) ) {
  // Set the local dev-server url as a path
  publicPath = project.servers.cdn.url;
}
```

I'm probably missing something here, what will that do?

yurynix:
We've discussed that with @ranyitz and @netanelgilad and decided that this PR will only handle the PR CI case, and nothing else.

We will try to see if using `yoshi start --no-test` together with `yoshi test --protractor` will solve the issue for local development.

PR #893, opened 2018-12-25
Add a note about lerna and tags

yurynix:
### 🔦 Summary
Add a note about tags and [lerna](https://github.com/lerna/lerna/tree/master/commands/version)

### 🗺️ Test Plan
N/A


yurynix:
Maybe also mention here or in the fed handbook:
```bash
git config --global remote.origin.tagOpt --tags
```

PR #896, opened 2018-12-26
Change Typescript transpliation target for ES modules

yairhaimo:
### 🔦 Summary
Currently we transpile es modules libraries to `es2015` in order to support tree shaking.
In order to support dynamic imports this PR changes the target to `esNext`.


yairhaimo:
This is a potential breaking change but with a low risk.
I dont think we should block this feature until the next major version.
The relevant parties were notified of the change and we'll revert if we receive any complaints.

yairhaimo:
Added `experimentalTSTarget` config flag in order to make this an opt in feature and not a breaking change.

PR #903, opened 2019-01-06
`update-node-version` - compare version correctly

yairhaimo:
### 🔦 Summary
Use node semver instead of regular comparison in order to compare version correctly.


yairhaimo:
This PR fixes a bug in the task. 
We removed this task altogether in Yoshi 3. Do we want to remove it also in Yoshi 2?

ranyitz:
@yairhaimo I think that it makes sense to also remove it here.

yairhaimo:
Close in favor of https://github.com/wix/yoshi/pull/907

PR #915, opened 2019-01-16
adds artifact id env var to bundle

aaronvine:
### 🔦 Summary
extracting artifact id from pom.xml and adding it to the bundle in process env var

### 🗺️ Test Plan
testing that bundle holds artifact id in the defined process env var

### DIFF
--- a/packages/yoshi-helpers/utils.js
+++ b/packages/yoshi-helpers/utils.js
@@ -220,3 +220,14 @@ module.exports.killSpawnProcessAndHisChildren = child => {
     });
   });
 };
+
+/**
+ * Gets the artifact id of the project at the current working dir
+ */
+module.exports.getProjectArtifactId = () => {
+  const artifactId = new xmldoc.XmlDocument(
+    fs.readFileSync(POM_FILE),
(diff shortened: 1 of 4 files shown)


packages/yoshi-helpers/utils.js
```diff
@@ -185,13 +185,28 @@ function concatCustomizer(objValue, srcValue) {
 
 module.exports.mergeByConcat = require('lodash/fp').mergeWith(concatCustomizer);
 
+/**
+ * Gets the artifact id of the project at the current working dir
+ */
+const getProjectArtifactId = () => {
+  try {
```
ronami: Why should we wrap it with a `try {} catch {}`?

ronami: I can see that we’re checking for the existence of a pom.xml file in other places. I think we should do the same here: instead of a try catch case (which willl also catch other errors than a missing file), we can check if the file exists and return an empty string if it doesn’t.

Wdyt?

aaronvine: yeah sounds better 👍 I'll update it

PR #916, opened 2019-01-16
yoshi-helpers: don't swallow module not found if its not direct module

technotronic12:
### 🔦 Summary
Mocha setup wasn't working and no error message was thrown. Apparently, a 'module-not-found' error was thrown of by one of the modules that mocha-setup.ts file was trying to import.

In tryRequire of Yoshi helpers we wanted not to throw errors if the specifically requested module was not found if one of its dependencies is missing we do want to see the error.


ronami:
My bad @technotronic12, done already in #969. Closing.

PR #930, opened 2019-01-21
Add process.env.IS_MINIFIED.

felixmosh:
### 🔦 Summary
This adds the ability to check if the bundle is minified or not and by that to consume the proper assets.

For example, in `bundle.min` to point to `styles.min.css`.

### 🗺️ Test Plan
The minified bundle should have `process.env.IS_MINIFIED === true` otherwise `process.env.IS_MINIFIED === false`

### DIFF
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -590,6 +590,7 @@ function createClientWebpackConfig({
         'process.env.NODE_ENV': JSON.stringify(
           isProduction ? 'production' : 'development',
         ),
+        'process.env.IS_MINIFIED': isDebug ? 'false' : 'true',
         'window.__CI_APP_VERSION__': JSON.stringify(
           artifactVersion ? artifactVersion : '0.0.0',
         ),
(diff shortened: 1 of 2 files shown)


packages/yoshi/config/webpack.config.js
```diff
@@ -590,6 +590,7 @@ function createClientWebpackConfig({
         'process.env.NODE_ENV': JSON.stringify(
           isProduction ? 'production' : 'development',
         ),
+        'process.env.IS_MINIFIED': isDebug ? 'false' : 'true',
```
ronami: It's really small but I think it would be nice to use: `JSON.stringify(isDebug)` instead, so it's consistent with the other two.

felixmosh: This is not the case, they use stringify in order to wrap the value as a string, but with my implementation it replaces it to a `boolean`

ronami:
I see why some tests are failing, I'm pushing a fix to `master`.

packages/yoshi/config/webpack.config.js
felixmosh: Yeah, It will work

Issue #822, opened 2018-11-19
Replace `url-loader` with `svg-url-loader` for SVGs


ronami:
### Summary

Should make some performance improvements, see https://github.com/bhovhannes/svg-url-loader on why and how.

This is a potentially breaking change, I think we should add it to v4 if we can before it's released.

ronami:
Thanks to @saarkuriel this is shipping in v4 🎉

PR #937, opened 2019-01-27
svg-url-loader (#822)

saarkuriel:
### 🔦 Summary
Use `svg-url-loader` for requiring svg files.
Used `iesafe: true` as we still need to support IE...

### DIFF
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -442,6 +442,22 @@ function createCommonWebpackConfig({
                   loader: 'svg-url-loader',
                   options: {
                     iesafe: true,
+                    noquotes: true,
+                    limit: 10000,
+                  },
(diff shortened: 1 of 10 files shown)


packages/yoshi/config/webpack.config.js
```diff
@@ -439,9 +439,25 @@ function createCommonWebpackConfig({
...
+                    noquotes: true,
+                    limit: 10000,
+                  },
+                },
+              ],
+            },
+            {
+              test: /\.svg$/,
+              issuer: {
+                test: reStyle,
```
ronami: I think we can remove this so every `svg` file that wasn't imported from a `css/scss` file will be catched by it. Wdyt?

saarkuriel: agreed

Issue #830, opened 2018-11-20
Schema for webpack config 'externals' field is outdated


AdirAmsalem:
# 🐛 Bug Report

### 😯 Current Behavior

when i use the following `externals` configuration:
```
"externals": {
  ...
  "firebase/app": ["firebase", "app"],
  ...
},
```

i get the following warning from `yoshi`:
```
Warning: Invalid configuration object. Yoshi has been initialised using a configuration object that does not match the API schema.
 - configuration/externals should be a string.
   Field Description: An exact matched dependency becomes external. The same string is used as external dependency.
 - configuration/externals/firebase~1app should be a string.
 - configuration/externals/firebase~1app should be an object.
 - configuration/externals/firebase~1app should be a boolean.
 - configuration/externals/firebase~1app should be one of these:
   string | object | boolean
```

it happens because of the schema defined here: https://github.com/wix/yoshi/blob/master/packages/yoshi-config/schema/yoshi-config-schema.json#L14-L58

### 🤔 Expected Behavior

it should not output any warning for arrays as it's valid and documented here: https://webpack.js.org/configuration/externals/#array
moreover, it’s practically working :)

### 💁 Possible Solution 

probably fixed here: https://github.com/webpack/webpack/pull/8043
so perhaps the schema we use should be updated.

### 👣 Steps to Reproduce 

use the aforementioned externals configuration and it'll show up.

thanks!

ronami:
Thanks for the issue @AdirAmsalem.

I talked with @netanelgilad about it and I think it's best to remove Webpack's schema from our code, and instead `require()` it from `webpack/schemas`. We can construct our own schema dynamically, so we won't have to chase after Webpack with fixes or changes to their schema.

PRs welcome :)

ronami:
Fixed in #941 thanks to @matveyok 🎉

PR #941, opened 2019-01-28
Update yoshi-config-schema to use webpack schema defaults

matveyok:
### 🔦 Summary
Update `yoshi-config-schema` to use webpack schema when same as default
Related Issue - #830


packages/yoshi-config/package.json
```diff
@@ -13,6 +13,7 @@
     "ajv-keywords": "^3.2.0",
     "chalk": "^2.4.1",
     "cosmiconfig": "^4.0.0",
-    "lodash": "^4.17.10"
+    "lodash": "^4.17.10",
+    "webpack": "4.28.4"
```
ronami: Should be a dev dependency, I probably missed it. I'll fix it tomorrow.

PR #947, opened 2019-01-29
Remove redundant include for graphql files

ronami:
### 🔦 Summary

Really small: Since #582 I don't think we need to whitelist which 3rd parties are being processed with `graphql` files, just process all of them (like other extensions).


packages/yoshi/config/webpack.config.js
```diff
@@ -454,7 +454,6 @@ function createCommonWebpackConfig({
         // Rules for GraphQL
         {
           test: /\.(graphql|gql)$/,
-          include: project.unprocessedModules,
```
ranyitz: horeyy!

PR #949, opened 2019-01-30
build-app: fix check if public dir exists

yaelhe:
### 🔦 Summary
...
build-app awaits on `fs.exists()`. This is an invalid call. `fs-extra` doesn't implement this function so it falls back to node's implementation. In node 8 it resolves with `undefined` (because this function doesn't return a promise) and in node 10 it fails with a `TypeError` as `Callback must be a function`, since no callback was passed.
`fs-extra` has `fs.pathExists()` which returns a promise as expected.


yaelhe:
Closing, had the wrong base branch but after changing it the diff was kept. Opened new PR https://github.com/wix/yoshi/pull/951

PR #951, opened 2019-01-30
build-app: fix check if public dir exists

yaelhe:
build-app awaits on `fs.exists()`. This is an invalid call. `fs-extra` doesn't implement this function so it falls back to node's implementation. In node 8 it resolves with `undefined` (because this function doesn't return a promise) and in node 10 it fails with a `TypeError` as `Callback must be a function`, since no callback was passed.
`fs-extra` has `fs.pathExists()` which returns a promise as expected.


yurynix:
Could you elaborate a bit more?

```
> const fs = require('fs-extra');
undefined
> fs.exists
[Function]
> const nodeFs = require('fs');
undefined
false
> fs.exists.toString()
'function (filename, callback) {\n  if (typeof callback === \'function\') {\n    return fs.exists(filename, callback)\n  }\n  return new Promise(resolve => {\n    return fs.exists(filename, resolve)\n  })\n}'
> 
```
```
node -v
v8.11.3
```
```
yoshi-monorepo@ /Users/yurym/wix/yoshi
├── fs-extra@5.0.0 
└─┬ lerna@2.11.0
  └── fs-extra@4.0.3 
```

https://github.com/jprichardson/node-fs-extra/blob/master/lib/fs/index.js#L65

yurynix:
I don't have the problem with the usage of `pathExists` instead of `exists`, however it solves no problem:

screen shot 2019-02-03 at 12 12 42

yurynix:
Cherry-picked https://github.com/wix/yoshi/commit/ee9e807d9e254d8a87f7736f00457bb12f9d8f08 for `test:app-flow` to pass

/cc @ronami

ronami:
🏆

yaelhe:
Thanks!

PR #917, opened 2019-01-16
Configure babel to ignore babelrc and use babel-preset-yoshi

ronami:
### 🔦 Summary

This PR moves Babel's configuration from the developer to Yoshi: Yoshi will ignore `.babelrc` and `babel.config.js` and instead will pre-configure Babel with `babel-preset-yoshi`.

This change will allow us to configure Babel in the best way possible: Like transpiling server code to the current Node version.

PR #954, opened 2019-01-30
Migrate to Jest 24

matveyok:
### 🔦 Summary
Upgrade to Jest 24

### 🗺️ Test Plan
Tests should pass 🤷‍♂️


matveyok:
@ronami FYI
I did a soft reset on the old branch and applied the changes on a new branch.

packages/yoshi/test/test.spec.js
```diff
@@ -606,7 +609,7 @@ describe('Aggregator: Test', () => {
             `,
         });
 
-        const buildResponse = project.execute('build', [], {
+        const buildResponse = project.verbose().execute('build', [], {
```
ronami: We probably needed it for debugging but we can remove it if the PR is ready to be merged.

packages/yoshi/test/test.spec.js
```diff
@@ -616,7 +619,7 @@ describe('Aggregator: Test', () => {
           staticsDomain,
         );
 
-        const testResponse = project.execute('test', ['--jest'], {
+        const testResponse = project.verbose().execute('test', ['--jest'], {
```
ronami: Same as . above.

packages/yoshi/test/test.spec.js
```diff
@@ -568,6 +568,9 @@ describe('Aggregator: Test', () => {
               jest: {
                 preset: 'jest-yoshi-preset',
               },
+              babel: {
```
ronami: Babel config is not needed here since #917. It was probably got in by a rebase.

packages/yoshi/test/test.spec.js
matveyok: build still fails :(

matveyok:
Related issue fixed in 24.1.0 - [7752](https://github.com/facebook/jest/pull/7752)

matveyok:
@ronami FYI build pass :)

docs/guides/jest-setup.md
```diff
@@ -141,6 +141,6 @@ If you want to run some code before your tests you can use one of the 3 followin
 - `<rootDir>/test/setup.server.(j|t)s`: Bootstrap (server)
 - `<rootDir>/test/setup.e2e.(j|t)s`: Puppeteer (e2e)
 
-These setup files are actually [Jests's `setupTestFrameworkScriptFile`](https://jestjs.io/docs/en/configuration#setuptestframeworkscriptfile-string)
+These setup files are actually [Jests's `setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration#setupFilesAfterEnv-string)
```
ranyitz:
```suggestion
These setup files are actually [Jests's `setupFilesAfterEnv`](https://jestjs.io/docs/en/configuration#setupfilesafterenv-array)
```

Fixed link

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -54,10 +54,17 @@ module.exports = {
...
+        // now it supports more than 1 test file, in future we can expose it here
+        const setupFilesAfterEnv = setupTestsFile ? [setupTestsFile] : []
+
         return {
           ...project,
 
-          transformIgnorePatterns: ['/node_modules/(?!(.*?\\.st\\.css$))'],
+          transformIgnorePatterns: [
+            '/node_modules/(?!(.*?\\.st\\.css$))',
+            require.resolve('babel-preset-yoshi'),
```
ranyitz: @matveyok can you elaborate on why it is needed? (and why it was not needed in jest@23)

packages/yoshi-helpers/utils.js
```diff
@@ -35,6 +35,7 @@ function logIfAny(log) {
 module.exports.createBabelConfig = (presetOptions = {}) => {
   return {
     presets: [[require.resolve('babel-preset-yoshi'), presetOptions]],
+    ignore: [require.resolve('babel-preset-yoshi')],
```
ranyitz: Same question here, why ignoring it at all?

And why both in jest's `transformIgnorePatterns` and here?

packages/jest-yoshi-preset/jest-preset.js
matveyok: short answer? this is the only config that worked :)
long answer? if I'm not adding `transformIgnorePatterns` I get this:
`Reentrant preset detected trying to load "/Users/matveyo/Documents/repositories/yoshi/packages/babel-preset-yoshi/index.js". This module is not ignored and is trying to load itself while compiling itself, leading to a dependency cycle. We recommend adding it to your "ignore" list in your babelrc, or to a .babelignore.`
and ignoring it in `createBabelConfig` fix those builds:
http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=1586823&buildTypeId=Yoshi_TestTemplates
http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=1586831&buildTypeId=Yoshi_TestAppFlow

matveyok: @ranyitz  👆

ranyitz: @matveyok So how can I recreate the issue locally? In which cases does it happen?

I'm not sure where do we require our babel preset and thus transpiling it, and I would like to verify that we're doing the required fix.

yanivefraim: @ranyitz - the tests that are failing are `test:integration`:
`Aggregator: Test: --jest: puppeteer environment`

My guess is that it happens because of this breaking change:
https://github.com/facebook/jest/pull/7562

matveyok:
@yanivefraim interesting, can you explain why your last commit fixes the problem?

yanivefraim:
> @yanivefraim interesting, can you explain why your last commit fixes the problem?

Looks like there's a bug in jest, where node_modules which are symlinked are not ignored.

yanivefraim:
I think we might remove babel [register hack](https://github.com/wix/yoshi/blob/version_4.x/packages/jest-environment-yoshi-puppeteer/globalSetup.js#L25) now. @ronami - wdyt?

PR #957, opened 2019-01-31
Add documentation for the why & how of using a lockfile

cowchimp:
### 🔦 Summary
We recently configured all template projects to not use a lockfile (#953) but this decision might not be suitable for everyone, so we want to be transparent about what this means, why you might choose differently, and how to start using a lockfile.
This new docs section tries to address these issues.

### 🗺️ Test Plan
`¯\_(ツ)_/¯`


docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+
+1. **Using a lock file** - if a `package-lock.json` is present, the exact dependency versions defined in that lock file will be downloaded when you `npm install`.  
+   The `package-lock.json` will be committed to source-control and will affect all developers working on the project and the CI builds.
+2. **Not using a lock file** - if `package-lock.json` is not present, whenever you do a fresh `npm install` the latest versions of your dependencies will be downloaded from the npm registry.
+
+This is a high-level explanation. For more details you can read up on the [NPM documentation](https://docs.npmjs.com/files/package-locks)
+
+## What are the advantages\disadvantages of each approach?
+
+Using a lockfile avoid bugs because of changes made in 3rd-party dependencies and helps ensure reproducible builds.
```
ronami: I think this sentence is missing a few parts, how about: `Using a lockfile helps avoid bugs: By locking the entire dependency tree changes in 3rd party library are introduced only when you initiate an upgrade`?

Should we explain how to upgrade a dependency if they work with a lockfile? Or at least link to npm.

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+1. **Using a lock file** - if a `package-lock.json` is present, the exact dependency versions defined in that lock file will be downloaded when you `npm install`.  
+   The `package-lock.json` will be committed to source-control and will affect all developers working on the project and the CI builds.
+2. **Not using a lock file** - if `package-lock.json` is not present, whenever you do a fresh `npm install` the latest versions of your dependencies will be downloaded from the npm registry.
+
+This is a high-level explanation. For more details you can read up on the [NPM documentation](https://docs.npmjs.com/files/package-locks)
+
+## What are the advantages\disadvantages of each approach?
+
+Using a lockfile avoid bugs because of changes made in 3rd-party dependencies and helps ensure reproducible builds.  
+Not having a lockfile reduces the need to remember to update the lockfile and reason about the changes, and dealing with lockfile conflicts.
```
ronami: I think that not having a lockfile is the default that most developers/team know. I think this sentence should tell what are the problems with a lockfile. Something like: `By using a lockfile, internal libraries that you want to always be up to date will have to be manually updated. Forgetting to update those libraries can cause problems that may be hard to detect`?

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+
+Generally there are two approaches to working with dependencies in the NPM ecosystem:
+
+1. **Using a lock file** - if a `package-lock.json` is present, the exact dependency versions defined in that lock file will be downloaded when you `npm install`.  
+   The `package-lock.json` will be committed to source-control and will affect all developers working on the project and the CI builds.
+2. **Not using a lock file** - if `package-lock.json` is not present, whenever you do a fresh `npm install` the latest versions of your dependencies will be downloaded from the npm registry.
+
+This is a high-level explanation. For more details you can read up on the [NPM documentation](https://docs.npmjs.com/files/package-locks)
+
+## What are the advantages\disadvantages of each approach?
```
ronami: Should we add that having a lock file results in significantly faster installations?

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+
+Using a lockfile avoid bugs because of changes made in 3rd-party dependencies and helps ensure reproducible builds.  
+Not having a lockfile reduces the need to remember to update the lockfile and reason about the changes, and dealing with lockfile conflicts.
+
+## What is the default\preferred approach?
```
ronami: I think that this describes the `how` to opt-into using or not using a lock file instead of what the preferred approach is, maybe change the title?

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+
+Overall the default NPM behavior is to use a lock file, and one is generated automatically when you run `npm install` for the first time.
```
ronami: I think it's irrelevant what the default npm behavior is for this title. I think that it can go in the `how` part. I think this part should explain what the recommended approach and why we think it is recommended. Wdyt?

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+
+## What is the default\preferred approach?
+
+Overall the default NPM behavior is to use a lock file, and one is generated automatically when you run `npm install` for the first time.  
+However for projects generated by `create-yoshi-app` a `package-lock.json` will not be generated.  
+This is because in Wix it's common to depend on other Wix packages that update often (on every commit), therefore you would need to update your `package-lock.json` very often and conflicts are likely.
+
+## How to start using a lockfile?
+
+An `.npmrc` (NPM configuration file) in the root folder instructs NPM whether to generate a lockfile or not using the `package-lock=false` configuration.
```
ronami: I would re-phrase it as an instruction to the user. Something like: `By default, newly generated projects do not use a lock file. To use a lock file, edit your .npmrc file with package-lock=true`?

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
...
+id: managing-dependencies
+title: Managing Dependencies
+sidebar_label: Managing Dependencies
+---
+
+Generally there are two approaches to working with dependencies in the NPM ecosystem:
+
+1. **Using a lock file** - if a `package-lock.json` is present, the exact dependency versions defined in that lock file will be downloaded when you `npm install`.  
+   The `package-lock.json` will be committed to source-control and will affect all developers working on the project and the CI builds.
+2. **Not using a lock file** - if `package-lock.json` is not present, whenever you do a fresh `npm install` the latest versions of your dependencies will be downloaded from the npm registry.
```
ranyitz: > the latest versions of your dependencies will be downloaded 

NPM will try to download the highest versions of the dependencies specified in the `package.json`. You can reference to this link - https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

ronami:
Looking good, this is something a lot of developers ask about. 🏆

docs/guides/managing-dependencies.md
```diff
@@ -0,0 +1,29 @@
+---
+id: managing-dependencies
+title: Managing Dependencies
+sidebar_label: Managing Dependencies
+---
+
+Generally there are two approaches to working with dependencies in the NPM ecosystem:
+
+1. **Using a lock file** - if a `package-lock.json` is present, the exact dependency versions defined in that lock file will be downloaded when you `npm install`.
```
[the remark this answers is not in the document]
cowchimp: Thanks. Will cover this in the advantages\disadvantages section.

docs/guides/managing-dependencies.md
cowchimp: I agree with all your other notes but here I think differently. I don't think we should "hide" the fact that we chose differently from the industry-standard, and do stay transparent on what that industry-standard is.
But I do mention why ultimately we chose differently and why we think that's the right approach for us.

PR #958, opened 2019-02-03
Update stylable runtime config

tomrav:
### 🔦 Summary
Upgraded Stylable webpack runtime config.


packages/yoshi/config/webpack.config.js
```diff
@@ -605,6 +605,11 @@ function createClientWebpackConfig({
         filename: '[name].stylable.bundle.css',
         includeCSSInJS: !stylableSeparateCss,
         optimize: { classNameOptimizations: false, shortNamespaces: false },
+        runtimeMode: 'shared',
+        globalRuntimeId: '__stylable_yoshi__',
```
hepiyellow: @tomrav @barak007 Didn't we say that the global runtime id would have the @styleable/runtime's major version embedded in it?

Is this `__stylable_yoshi__` only a prefix, and `styleable-webpack-plugin` appends the major version?

packages/yoshi/config/webpack.config.js
```diff
@@ -605,6 +605,11 @@ function createClientWebpackConfig({
...
+        generate: {
+          runtimeStylesheetId: 'module'
```
hepiyellow: @tomrav @barak007 Didn't we say that the stylesheetId would be according to the Namespace (and not the webpack module)?

A Namespace which includes: project name, version, path of stylesheet and major version of `@stylable/runtime`.

barak007: ya this should be runtimeStylesheetId: 'namespace'

packages/yoshi/config/webpack.config.js
tomrav: We are not adding runtime major versions at this time. We will consider it for future releases.

packages/yoshi/config/webpack.config.js
tomrav: Good catch. 👍

packages/yoshi/config/webpack.config.js
hepiyellow: Why?
This means that BusinessManager could still be broken without a solution unless ALL BM modules (bundles) use latest yoshi 3.x.
@shlomitc @ronami

hepiyellow: ok. It is not an issue since, bundles using yoshi 3 will get a different namespace-hash in the class-names, so they won't collide.

hepiyellow:
Looks like it is working with the contacts project:
screen shot 2019-02-04 at 14 28 42

@avrahamr @malachico

avrahamr:
Way to go!!!! thanks a lot!!!
And we don't need to change anything, right?

hepiyellow:
@avrahamr right.

hepiyellow:
@ronami @yairhaimo @shlomitc 
I also tested with WSR, in CI. And e2e tests pass.

We can merge this.

hepiyellow:
@avrahamr Can you confirm that the Your `contacts-web`'s Color-Picker now has a white little arrow?

avrahamr:
You mean contact-picker, right? :-)
@malachico tried it and it still didn't work, he will try again later and report back.

malachico:
Hi @hepiyellow ! thanks for the fix. 
currently I've cleaned and reinstalled my modules and still get the black arrow, 
should I update anything? 
is it in production? or should I use latest?

hepiyellow:
@avrahamr yes. Contact-Picker, sorry.

@malachico yoshi release it yesterday, if you are running locally,
make sure you have `yoshi 3.30.0`, and `@stylable/webpack-plugin 1.1.0`

packages/yoshi/config/webpack.config.js
```diff
@@ -605,6 +605,11 @@ function createClientWebpackConfig({
         filename: '[name].stylable.bundle.css',
         includeCSSInJS: !stylableSeparateCss,
         optimize: { classNameOptimizations: false, shortNamespaces: false },
+        runtimeMode: 'shared',
+        globalRuntimeId: '__stylable_yoshi__',
+        generate: {
+          runtimeStylesheetId: 'namespace'
```
hepiyellow: @tomrav Hi,
I thought that this change, was supposed to have the `st-id` be a the stylesheet namespace.
AND that this namespace includes a hash which is composed of:
- project name
- project version
- path to stylesheet

Was that not true?

FYI @NitayRabi

tomrav: That's not the issue this time around. What's happening is a multi-build project where different components are being consumed from a third party in different builds.

This leads to different optimizations occurring, resulting in clashes.

PR #964, opened 2019-02-04
add basic unit tests for `yoshi-config`'s `validateConfig`

cowchimp:
### 🗺️ Test Plan
Add tests for `validateConfig` in `yoshi-config`.
It does a basic sanity unit test for our use of `Ajv` and `YoshiOptionsValidationError`.
Purposefully not checking the real `yoshi-config-schema` but that's open for debate.


packages/yoshi-config/__tests__/validate-config.spec.js
```diff
@@ -0,0 +1,23 @@
+const validateConfig = require('../utils/validate-config');
+
+describe('validateConfig', () => {
+  it('should not throw error if config does not match schema', () => {
```
ranyitz:
```suggestion
  it('should not throw error if config does match schema', () => {
```

PR #965, opened 2019-02-04
 fail early if Yoshi is being run with an old Node version

### LINKED ISSUE #956: Fail early with a clear message if Yoshi is being run with an old Node version
### Summary

Similar to #537, sometimes developers waste time trying to debug things we can have a clear error message about.

__Note:__ This check has to run **before** any unsupported JavaScript syntax is used.

cowchimp:
fixes #956

### 🔦 Summary
Makes the Yoshi CLI break early if you're using an outdated node version (I went with `>=8.0.0` but it's configurable of course)

### 🗺️ Test Plan
Added a unit test for the `verifyNodeVersion` function.
An e2e is problematic because it's difficult to mock `process.version` in a process spawned by `sh.exec`.


cowchimp:
@ronami reused the logic for `create-yoshi-app` as we discussed.
can you please take another look?

packages/yoshi/bin/yoshi.js
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env node
 const importLocal = require('import-local');
+const verifyMinimumNodeVersion = require('yoshi-helpers/verifyMinimumNodeVersion');
+
+verifyMinimumNodeVersion(`8.7.0`);
```
yairhaimo: move version to `constants.js`

yairhaimo:
Looks great! ☝️ One comment

PR #967, opened 2019-02-05
create-yoshi-app: add sentry and fedops to client and fullstack apps

yaelhe:
Add Sentry loader CDN to html file. Users should change its DSN to a new one they will create (perhaps with `add-sentry` tool) therefore there is a comment above it. The E2E test should still pass after the DSN change. This test actually makes sure that the link is still correct.
Fedops is calling `.appLoaded()` from `componentDidMount()` which is a naïve implementation. I expect users moving it to the appropriate place if it makes more sense. For example, after the application's store was initialised with some async stuff. The E2E should pass after any kind of change.


yurynix:
> Fedops is calling `.appLoaded()` from `componentDidMount()` which is a naïve implementation. I expect users moving it to the appropriate place if it makes more sense. For example, after the application's store was initialised with some async stuff. The E2E should pass after any kind of change.

Maybe it should be mentioned in a comment there? What you say?

yaelhe:
> Maybe it should be mentioned in a comment there? What you say?

I don't like giving our users a lot of instructions on how their app should behave, I think they can figure almost everything on their own by reading the README (they do that today). I think this specific case is obvious, because the name of the function is `.appLoaded()`, so I assume everyone would understand it should be after the app has fully loaded. I might be wrong and I have no strong feelings towards it.

packages/create-yoshi-app/templates/client/{%packagejson%}
```diff
@@ -13,22 +13,26 @@
     "pretest": "yoshi build",
     "test": "yoshi test --jest",
     "posttest": "yoshi lint",
-    "release": "yoshi release"
+    "release": "yoshi release && fedops-api sync"
```
ronami: Can this be done on CI/Yoshi? I would prefer that the release script stays as simple as `yoshi release`. Wdyt?

packages/create-yoshi-app/templates/client/src/client.js
```diff
@@ -13,7 +14,7 @@ wixAxiosConfig(axios, { baseURL });
 
 ReactDOM.render(
   <I18nextProvider i18n={i18n(locale)}>
-    <App />
+    <App fedopsLogger={fedopsLogger.create('{%projectName%}')} />
```
ronami: Most users will need to access the fedops logger instance from more than their main component. Some may start passing it manually from a component to its children (I don't think that's what we want).

What do you think about setting it up similarly to `i18next` with context/hoc setup?

ronami: Can we add a comment linking to fedops logger's docs? just so it's less confusing for new developers.

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
```diff
@@ -12,7 +13,7 @@ describe('App', () => {
   it('renders a title correctly', () => {
     wrapper = mount(
       <I18nextProvider i18n={i18n}>
-        <App />
+        <App fedopsLogger={fedopsLogger.create('{%projectName%}')} />
```
ronami: Is this going to get a request out? Is there a testkit/way to mock it?

packages/create-yoshi-app/templates/client/src/index.vm
```diff
@@ -5,9 +5,16 @@
...
     <link rel="stylesheet" href="${clientTopology.staticsBaseUrl}app#if(!${debug}).min#{end}.css">
-    <script src="https://${clientTopology.staticsDomain}/polyfill/v2/polyfill#if(!${debug}).min#{end}.js?features=default,es6,es7,es2017&flags=gated&unknown=polyfill&rum=0"></script>
-    <script src="https://${clientTopology.staticsDomain}/unpkg/react@15.6.1/dist/react#if(!${debug}).min#{end}.js"></script>
-    <script src="https://${clientTopology.staticsDomain}/unpkg/react-dom@15.6.1/dist/react-dom#if(!${debug}).min#{end}.js"></script>
+    <script type="text/javascript">
+      (function(x,e,o,s,n){var a=window.fedops||{};a.apps=a.apps||{};a.apps[x]={startLoadTime:e&&e.now&&e.now()};try{a.sessionId=o.getItem("fedops.logger.sessionId")}catch(x){}a.sessionId=a.sessionId||"xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g,function(x){var e=s()*16|0,o=x=="x"?e:e&3|8;return o.toString(16)});window.fedops=a;var d="//frog.wix.com/fed?appName="+x+"&src=72&evid=14&session_id="+a.sessionId+"&_="+s();if(!!n&&!!n.sendBeacon){n.sendBeacon(d)}else{(new Image).src=d}})("{%projectName%}",window.performance,window.localStorage,Math.random,window.navigator);
```
ronami: Is it possible to make to have less of an impact on the project? Also, so we can change it without having to migrate a lot of projects. Something like the following:

```html
<script src="wix-fedops.js"></script>
<script>
  wixFedops('{%projectName%}');
</script>
```

packages/create-yoshi-app/templates/client/{%packagejson%}
yaelhe: I agree. This should be a major change though, as currently everyone is explicitly calling it, and we don't want the build/release step calling it twice. Let's add it to yoshi 4 and mention that in the migration guid? If we do that, the generators should move to yoshi 4 too.

packages/create-yoshi-app/templates/client/test/e2e/app.spec.js
```diff
@@ -1,8 +1,56 @@
...
     await page.waitForSelector('h2');
 
     expect(await page.$eval('h2', e => e.innerText)).toEqual('Hello World!');
   });
+
+  describe('Monitoring', () => {
+    describe('Sentry', () => {
+      const { testkit } = sentryTestkit();
+
+      beforeEach(() => {
```
ronami: What do you think about extracting this setup to our Puppeteer setup?

packages/create-yoshi-app/templates/client/test/e2e/app.spec.js
```diff
@@ -1,8 +1,56 @@
...
+
+      afterEach(() => {
+        testkit.reset();
+        testkit.puppeteer.stopListening(page);
+      });
+
+      it('checks that application reports errors to Sentry', async () => {
+        await page.goto('http://localhost:3100');
+        const error = 'Puppeteer test capture error check';
+        await page.addScriptTag({ content: `throw new Error('${error}');` });
```
ronami: I'm not sure we should generate code that tests it. I think we should have sample projects to test that our setup works correctly. I think users should add tests like this only in very specific situations, I think it may be confusing to generate it for every project.

packages/create-yoshi-app/templates/client/test/e2e/app.spec.js
```diff
@@ -1,8 +1,56 @@
...
+        await eventually(() => {
+          const { value } = testkit.getExceptionAt(0);
+          expect(value).toEqual(error);
+        });
+      });
+    });
+
+    describe('Fedops', () => {
+      beforeEach(() => {
+        fedopsTestkit.puppeteer.startListening(page);
```
ronami: I think we can consider adding it too to our Jest Puppeteer setup.

packages/create-yoshi-app/templates/client/src/client.js
yaelhe: I created this provider `fedops-react` in fedops repo, but I don't like this idea. Most actions start in a react component and end inside the store (be it mobx, redux or anything else) when the action ends. I would not instantiate fedops the way I did here. That's why I say it's a naive implementation. I don't recommend storing fedops in react context.

packages/create-yoshi-app/templates/client/test/e2e/app.spec.js
yaelhe: This test makes sure that sentry is set up properly, even after changing the DSN, with `add-sentry` or manually.

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
yaelhe: There is a testkit. I actually added `useFakeReporter()` function for that. Is there a place to setup tests without affecting puppeteer tests?

packages/create-yoshi-app/templates/client/test/e2e/app.spec.js
yaelhe: I think it would be better to put it inside the project so it won't be hidden, maybe `setup.e2e.js`, WDYT?

packages/create-yoshi-app/templates/client/src/client.js
ranyitz: Can `fedopsLogger.create('{%projectName%}')` be called somewhere else? (maybe in the initialization script, and fedopslogger instance would be on `window`?)

We can, of course, add integration with React if needed, using the instance from the window.

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
ranyitz: Do not create a **real** fedops instance in tests.

packages/create-yoshi-app/templates/client/src/index.vm
```diff
@@ -17,6 +24,6 @@
       window.__LOCALE__ = '${locale}';
       window.__STATICS_BASE_URL__ = '${clientTopology.staticsBaseUrl}';
     </script>
-    <script src="${clientTopology.staticsBaseUrl}app.bundle#if(!${debug}).min#{end}.js"></script>
+    <script crossorigin src="${clientTopology.staticsBaseUrl}app.bundle#if(!${debug}).min#{end}.js"></script>
```
ranyitz: Adding `crossorigin` tag to all scripts may be a good thing, but I think it would be better to separate it into a different PR and explain there why this decision was made (as it's probably not related to the fedops logger)

packages/create-yoshi-app/templates/client/{%packagejson%}
```diff
@@ -13,22 +13,26 @@
...
   },
   "devDependencies": {
     "enzyme": "~2.9.0",
     "express": "~4.15.0",
     "express-session": "~1.15.0",
+    "fedops-grafana-api": "^2.0.6",
```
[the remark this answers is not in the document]
yaelhe: Yes, we agreed to move it to yoshi release

packages/create-yoshi-app/templates/client/src/index.vm
yaelhe: >> Do not create a **real** fedops instance in tests.
We are not. A fake reporter will be used with yoshi jest setup

yaelhe: > Adding `crossorigin` tag to all scripts may be a good thing, but I think it would be better to separate it into a different PR and explain there why this decision was made (as it's probably not related to the fedops logger)

This is important for getting stack traces in sentry, so I do want it to come together. There is not a lot of benefit in using sentry without having stacktraces.

packages/create-yoshi-app/templates/client/src/client.js
```diff
@@ -5,6 +5,7 @@ import { I18nextProvider } from 'react-i18next';
 import { wixAxiosConfig } from '@wix/wix-axios-config';
 import App from './components/App';
 import i18n from './i18n';
+import * as fedopsLogger from '@wix/fedops-logger';
```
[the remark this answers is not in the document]
yaelhe: It depends on the way it was exported, and unfortunately it won't work.

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
```diff
@@ -3,6 +3,7 @@ import { mount } from 'enzyme';
 import { I18nextProvider } from 'react-i18next';
 import App from './App';
 import i18n from '../../../test/helpers/i18n.mock';
+import * as fedopsLogger from '@wix/fedops-logger';
```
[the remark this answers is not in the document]
yaelhe: ditto

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
yaelhe: Just to clarify, we want a real instance. We don't want a real reporter.

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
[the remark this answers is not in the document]
ranyitz: I meant that it's not needed to bring everything, we can only bring the `create` function here:
```js
import { create as createLogger } from '@wix/fedops-logger';
```

In any case, I think that the creation of the logger should be somewhere else, better in the script that initializes fedops.

packages/create-yoshi-app/templates/client/src/client.js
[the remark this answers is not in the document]
ranyitz: I meant that it's not needed to bring everything, we can only bring the `create` function here:
```js
import { create as createLogger } from '@wix/fedops-logger';
```

In any case, I think that the creation of the logger should be somewhere else, better in the script that initializes fedops.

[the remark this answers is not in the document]
yaelhe: It was this way at first, and was changed after @yairhaimo 's review. This is a matter of taste and I'm fine with either way. There is nothing that will get eliminated with tree-shaking in both cases.

packages/create-yoshi-app/templates/client/{%packagejson%}
yairhaimo: This was moved to run in ci

packages/create-yoshi-app/templates/client/src/index.vm
```diff
@@ -5,9 +5,18 @@
     <title>${title}</title>
     <link type="image/png" href="https://www.wix.com/favicon.ico" rel="shortcut icon">
     <link rel="stylesheet" href="${clientTopology.staticsBaseUrl}app#if(!${debug}).min#{end}.css">
-    <script src="https://${clientTopology.staticsDomain}/polyfill/v2/polyfill#if(!${debug}).min#{end}.js?features=default,es6,es7,es2017&flags=gated&unknown=polyfill&rum=0"></script>
-    <script src="https://${clientTopology.staticsDomain}/unpkg/react@16.7.0/umd/react#if(!${debug}).production.min#{else}.development#{end}.js"></script>
-    <script src="https://${clientTopology.staticsDomain}/unpkg/react-dom@16.7.0/umd/react-dom#if(!${debug}).production.min#{else}.development#{end}.js"></script>
+
+    <!-- Run the add-sentry tool to replace the link below. More information at: https://github.com/wix-private/fed-infra/tree/master/add-sentry -->
```
yairhaimo: Run `npx add-sentry` to add...

packages/create-yoshi-app/templates/client/src/index.vm
```diff
@@ -5,9 +5,18 @@
...
+    <script crossorigin="anonymous" src="https://js.sentry-cdn.com/d4e86aa8d11a4febad6a0093df51cc3d.min.js"></script>
```
yairhaimo:
```suggestion
```

PR #969, opened 2019-02-05
Fix checking if file is exist.

rudnitskih:
Original issue:
We had such file `mocha-setup.js`.

```
require('some-not-exsisting-file.js);

...other mocha configuration
```

Inside the yoshi we are cheking if file exsist using function `tryRequire`. If the error code is `MODULE_NOT_FOUND` we don't throw error anymore and just skip this file. In case `mocha-setup.js` require a file which doesn't exist yoshi doesn't produce any errors.

The fix is checking only required file.  The issue was discussed with Artem Yavorskyi.
### 🗺️ Test Plan
There are not tests here because I found onle two tests which located at `yoshi-helpers`.

### DIFF
--- a/packages/yoshi-helpers/utils.js
+++ b/packages/yoshi-helpers/utils.js
@@ -164,14 +164,14 @@ module.exports.toIdentifier = str => {
 };
 
 module.exports.tryRequire = name => {
+  let absolutePath;
   try {
-    return require(name);
-  } catch (ex) {
-    if (ex.code === 'MODULE_NOT_FOUND') {
-      return null;
-    }
-    throw ex;
+    absolutePath = require.resolve(name);
+  } catch (e) { // The module has not found
+    return null;
   }
+
+  return require(absolutePath);
 };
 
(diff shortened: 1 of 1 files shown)


rudnitskih:
@ranyitz I like your solution more!

ranyitz:
@rudnitskih Thanks, Can you update the PR and we'll merge?

rudnitskih:
@ranyitz updated. Do we need to merge this PR to master as well?

ronami:
Yeah, it's not a breaking change and I'll be happy if we can get it into `master`. We can then rebase it into v4 as well.

ranyitz:
@rudnitskih, I've set the base for this PR to master, but you need to add your changes to the `master` branch. (currently, they seem to be appended to `version_4`)

image

Just cherry pick your two commits after resetting it to origin master.

Something like that:
```bash
git reset origin/master --hard
git cherry-pick [1st commit hash]
git cherry-pick [2nd commit hash]
git push --force
```

And we should see only those two commits.

rudnitskih:
@ranyitz Please merge, sorry forgot about this PR

PR #978, opened 2019-02-10
Do not pack dist and test in yoshi main package

ranyitz:
### 🔦 Summary
Yoshi was packing and publishing directories like ~~`dist` and~~ `test` which made the package size larger. It can slow down install times of many devs with no apparent reason.

> I've found out later on that actually we're not running tests on the ci that publishes, so we don't have `dist` directory in the packed artifact, but it's good to remove it in any case.

### Before
image

### After
screen shot 2019-02-10 at 15 19 48

______

I've ended up removing most of the `__tests__` from the published code.

* I've used `.npmignore` because it made more sense in places where I've needed to remove one directory and files may move around and accidentally won't be published.

Also `files` didn't work for me in `create-yoshi-app`. If it's important enough for someone, he/she can investigate it.


packages/create-yoshi-app/.npmignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+./__tests__
```
ranyitz: So it won't work for inner `__tests__` directories

packages/create-yoshi-app/.npmignore
```diff
@@ -0,0 +1,3 @@
+node_modules
```
ronami: I think it's redundant adding `node_modules` to `.npmignore`, should already happen by default.

PR #981, opened 2019-02-12
[WIP] call app-server with non-concurrent runner

romansh-wixcom:
### 🔦 Summary
App sever runs in several processes what is a reason why sometimes we have up a few app-servers

### 🗺️ Test Plan
1. `yoshi start`
2. change a JS file
3. see a few `app-server` (`--entryPoint`) processes


romansh-wixcom:
This PR depends on PR to haste - https://github.com/wix/haste/pull/201

PR #985, opened 2019-02-14
Use legacy decorators by default to not break most of the projects.

yavorsky:
### 🔦 Summary
Enable `legacy` option for decorators by default.
We don't want to break almost all projects and allow people to use new decorators with `legacyDecorators: false`.

### DIFF
--- a/packages/babel-preset-yoshi/index.js
+++ b/packages/babel-preset-yoshi/index.js
@@ -23,6 +23,12 @@ module.exports = function(api, opts = {}) {
   const options = normaliseOptions(opts);
   const inWebpack = process.env.IN_WEBPACK;
   let { modules } = options;
+  const { legacyDecorators = true } = options;
+  let decoratorsBeforeExport;
+  // `legacyDecorators` and `legacy` are not compatible
+  if (!legacyDecorators) {
+    decoratorsBeforeExport = true;
+  }
   if (typeof modules === 'undefined') {
     modules = inWebpack ? false : DEFAULT_MODULES;
   }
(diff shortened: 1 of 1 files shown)


packages/babel-preset-yoshi/index.js
```diff
@@ -51,9 +57,11 @@ module.exports = function(api, opts = {}) {
       [
         require('@babel/plugin-proposal-decorators').default,
         {
+          // Use the legacy (stage 1) decorators syntax and behavior.
+          legacy: legacyDecorators,
           // Enable export after decorator syntax. It's also a part of the spec and tc39 is not made a decision about it.
           // Read more https://github.com/tc39/proposal-decorators/issues/69
-          decoratorsBeforeExport: true,
+          decoratorsBeforeExport,
```
ronami: Can this break existing behavior? It's currently set to `true` and since `legacyDecorators` is `true` by default, this is going to be `false` for most projects.

ronami:
Looks good 🏆 

Since #917 this will only help users that use `babel-preset-yoshi` without using Yoshi. Is this the use-case for `wix-style-react`?

packages/babel-preset-yoshi/index.js
yavorsky: `decoratorsBeforeExport` and `legacy` is contradictory options, you can't use both.
`legacy` already implements `decoratorsBeforeExport` logic, so with `decoratorsBeforeExport: false, legacy: true`, we won't break something in this part.

ranyitz:
Closing, please open if still relevant

PR #992, opened 2019-02-19
Improve app flow: Show server errors in the browser and refresh the browser on server changes

ronami:
### Summary

This PR refactors the `start-app` script to easily support sending and receiving signals from the user's server process and the webpack dev server. It now shows server compilation errors in the browser and refreshes the browser when the server has been updated/restarted.

I also covered the use-case when the user's server has died and needs to be resurrected.


packages/yoshi/src/server-process.js
```diff
@@ -0,0 +1,80 @@
...
+
+  onMessage(response) {
+    this._resolve(response);
+  }
+
+  end() {
+    this.child.kill();
+  }
+
+  send(message) {
```
netanelgilad: I know this may be an overkill. But saving the resolve and reject like that can cause race conditions if several messages are sent before a response is received. I guess this can happen in 2 consecutive changes that happen quickly I guess. Maybe a queue based solution would be more robust?

packages/yoshi/src/commands/start-app.js
```diff
@@ -143,42 +159,30 @@ module.exports = async () => {
...
+  try {
+    await serverProcess.initialize();
+  } catch (error) {
+    console.log();
+    console.log(
+      chalk.red(`Couldn't find a server running on port ${chalk.bold(PORT)}.`),
+    );
+    console.log(
+      chalk.red(
+        `Please check ${chalk.bold(cliArgs.server)} starts on this port.`,
```
netanelgilad: Since the server may fail to start (even if it tries to start on the correct port), maybe change the wording to say something like `Please check that your server starts up correctly and that it listens on the expected port`. You get the jest..

packages/yoshi/package.json
```diff
@@ -134,13 +134,15 @@
     "@babel/plugin-syntax-jsx": "^7.0.0",
     "@babel/preset-env": "^7.1.0",
     "@babel/runtime": "^7.1.5",
+    "@types/ejs": "^2.6.2",
```
netanelgilad: Why do we install `@types` packages?

ronami: Locally we symlink kitchensink project dependencies, a bit annoying.

packages/yoshi/src/commands/start-app.js
ronami: Sounds good, thanks 👍

PR #998, opened 2019-02-21
mocha-teamcity-reporter to ^2

roymiloh:
v1 is now broken with mocha@6, v2 has the fix.


roymiloh:
@ranyitz could be nice if he would fix v1 as well, but I don't feel like waiting for it now. Better to upgrade on our end.

PR #1006, opened 2019-02-24
Make jest default

### LINKED ISSUE #652: Make Jest the default test runner
# 🙋 Feature Request

### 😯 Current Behavior

Mocha is currently the default test runner

### 🤔 Expected Behavior

Jest should be the default test runner.
The change should also be done in the generators (remove explicit `--jest`).

yanivefraim:
fixes #652


docs/api/cli.md
```diff
@@ -167,18 +169,17 @@ Allow test debugging (works for mocha, jest & protractor), process won't start u
...
-```
+- Jest test setup:
+
+  Every other argument you'll pass to `yoshi test` will be forwarded to jest, For example:
+
+  `yoshi test --forceExit foo.spec.js`
+
+  Will run jest on `foo.spec.js` file and will apply [`forceExit`](https://jestjs.io/docs/en/cli#forceexit).
 
-- Note that when specifying multiple flags, only the first one will be considered, so you can't compose test runners (for now).
```
yanivefraim: This one is confusing / not correct since we can compose test runners. For example: `npx yoshi test --mocha --protractor`

@ronami - wdyt?

docs/api/cli.md
```diff
@@ -129,11 +129,17 @@ You can specify multiple entry points in your `package.json` file. This gives th
 
 ## test
 
+Executes tests using `jest` as default.
+
 ### options
 
+#### `--jest`
+
+Run tests with Jest - this is the default
```
ranyitz:
```suggestion
Run tests with Jest (_default_)
```

PR #1009, opened 2019-02-24
add fedops grafana sync to yoshi release

yaelhe:
Currently every project that has fedops is calling `fedops-api sync` in their build/release step. Added the call to yoshi release script.

Reference: https://github.com/wix/yoshi/pull/967#discussion_r257680447


packages/yoshi-config/paths.js
```diff
@@ -26,4 +27,5 @@ module.exports = {
   POM_FILE,
   STATS_FILE,
   TSCONFIG_FILE,
+  FEDOPSCONFIG_FILE,
```
ronami: I know it's minor but please I think `FEDOPS_CONFIG_FILE` helps to understand it a bit faster.

yaelhe: I agree. My intention was to be consistent with `TSCONFIG_FILE` above it. Maybe it should be `TS_CONFIG_FILE` too?

ronami: Good point, I think it can go both ways. I see `tsconfig` as a single entity, so it makes sense to call it `TSCONFIG_FILE`.

packages/yoshi/package.json
```diff
@@ -31,6 +31,7 @@
     "@svgr/webpack": "4.1.0",
     "@types/graphql": "14.0.7",
     "@wix/depkeeper-preset-yoshi": "1.0.17",
+    "@wix/fedops-grafana-api": "^3.0.0",
```
ranyitz: Please use fixed version (no `^`)

packages/yoshi/src/commands/release.js
```diff
@@ -11,6 +14,16 @@ const runner = createRunner({
...
+const syncFedopsGrafana = async () => {
+  if (await fs.pathExists(FEDOPSCONFIG_FILE)) {
+    const fedopsJson = await fs.readJson(FEDOPSCONFIG_FILE);
+    return fedopsGrapanaApi.sync(fedopsJson);
+  }
+};
+
+module.exports = runner.command(() => {
+  return Promise.all([
+    wnpm.prepareForRelease({ shouldBumpMinor }),
```
ranyitz: 1. Can those tasks be serial instead of parallel?
2. Should `syncFedopsGrafana` prevent a release? (In case of a failure) I don't know how graceful is it, but it will be able to prevent releases if an error will be thrown.

cc @ronami

packages/yoshi/src/commands/release.js
```diff
@@ -2,6 +2,9 @@ const wnpm = require('wnpm-ci');
 const { createRunner } = require('haste-core');
 const LoggerPlugin = require('../plugins/haste-plugin-yoshi-logger');
 const parseArgs = require('minimist');
+const fs = require('fs-extra');
+const { FEDOPSCONFIG_FILE } = require('yoshi-config/paths');
+const fedopsGrapanaApi = require('@wix/fedops-grafana-api');
```
ranyitz: image

This is a private dependency.

cc @ronami

yaelhe:
It was added to the CI scripts, superseding this.

PR #1027, opened 2019-02-26
Basic structure to make generators look better

ronami:
### 🔦 Summary

Still a work in progress, will update in a bit.

__Notice:__ I only updated the fullstack generator, if this makes sense we'll update the rest.


ronami:
Closing. We can use as a reference for future PRs.

PR #1028, opened 2019-02-26
add an option to load the config without validating it

ranyitz:
Currently, when requiring `yoshi-config` we get a config that validates itself for the first time that it loads.

We want to ensure that the same behavior remains but we do want to give an API for other programs that want to use `yoshi-config` without the validation functionality.

### The new API
```js
const loadConfig = require(`yoshi-config/loadConfig`);

const config = loadConfig({ validate: false, useCache: true });
```

### 🗺️ Test Plan
All tests should pass


packages/yoshi-config/loadConfig.js
```diff
@@ -0,0 +1,125 @@
...
+const validateConfig = require('./utils/validate-config');
+const YoshiOptionsValidationError = require('./utils/YoshiOptionsValidationError');
+const schema = require('./schema/yoshi-config-schema');
+
+let _config;
+
+const loadConfig = ({ validate, useCache } = { validate: false }) => {
+  let loadedConfig;
+
+  if (!useCache) {
```
ronami: Is there a reason not to use cache?

ranyitz: Not in our usecase, but if you're an app that wants to load multiple yoshi configs you'll need it

PR #1035, opened 2019-02-27
configure npm to not automatically add `^` to installed dependency

cowchimp:
### 🔦 Summary
Now that we're relying on renovate to update deps and not npm ranges, the default behavior when installing a new dependency should be to set as exact dependency.

### 🗺️ Test Plan
Existing suite


cowchimp:
merging my own PRs all like

pat on own back

PR #1040, opened 2019-02-27
Basic app flow migration

ronami:
### 🔦 Summary

This is a very basic migration guide for users that want to opt-into app flow during alpha/beta testing. We can, later on, close this PR and use its content as part of a standard migration guide.

### DIFF
--- a/APP_FLOW.md
+++ b/APP_FLOW.md
@@ -68,6 +68,10 @@ Finally, opt-into app flow by changing your `package.json` to include:
 
 ```diff
 {
+  "scripts": {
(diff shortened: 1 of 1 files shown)


APP_FLOW.md
```diff
@@ -0,0 +1,123 @@
...
+  return app;
+};
+```
+
+Finally, opt-into app flow by changing your `package.json` to include:
+
+```diff
+{
+  "scripts": {
+-   "start": "yoshi start --entry-point=index-dev.js"
```
ranyitz: It worth explaining about `--server`/`--entry-point` API change on a different comment above.

PR #1047, opened 2019-02-28
run eslint even after stylelint errors

### LINKED ISSUE #1039: Yoshi lint bails on stylelint, and does not continue to other linters
# 🐛 Bug Report

### 😯 Current Behavior

If there's a stylelint error, TSLint will not run (for example)

### 🤔 Expected Behavior

Should run all linters in all cases

yanivefraim:
fixes #1039 

wip, needs some refactoring


packages/yoshi/test/lint.spec.js
```diff
@@ -426,6 +426,41 @@ p {
       expect(res.stderr).to.contain('src/b.scss');
       expect(res.code).to.equal(1);
     });
+
+    it('should run eslint even if stylelint failed', () => {
```
ronami: I think it would be nice to have a similar test but with TSLint. Just so we cover that flow as well. Wdyt?

packages/yoshi/src/commands/lint.js
```diff
@@ -44,71 +44,85 @@ module.exports = runner.command(async () => {
...
     });
   }
 
   async function runEsLint(pattern) {
-    return printAndExitOnErrors(async () => {
-      if (prelint) {
-        await execa.shell(prelint, { stdio: 'inherit' });
-      }
+    if (prelint) {
+      await execa.shell(prelint, { stdio: 'inherit' });
```
ronami: We run `prelint` only if we run ESLint/TSLint but not if we only run Stylelint. I think we should run `prelint` if either of them runs. Also, instead of calling it from every function, I think we can refactor to call it just once if we have any files to lint. Wdyt?

PR #1053, opened 2019-03-03
Use sockjs instead of ipc for server HMR communication

ronami:
### 🔦 Summary

This PR attempts to fix a problem where some users don't start their server in the same process their `--server` is running. Instead, they spawn additional processes from that process that then start their server.

This is a problem with how server HMR works because it requires communication between Yoshi and the server bundle. Yoshi has to be the parent of the process that's running the server for it to work.

Instead of relying on IPC communication, this PR replaces it with web sockets so no matter how many processes are between Yoshi and the server, communication still works.

### DIFF
--- a/packages/yoshi/src/socket-server.js
+++ b/packages/yoshi/src/socket-server.js
@@ -6,7 +6,7 @@ module.exports = class SocketServer extends EventEmitter {
   constructor() {
     super();
 
-    this.connection = null;
+    this.connections = [];
 
     this.server = http.createServer();
     this.socket = sockjs.createServer();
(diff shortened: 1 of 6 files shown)


ronami:
I changed my implementation to support multiple connections instead of just one, I didn't see a use-case but the tests failed so eventually I did: When you're running `npx yoshi start` and then run your E2E tests (which start your server via `jest-yoshi.config.js`'s `server` prop), you have two instances of your server.

packages/yoshi/src/socket-server.js
```diff
@@ -0,0 +1,46 @@
...
+  send(message) {
+    this.connections.forEach(connection => {
+      connection.write(JSON.stringify(message));
+    });
+  }
+
+  async initialize() {
+    if (!this.server.listening) {
+      await new Promise((resolve, reject) => {
+        this.server.listen(4000, err => (err ? reject(err) : resolve()));
```
ranyitz: If we could verify that the port that we choose for the server is free it would be much better.

yanivefraim: This PR will fix an issue in one of our projects. Let's merge it like this and fix port issue later (I'll open an issue for that)

PR #1058, opened 2019-03-03
test improvements 

yanivefraim:
1. add `page.waitForNavigation` timeout: will allow us to see the exact place of failure (today we have a timeout for the whole test)

2. remove redundant `--jest` flag


ronami:
Looks good. I wanted to be able to set our timeout for all tests. I looked and I found this: https://github.com/GoogleChrome/puppeteer/blob/v1.12.2/docs/api.md#pagesetdefaultnavigationtimeouttimeout. Wdyt?

PR #1059, opened 2019-03-03
handle svg in tests (mocha,jest,jest-yoshi-preset) for react 15

ranyitz:
### 🔦 Summary
While returning a string is valid in React 16, it is not in React 15.
This PR should change the file transformer to a react 15 valid one.

### 🗺️ Test Plan
Tests need to be added after we'll change the app flow test to work against React 15 instead of React 16.

@ronami


ranyitz:
@ronami I agree, I think we should.

That's a big breaking change though, so version 4 could be a good opportunity.

PR #1066, opened 2019-03-04
faster TS for appflow, using incremental build

yanivefraim:
Typescript 2.7 introduced incremental "builder" API, which makes typescript compilation much much faster. fork-ts-checker-webpack-plugin added support for this feature starting alpha-6 version (CRA are using it, so we think it is safe enough).

One more important feature is the ability to refresh the page after a server error and still see the error overlay on the browser. This is possible by passing server errors to devServer, using devServer._stats = stats;


packages/yoshi/src/commands/start-app.js
```diff
@@ -105,6 +105,8 @@ module.exports = async () => {
   });
 
   serverCompiler.watch({ 'info-verbosity': 'none' }, async (error, stats) => {
+    // persist state for page refresh
```
ronami: I think we can explain it better as it's a bit hacky.

Issue #1082, opened 2019-03-06
On "yoshi start" open the browser window on a custom url/route


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior

When running `yoshi start` it opens the browser on `localhost:${PORT}/`

### 🤔 Expected Behavior

The user could configure what is the route which opens every time that `yoshi start runs`.

### 💁 Possible Solution

Add a CLI option (e.g. `--default-route=my-route`) that will cause yoshi to open the browser on `http://localhost:3000/my-route`.

I thought about `route` but we can also go with `--default-url` and give completely custom behavior to the first route that opens when running `yoshi start`.

### 🔦 Context

I work on a fullstack app that serves multiple routes, the route that I work on is not always `/` and I would love to configure a different route for every command.

For example `npm run start:bizmgr` should open a specific route.

ronami:
Good point. I like `--url` for simplicity, seeing `npx yoshi start --url=http://localhost:1234/hello` makes me understand that it would open this URL. Wdyt?

ronami:
Closed at #1084.

PR #1084, opened 2019-03-06
Support opening the browser on a different url with `--url`

ronami:
### 🔦 Summary

See #1082. This PR proposes to use `--url` as a flag to open the browser on a different URL.


packages/yoshi/src/commands/start-app.js
```diff
@@ -11,6 +12,7 @@ const cliArgs = parseArgs(process.argv.slice(2), {
   default: {
     server: 'index.js',
     https: false,
+    url: `http://localhost:${PORT}`,
```
ranyitz: Note that when `https` is `true` we should use `https` in the url.

ronami: Thanks, nice catch.

ronami:
> * We should update Yoshi docs with `--url` and maybe also explain about the use cases where it might be useful (according to #1082)
> * We can also update docs about https support we just added?

That's correct, app flow is missing on documentation, I hope to get to it today 🙏

PR #1090, opened 2019-03-10
Html Webpack Plugin MVP

ronami:
### 🔦 Summary

This PR introduces [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin). It will help remove redundant boilerplate from projects and allow us to support features like `initial` split chunks.

I learned and took inspiration from #805, it saved me a lot of work.

#### Points to consider

- I went with a convention of matching template name with entry name: If a template has the same name of a Webpack entry, Yoshi will add all of that entry's chunks into it. I don't see a reason for having more than one entry per template. **Does one exist?** I'll be happy for more info on this.

- We have two separate compilations: Minified and non-minified (debug). There is a possibility of them having a different number/names of chunks (not 1:1 mapping of adding `.min`). To solve that, and hopefully make it simpler, I decided to generate two distinct template files for every source template. This will require the developer to load `index.ejs` or `index.debug.ejs` (during debug). This is instead of how it currently is: By passing `debug` and checking inside the template (not necessary now).

- We create JavaScript bundles that contain the correct and full `publicPath` to their assets. It made sense to me to do the same for our templates: Templates already have full static paths to their static chunks. Templates no longer need to ask for their statics URL on standard conditions (and we can inject it during build time so no `erb` shenanigans would be required).

- I configured Webpack to use split chunks by default: Apps split all chunks (both, `initial` and `async`) while libraries only `async`. Once more apps use code splitting this can have a pretty big effect. For libraries to be able to enjoy `initial` chunks,  we can consider using [assets-webpack-plugin](https://github.com/ztoben/assets-webpack-plugin).

- In an effort to remove boilerplate I also configured HtmlWebpackPlugin to inject `polyfill.io`'s URL before the app's scripts. We can also use it as a platform to inject or change other parts of user's apps (sentry monitoring?).

- I removed the `splitChunks` config option: **Is still there a reason/use-case to configure `splitChunks` with anything other than the current default?**

- This is not going to work for those that use right-to-left CSS. We can support it later on (or completely remove the RTL CSS feature as I think it's not very common).

Thanks for reading,


netanelgilad:
Looks cool! The implementation looks really simple, but I think most of the questions around this are design/API related.

> I went with a convention of matching template name with entry name: If a template has the same name of a Webpack entry, Yoshi will add all of that entry's chunks into it. I don't see a reason for having more than one entry per template. Does one exist? I'll be happy for more info on this.

I'm also not sure about this. The convention makes sense and is probably a good default, but I'm pretty sure a use case would come up with a need to configure this (or maybe other parts). When I started working on this, one of the main issues I thought about is how they can configure the templates, because I think most would need to in some way. Here is an example: https://github.com/wix-private/inbox/blob/master/chat-widget/src/index.vm. This is probably not related to matching entry points to templates, but about config for this feature in general.

> We have two separate compilations: Minified and non-minified (debug). There is a possibility of them having a different number/names of chunks (not 1:1 mapping of adding .min). To solve that, and hopefully make it simpler, I decided to generate two distinct template files for every source template. This will require the developer to load index.ejs or index.debug.ejs (during debug). This is instead of how it currently is: By passing debug and checking inside the template (not necessary now).

I really like this solution. Note that this would require server changes, and would make migration harder. Something to consider.

> In an effort to remove boilerplate I also configured it to HtmlWebpackPlugin to inject polyfill.io's URL before the app's scripts. We can also use it as a platform to inject or change other parts of user's apps (sentry monitoring?).

This is pretty cool, but would also require some config or other considerations. Maybe best to leave to another issue/PR to keep this one clearer?

> This is not going to work for those that use right-to-left CSS. We can support it later on (or completely remove the RTL CSS feature as I think it's not very common). 

Thought so too, but when we asked around about removing support for this plugin a few projects replied that they are using it so apparently we need support for RTL css.

Awesome stuff! 🐩

packages/yoshi/config/webpack.config.js
```diff
@@ -622,12 +687,23 @@ function createClientWebpackConfig({
         ? [
             // https://github.com/webpack-contrib/mini-css-extract-plugin
             new MiniCssExtractPlugin({
-              filename: isDebug ? '[name].css' : '[name].min.css',
+              filename: isDebug
+                ? '[name].css'
+                : addHashToAssetName('[name].min.css'),
+              chunkFilename: isDebug
+                ? '[name].chunk.css'
+                : addHashToAssetName('[name].chunk.min.css'),
```
netanelgilad: check is name is unique

packages/yoshi-helpers/utils.js
```diff
@@ -212,15 +212,24 @@ const getProjectArtifactId = () => {
 
 module.exports.getProjectArtifactId = getProjectArtifactId;
 
+const getArtifactVersion = () => {
```
netanelgilad: Change to something that would make it clearer on what it does: `getArtifactVersionInCdnPath`?

ronami:
I made some changes to this PR so it can be merged as opt-in into version 4.

I plan on merging this into version 4 to test it on a few selected projects. We can, later on, enable other features that depend on this one like configuring split chunks with better defaults.

To opt-in, users will need to turn `experimentalHtmlFeature` to `true` in Yoshi's config. That will make Yoshi generate `ejs`/`vm` files with the strategy explained above and include a polyfill link.

It will also add a content hash to every file/chunk and set Webpack's `publicPath` to one that doesn't' use the project's version. That should make it so not all the cache isn't cleared between deployments.

packages/yoshi-config/loadConfig.js
```diff
@@ -91,6 +91,7 @@ const loadConfig = ({ validate, useCache } = { validate: false }) => {
     resolveAlias: getConfig('resolveAlias', {}),
     keepFunctionNames: getConfig('keepFunctionNames', false),
     umdNamedDefine: getConfig('umdNamedDefine', true),
+    experimentalHtmlFeature: getConfig('experimentalHtmlFeature'),
```
ranyitz:
```suggestion
    experimentalBuildHtml: getConfig('experimentalBuildHtml'),
```

packages/yoshi-helpers/utils.js
```diff
@@ -212,15 +212,24 @@ const getProjectArtifactId = () => {
 
 module.exports.getProjectArtifactId = getProjectArtifactId;
 
+const getArtifactVersionInCdnPath = () => {
+  if (project.experimentalHtmlFeature) {
```
ranyitz:
```suggestion
  if (project.experimentalBuildHtml) {
```

packages/yoshi-helpers/utils.js
```diff
@@ -212,15 +212,24 @@ const getProjectArtifactId = () => {
...
+    return 'dist';
```
ranyitz: There is a disambiguation between the local `dist` directory and the strategy of our system regarding the `dist` in the CDN path.

Worth adding a comment above it.

packages/yoshi/config/webpack.config.js
```diff
@@ -95,6 +98,14 @@ function exists(entry) {
   );
 }
 
+function addHashToAssetName(name) {
+  if (project.experimentalHtmlFeature) {
```
```suggestion
  if (project.experimentalBuildHtml) {
```

packages/yoshi/config/webpack.config.js
```diff
@@ -95,6 +98,14 @@ function exists(entry) {
...
+    return name.replace('[name]', `[name].[contenthash:8]`);
```
```suggestion
    return name.replace('[name]', '[name].[contenthash:8]');
```

packages/yoshi/config/webpack.config.js
```diff
@@ -445,13 +460,14 @@ function createCommonWebpackConfig({
                 test: /\.(j|t)sx?$/,
               },
               use: [
-                require.resolve('@svgr/webpack'),
+                '@svgr/webpack',
```
ranyitz: Is this intentional?

packages/yoshi/config/webpack.config.js
```diff
@@ -606,6 +626,47 @@ function createClientWebpackConfig({
     plugins: [
       ...config.plugins,
 
+      // https://github.com/jantimon/html-webpack-plugin
+      ...(project.experimentalHtmlFeature && fs.pathExistsSync(config.context)
```
```suggestion
      ...(project.experimentalBuildHtml && fs.pathExistsSync(config.context)
```

packages/yoshi/config/webpack.config.js
```diff
@@ -622,12 +683,23 @@ function createClientWebpackConfig({
...
+                : addHashToAssetName('[name].min.css'),
+              chunkFilename: isDebug
+                ? addHashToAssetName('[name].chunk.css')
+                : addHashToAssetName('[name].chunk.min.css'),
             }),
             // https://github.com/wix-incubator/tpa-style-webpack-plugin
             ...(project.enhancedTpaStyle ? [new TpaStyleWebpackPlugin()] : []),
             // https://github.com/wix/rtlcss-webpack-plugin
-            new RtlCssPlugin(isDebug ? '[name].rtl.css' : '[name].rtl.min.css'),
+            ...(!project.experimentalHtmlFeature
```
```suggestion
            ...(!project.experimentalBuildHtml
```

muteor:
Cool to see this being worked on :)

> I removed the splitChunks config option: Is still there a reason/use-case to configure splitChunks with anything other than the current default?

I would keep this option as I was planning on using a vendors chunk doing regex like `/[\\/]node_modules[\\/](?!(@wix|playable))/`. This was to split out the common least updated node_modules packages while keeping the frequently changing @wix modules in the main chunks. The idea being that I can long cache the vendors chunk using the hashing webpack does.

PR #1101, opened 2019-03-12
Remove an unused and undocumented way to disable thread optimization in Typescript

ronami:
### 🔦 Summary

Small maintenance PR: I didn't find any use-cases for it.


cowchimp:
This was a contribution to Yoshi I made while in ADI.
I can confirm they switched to Yoshi 3 and are no longer using this flag.
Thanks.

PR #1104, opened 2019-03-12
Bump stylable patch version to fix a :global() scope issue

mykas:
### 🔦 Summary

wix-style-react is using yoshi version 4 alpha already. We have encountered an issue with stylable when using `:global()` scope. Next patch version of stylable suppose to fix this :) 

https://github.com/wix/stylable/releases/tag/%40stylable%2Fcore%401.0.5


ronami:
Thanks for the PR.

Does this PR introduce any breaking changes? Also, should we also include it in v3?

mykas:
Stylable only fixed the bug in this patch version and updated their internal dependencies for react/types etc to stable versions. No other changes. 

it should be included in V3 too.

PR #1105, opened 2019-03-12
Update jest transform

yanivefraim:
In TS projects we want TS to transpile both TS and JS files.

### DIFF
--- a/packages/jest-yoshi-preset/jest-preset.js
+++ b/packages/jest-yoshi-preset/jest-preset.js
@@ -2,6 +2,7 @@ const fs = require('fs');
 const globby = require('globby');
 const { envs } = require('./constants');
 const globs = require('yoshi-config/globs');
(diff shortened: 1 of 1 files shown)


ronami:
I think we can add a test for that pretty easily with the new integration tests. Wdyt?

yanivefraim:
> I think we can add a test for that pretty easily with the new integration tests. It should test the use-case of having a JavaScript file in a TypeScript project that will need transpilation. Wdyt?

cool. I would add another test for (already merged) `regenerator-runtime`

ronami:
Just realized, we do the same mistake when we bundle with Webpack: `.js` files go through `babel-loader` and `.tsx?` file go through ts-loader.

ronami:
Closing for now. This will be a breaking change that we plan on including in the next version.

PR #1115, opened 2019-03-14
Server side rendering: generate long/short class names in prod/dev respectively

ronami:
### 🔦 Summary

Currently, universal apps always get long class names generated both, locally and on production. This PR changes it to always generate short class names in production and short ones locally.

This change **may** make it harder to debug production code as it will always contain short class names even in the debug bundle.


packages/yoshi/config/webpack.config.js
```diff
@@ -127,7 +127,7 @@ const getStyleLoaders = ({
   const cssLoaderOptions = {
     camelCase: true,
     sourceMap: !!separateCss,
-    localIdentName: isDebug ? localIdentName.long : localIdentName.short,
+    localIdentName: isProduction ? localIdentName.long : localIdentName.short,
```
```suggestion
    localIdentName: !isProduction ? localIdentName.long : localIdentName.short,
```

PR #1122, opened 2019-03-17
Verify that all yoshi related dependencies are in the same major version

### LINKED ISSUE #1119: Verify that versions of Yoshi and related packages match
### Summary

Sometimes Yoshi is installed in one version and some of its related packages (`jest-yoshi-preset`, `yoshi-style-dependencies` and `yoshi-angular-dependencies`) are installed in a different version. This can happen to a lot of users when we bump a major version and they need to remember to change all of them. If the versions don't match it can cause bugs that are hard to detect.

To help detect these problems easier, we can check that Yoshi and those packages are on the same version and fail with a clear error otherwise. We can do this check in every command (`start`, `build`, `test`).

ronami:
### 🔦 Summary

Fixes #1119.


packages/yoshi/src/verify-dependencies.js
```diff
@@ -0,0 +1,40 @@
...
+const relatedPackages = [
+  'jest-yoshi-preset',
+  'yoshi-style-dependencies',
+  'yoshi-angular-dependencies',
+];
+
+module.exports = async () => {
+  const outdatedPackages = relatedPackages
+    .map(packageName => path.join(packageName, 'package.json'))
+    .map(packageName => importCwd.silent(packageName))
```
ranyitz: One of those?

```suggestion
    .map(packageJsonPath => importCwd.silent(packageJsonPath))
```

```suggestion
    .map(importCwd.silent)
```

packages/yoshi/src/verify-dependencies.js
```diff
@@ -0,0 +1,40 @@
...
+      const diff = pkg && semver.diff(pkg.version, version);
+
+      if (diff) {
+        return diff.includes('major');
+      }
+    });
+
+  if (outdatedPackages.length > 0) {
+    console.log(chalk.red('Command failed.\n'));
+    console.log(chalk.red('Some packages related to Yoshi are outdated.\n'));
```
ranyitz: This will raise a warning for cases when those related packages are in a major version this is greater than Yoshi.

image

I don't currently think about a use case that this would happen, but we can either add a `semver.gt` check or change the warning message to something more generic if we want to prevent users from using different versions of Yoshi's dependencies.

packages/yoshi/src/verify-dependencies.js
```diff
@@ -0,0 +1,40 @@
+const path = require('path');
+const chalk = require('chalk');
+const semver = require('semver');
+const importCwd = require('import-cwd');
+
+const { version } = require('../package.json');
```
ranyitz: Ease readability later on

```suggestion
const { version: yoshiVersion } = require('../package.json');
```

packages/yoshi/src/verify-dependencies.js
```diff
@@ -0,0 +1,40 @@
...
+        return diff.includes('major');
+      }
+    });
+
+  if (outdatedPackages.length > 0) {
+    console.log(chalk.red('Command failed.\n'));
+    console.log(chalk.red('Some packages related to Yoshi are outdated.\n'));
+    console.log(
+      chalk.red(
+        `Please install them in the same version of Yoshi (${version}):\n`,
```
ranyitz:
```suggestion
        `Please install them in the same version of Yoshi (${yoshiVersion}):\n`,
```

packages/yoshi/src/verify-dependencies.js
```diff
@@ -0,0 +1,40 @@
...
+  'yoshi-style-dependencies',
+  'yoshi-angular-dependencies',
+];
+
+module.exports = async () => {
+  const outdatedPackages = relatedPackages
+    .map(packageName => path.join(packageName, 'package.json'))
+    .map(packageName => importCwd.silent(packageName))
+    .filter(pkg => {
+      const diff = pkg && semver.diff(pkg.version, version);
```
ranyitz:
```suggestion
      const diff = pkg && semver.diff(pkg.version, yoshiVersion);
```

PR #1123, opened 2019-03-17
migrate fullstack template to use app flow

ranyitz:
### 🔦 Summary
Followed [this guide](https://github.com/wix/yoshi/blob/3eafa61124327d868d7ec8b80d69c4083d380982/APP_FLOW.md) and migrate the fullstack and fullstack-typescript templates to use the new __app flow__

### 🗺️ Test Plan
All tests pass

### 📓 Notes
I used `"bootstrap-hot-loader": "^3.19.3"` only because it's the only available version, I think it worth changing it to a perversion of 4 when it will be available.


packages/create-yoshi-app/templates/fullstack-typescript/{%packagejson%}
```diff
@@ -8,7 +8,7 @@
     "email": "{%authorEmail%}"
   },
   "scripts": {
-    "start": "yoshi start --entry-point=index-dev.js",
+    "start": "yoshi start --server=index-dev.js",
```
hugebdu: @ranyitz r u sure there's `--server` option? [docs](https://wix.github.io/yoshi/docs/api/cli#start) do not mention anything about it, and when I try it I get
```
➜  test-server git:(master) ✗ npm start

> @wix/test-server@1.0.0 start /private/var/folders/dy/_06bhr553ms1nj0vzx5ydldxqs80sf/T/2da838486cb9bed636f0d449b75f2bfc/test-server
> yoshi start --server=index-dev.js

error: unknown option `--server'
```

ranyitz: @hugebdu 

`--server` option exists only in the new flow called - "App Flow"

See the possible options [here](https://wix.github.io/yoshi/docs/guides/app-flow#npx-yoshi-start)

As this flow is getting more and more to the mainstream, and because this is probably the only thing which is not yet in the docs. I agree that it's a good idea to add it there (and also implement its behavior to the old flow).

I'll open an issue.

Thanks for bringing that up!

hugebdu: @ranyitz 
I'm not much familiar with flows I guess :)
anyway, I'm going to add `index-dev.js` to server template.
Should I stick to `--entry-point`?

ranyitz: Currently yes, when we'll add the `--server` option to the "old flow" we'll also change it in the `server` template.

This is the [App flow](https://wix.github.io/yoshi/docs/guides/app-flow), it's about targeting apps and giving better developer experience.

I'm not sure yet, but maybe we'll want a "Server Flow". Or a specific experience for creating a server application in Wix.

hugebdu: @ranyitz 10x

ranyitz: I'll elaborate,

In the past, Yoshi took care of every type of application (Server, Library, Fullstack, universal, Client, Component Library and more) without specifying it explicitly.

It created a lot of problems when some features from a specific flow collide with others, or the users misused them.

We've decided to aim towards Zero Configuration flows for each type of application. The first one we're tackling is the `App` (client & full stack) which are our biggest chunk of users.

As creating node server becoming more and more common, I guess we'll need to work together into tailor its experience for users.

Issue #922, opened 2019-01-17
`close of undefined` causes my tests to fail sometimes (flaky)


sidoruk-sv:
# 🐛 Bug Report

Flaky both on local machine and CI (can provide link to failed CI builds in Slack)

```
{ ShellJSInternalError: EEXIST: file already exists, mkdir '/Users/username/projects/projectname/target/logs'
    at Object.mkdirSync (fs.js:753:3)
    at mkdirSyncRecursive (/Users/username/projects/projectname/node_modules/shelljs/src/mkdir.js:24:8)
    at /Users/username/projects/projectname/node_modules/shelljs/src/mkdir.js:81:9
    at Array.forEach (<anonymous>)
    at Object._mkdir (/Users/username/projects/projectname/node_modules/shelljs/src/mkdir.js:59:8)
 FAIL   e2e  test/e2e/app.spec.tsis/projects/projectname/node_modules/shelljs/src/common.js:384:25)
  ● Test suite failed to run

    TypeError: Cannot read property 'close' of undefined

      at PuppeteerEnvironment.teardown (node_modules/jest-environment-yoshi-puppeteer/index.js:39:28)
          at Generator.throw (<anonymous>)

```
### 🌍 Your Environment

```
  System:
    OS: macOS 10.14.2
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: 1.13.0 - ~/.nvm/versions/node/v10.15.0/bin/yarn
    npm: 6.5.0 - ~/.nvm/versions/node/v10.15.0/bin/npm
    Watchman: Not Found
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 64.0.2
    Safari: 12.0.2
  npmPackages:
    @storybook/core:  4.1.6 
    @storybook/react: ^4.1.4 => 4.1.6 
    babel-preset-yoshi:  3.17.0 
    eslint-config-yoshi:  3.25.1 
    eslint-config-yoshi-base:  3.25.1 
    jest-environment-yoshi-bootstrap:  3.23.0 
    jest-environment-yoshi-puppeteer:  3.24.1 
    jest-yoshi-preset: ^3.24.1 => 3.24.1 
    storybook: Not Found
    tslint-config-yoshi: ^3.22.5 => 3.22.5 
[...]

PR #1132, opened 2019-03-17
adds sensible defaults for bootstrap environment

ranyitz:
### 🔦 Summary
This PR attempts to solve https://github.com/wix/yoshi/issues/922 and also to reduce boilerplate in fullstack (bootstrap) projects.

The end goal is to be able and let the user bootstrap the server and running the config emitter without extra boilerplate of custom options, while the preset/environment will take care of supplying the correct params on `process.env`.

While this is the first step, after merging https://github.com/wix-platform/wix-node-platform/pull/2154 and another one to take care of `process.env` override in the testkit the configuration in the user's project should look like this:

```js
module.exports.emitConfigs = () => {
  const emitter = configEmitter();
  return emitter.val('scripts_domain', 'static.parastorage.com').emit();
};

module.exports.bootstrapServer = () => {
  return testkit.app('./index');
};
```

### 🗺️ Test Plan
All templates tests pass

### 📔 Note
~~We can (and should) use the same methodology for `start` in order to reduce boilerplate from `index-dev` setup files. (This PR won't include it)~~

I've ended up changing it in the same PR because we use the `environment` from both start and test.


packages/create-yoshi-app/templates/fullstack-typescript/environment.js
```diff
@@ -5,24 +5,18 @@ const configEmitter = require('@wix/wix-config-emitter');
...
+module.exports.bootstrapServer = () => {
   return testkit.app('./index', {
-    env: {
-      PORT: port,
-      MANAGEMENT_PORT: managementPort,
-      NEW_RELIC_LOG_LEVEL: 'warn',
-      APP_CONF_DIR: appConfDir,
-      DEBUG: '',
-    },
+    env: process.env,
```
ranyitz: There is currently a constraint to this weird (but better) api.

bootstrap's defaults override `process.env`'s

https://github.com/wix-platform/wix-node-platform/blob/master/bootstrap/wix-bootstrap-testkit/lib/embedded-server.js#L16

packages/create-yoshi-app/templates/fullstack-typescript/environment.js
```diff
@@ -5,24 +5,18 @@ const configEmitter = require('@wix/wix-config-emitter');
 
 // take erb configurations from source folder, replace values/functions,
 // remove the ".erb" extension and emit files inside the target folder
-module.exports.emitConfigs = ({ targetFolder }) => {
+module.exports.emitConfigs = () => {
   const emitter = configEmitter({
-    sourceFolders: ['./templates'],
-    targetFolder,
+    sourceFolders: [process.env.APP_TEMPL_DIR],
+    targetFolder: process.env.APP_CONF_DIR,
```
ranyitz: When [this PR](https://github.com/wix-platform/wix-node-platform/pull/2154) will be merged we could remove those...

packages/yoshi/src/tasks/app-server/index.js
```diff
@@ -46,6 +46,11 @@ function initializeServerStartDelegate({
 
     const env = Object.assign(defaultEnv, process.env, {
       PORT: port,
+      MANAGEMENT_PORT: Number(port) + 1,
```
ranyitz: I wasn't sure whether this port is a number or a string ...

Feel free to check and remove the `Number` if needed.

packages/jest-environment-yoshi-bootstrap/index.js
```diff
@@ -14,6 +33,8 @@ module.exports = class BootstrapEnvironment extends NodeEnvironment {
           getPort,
           staticsUrl: projectConfig.servers.cdn.url,
           appConfDir,
+          appLogDir,
```
yanivefraim: Not specifically related to this PR, but since we are not using those in the generators, I think we might want to add it to our docs. WDYT?

yanivefraim:
Super cool 🥇 

Will definitely make our lives much easier! Thanks!

packages/jest-environment-yoshi-bootstrap/index.js
ranyitz: I'm not sure that those params are documented at all.

I'm sure that it will be good to document them.
I've added them as backward compatibility.

PR #1137, opened 2019-03-18
replaced enzyme with react-testing-library

saarkuriel:
### 🔦 Summary
Replaced testing library from `enzyme` to `react-testing-library` in the following generators:

* `client` 
* `client-typescript`
* `server`
* `server-typescript`

### 🗺️ Test Plan
...

### DIFF
--- a/packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
+++ b/packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
@@ -1,21 +1,19 @@
-import React from 'react';
-import { mount } from 'enzyme';
+import * as React from 'react';
+import { render, cleanup } from 'react-testing-library';
 import { I18nextProvider } from 'react-i18next';
 import App from './App';
 import i18n from '../__mocks__/i18n';
 
 describe('App', () => {
-  let wrapper;
-
-  afterEach(() => wrapper.unmount());
+  afterEach(cleanup);
 
   it('renders a title correctly', () => {
-    wrapper = mount(
+    const { getByTestId } = render(
       <I18nextProvider i18n={i18n}>
         <App />
       </I18nextProvider>,
     );
 
-    expect(wrapper.find('h2')).toHaveLength(1);
+    expect(getByTestId('app-title')).not.toBeNull();
   });
 });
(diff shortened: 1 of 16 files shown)


netanelgilad:
@saarkuriel From your experience, what do you think about the setup mentioned in [`react-testing-library` docs](https://testing-library.com/docs/react-testing-library/setup)? 

Maybe we should include it in the generated projects? The cleanup stuff looks useful.

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
```diff
@@ -1,21 +1,19 @@
-import React from 'react';
-import { mount } from 'enzyme';
+import * as React from 'react';
```
ranyitz: Is that intentional?

packages/create-yoshi-app/templates/fullstack/src/components/App/App.spec.js
```diff
@@ -1,21 +1,21 @@
...
-    wrapper = mount(
+    const { queryByTestId } = render(
       <I18nextProvider i18n={i18n}>
         <App />
       </I18nextProvider>,
     );
 
-    expect(wrapper.find('h2')).toHaveLength(1);
+    expect(queryByTestId('app-title')).not.toBeNull();
+    expect(queryByTestId('app-title')).toBeDefined();
```
ranyitz: I think that's can be removed

packages/create-yoshi-app/templates/client/src/components/App/App.spec.js
saarkuriel: nope. fixing

packages/create-yoshi-app/templates/fullstack/src/components/App/App.spec.js
saarkuriel: correct. removing

saarkuriel:
> @saarkuriel From your experience, what do you think about the setup for `jest` mentioned in [`react-testing-library` docs](https://testing-library.com/docs/react-testing-library/setup)?
> 
> Maybe we should include it in the generated projects? The cleanup stuff looks useful.

I wasn't familiar with this, it looks great! On the one hand it looks like the best place to add it to reduce boilerplate is in `jest-yoshi-preset`, but since `jest-yoshi-preset` is a separate package it will break those who don't have `react-testing-library`. 

One possible solution is to add it in each template we just changed in `jest-yoshi.config`.

WDYT?

ranyitz:
@netanelgilad @saarkuriel I think [that's](https://testing-library.com/docs/react-testing-library/setup#cleanup) a great idea to reduce boilerplate!

I've opened a [separate issue](https://github.com/wix/yoshi/issues/1139) so we could merge this one and continue working on the other.

PR #1138, opened 2019-03-18
Update business manager template to use dynamic import

yurynix:
### 🔦 Summary
This PR updates biz-mgr templates to use dynamic import instead of `react-module-container` for loading of the actual module.

### 🗺️ Test Plan
- run `npm run create-yoshi-app:dev`
- Choose biz-mgr babel template
- assert created project's tests are passing `npm test`, and you are able to start the project `npm start` (you'll need `git init` for it)


packages/create-yoshi-app/templates/business-manager-module-typescript/{%packagejson%}
```diff
@@ -60,7 +60,7 @@
     },
     "yoshi": {
       "entry": {
-        "{%projectName%}-module": "./module",
+        "{%projectName%}-module": "./module.tsx",
```
ronami: is this needed?

packages/create-yoshi-app/templates/business-manager-module-typescript/{%packagejson%}
```diff
@@ -8,7 +8,7 @@
     "email": "{%authorEmail%}"
   },
   "scripts": {
-      "start": "yoshi start --entry-point ./dist/test/dev/server.js",
+      "start": "PORT=5000 yoshi start --entry-point ./dist/__tests__/dev/server.js",
```
ronami: any reason not to stay with the default?

yurynix: Yoshi would wait for port 3000 which is never available, and hangs. biz-mgr convention is to use port 5000 and that's what their testkit does by default.

packages/create-yoshi-app/templates/business-manager-module-typescript/{%packagejson%}
yurynix: It's needed for the React component `BMLazyComponent`, otherwise it doesn't compile

PR #1140, opened 2019-03-18
Add Stylable Webpack plugin project level hashing

NitayRabi:
### 🔦 Summary
- Due to Stylable build optimisation changes between projects there are collisions causing issues with several Stylable runtimes running on the same page.
- Adding namespaces per project should resolve these issues.

### 🗺️ Test Plan
...


packages/yoshi/config/webpack.config.js
```diff
@@ -115,6 +118,25 @@ const entry = project.entry || project.defaultEntry;
 
 const possibleServerEntries = ['./server', '../test/dev-server'];
 
+// Namespace factory for stylable hashing.
+function resolveNamespaceFactory(prefix) {
```
yurynix: Can we import https://github.com/wix/stylable/blob/df401015ddad962e37342e2bb8cc73295019e5d1/packages/node/src/resolve-namespace.ts#L6

from `@stylable/node` instead?

yurynix:
Hi!
Thank you for the PR!
How can we reproduce the issue and test the solution?

packages/yoshi/config/webpack.config.js
NitayRabi: Yes of course - it's a new function they added - I'll import it from there.

NitayRabi:
> Hi!
> Thank you for the PR!
> How can we reproduce the issue and test the solution?

Quite herd to reproduce the issue - it's seen when multiple WSR builds are used with a different mix of components.

but after this change each project should get a different hashed `st-id` attribute for the `<style>` tags

packages/yoshi/config/webpack.config.js
yurynix: Overall I'm not sure it's yoshi's problem to solve,
it sounds to me like a Stylable issue, maybe it's best to solve it there without any `yoshi` involvment?
What do you think?

NitayRabi: Well it's something that should be configured in the Webpack plugin and not set as a default configuration in Stylable IMO - @tomrav @barak007 your opinion ?

yurynix: Why it shouldn't be the default?
and if it shouldn't be the default, why not provide a configuration such as:
```
useProjectNamespacing: true|false
```

or maybe:
```
namespacePrefix: process.cwd()
```

yurynix: Gave it some more look, IMO it should be done here:
https://github.com/wix/stylable/blob/0a68832f1aaee9095edd51fa23692448fd6ac583/packages/webpack-plugin/src/StylableWebpackPlugin.js#L84
import `resolveNamespaceFactory` instead of `resolveNamespace`, and call it with `process.cwd()`

tomrav: Using a custom resolver is only required in complex multi-part builds, when you want to enforce prefixes in your generated css classes. By default most Stylable projects do not require this ability, and work out of the box using the provided resolver.

The factory method provided should be used, we released it in a new version yesterday especially for this purpose. You can pass it the `ROOT_DIR` constant already created above in the file.

packages/yoshi/config/webpack.config.js
```diff
@@ -8,6 +8,9 @@ const CaseSensitivePathsPlugin = require('case-sensitive-paths-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const TerserPlugin = require('terser-webpack-plugin');
 const StylableWebpackPlugin = require('@stylable/webpack-plugin');
+const {
+  resolveNamespaceFactory,
+} = require('@stylable/node/dist/resolve-namespace');
```
tomrav:
```suggestion
} = require('@stylable/node');
```

Should also work.

ronami:
Is this a breaking change? Should we apply it to Yoshi v3 too?

Thanks

NitayRabi:
> Is this a breaking change? Should we apply it to Yoshi v3 too?
> 
> Thanks

Not a breaking change (after consulting with @tomrav) - should be fixed in both Yoshi v2 and v3

packages/yoshi/config/webpack.config.js
```diff
@@ -654,6 +657,7 @@ function createClientWebpackConfig({
         generate: {
           runtimeStylesheetId: 'namespace',
         },
+        resolveNamespace: resolveNamespaceFactory(ROOT_DIR),
```
yurynix: What you think about changing that to package name and `project.name`?
`ROOT_DIR` will probably always change on CI.

ronami:
@tomrav didn't we solve this in #958?

packages/yoshi/config/webpack.config.js
NitayRabi: Looking at the implementation of - https://github.com/wix/stylable/blob/master/packages/node/src/resolve-nmespace.ts
`ROOT_DIR` is used to fetch `package.json` and get the name and version from there.

PR #1141, opened 2019-03-19
support environment parameters also in app flow

ranyitz:
### 🔦 Summary
After [this PR](https://github.com/wix/yoshi/pull/1132) we're assuming in our fullstack template that the server will start with envirnoment parameters (in order to spare boilerplate from the user)

The `start-app` which is the start command of the `app flow` has a different logic to running the server.

This PR adds the same parameters in order to make it work also using the app flow.


packages/yoshi/src/server-process.js
```diff
@@ -30,6 +30,11 @@ module.exports = class Server {
         ...process.env,
         NODE_ENV: 'development',
         PORT,
+        MANAGEMENT_PORT: Number(PORT) + 1,
```
ranyitz: `Number` is probably not crucial here. We can refactor this whole thing later on and maybe reuse it between the 2 servers

ranyitz:
fullstack templates are now broken in `create-yoshi-app` I'm merging in order to fix them.

If we want to improve this implementation (and we probably should) we can do it after we release the fix.

PR #1147, opened 2019-03-20
Simplify index.js for fullstack projects

ranyitz:
### 🔦 Summary
Inspired by @roymiloh https://github.com/wix-private/fed-infra/tree/master/wix-bootstrap-renderer#usage It would be less bloated like that.

It's a file that no one should change in any case, rather be small and simple


ronami:
Looks like this went in #1148. If this is still relevant re-open this PR.

PR #1155, opened 2019-03-20
preserve backwards compatibility for users that uses bootstrap's defaults

ranyitz:
### 🔦 Summary
While https://github.com/wix/yoshi/pull/1132 was intended to improve the experience for our users, it also broke users who used node platform defaults.

Those users are not using `index-dev` and not generating the config each `npm start`. Instead, they hold a fixture of the config located in `./test/configs`.

In order to support both use cases @ronami suggested that we'll check for the existence of `./test/configs` and assume that when it located there, the user probably wants to use it as default.

This PR adds this check and configure the default `APP_CONFIG_DIR` to support the old use case.

We can, later on, deprecate it if we'd like.


ranyitz:
Closing due to https://github.com/wix/yoshi/pull/1153

PR #1160, opened 2019-03-20
Add to our jest preset a graphql transform

### LINKED ISSUE #1080: Support importing a GraphQL file in Jest
In production code, we use a Webpack loader for importing GraphQL files. In Jest, we need a Jest transformer to do that. See [here](https://www.apollographql.com/docs/react/recipes/webpack.html#Jest) for more details.

** We already have failing tests for that (thanks @ronami for writing those amazing tests!!), all we need is to add the [transformer](https://github.com/remind101/jest-transform-graphql):

- [Javascript failing test](https://github.com/wix/yoshi/blob/version_4.x/test/projects/kitchensink-javascript/src/components/features/graphql-inclusion.spec.js#L7)

- [TS failing test](https://github.com/wix/yoshi/blob/version_4.x/test/projects/kitchensink-typescript/src/components/features/graphql-inclusion.spec.tsx#L7)

yurynix:
### 🔦 Summary
Closes #1080

### 🗺️ Test Plan
Watch the test pass 🙂


package.json
```diff
@@ -21,7 +21,8 @@
     "test:unit": "npx jest --config='./jest.unit.config.js'",
     "test:app-flow": "node scripts/runKitchensink.js",
     "test-babel-plugin-transform-hmr-runtime": "npm test --prefix ./packages/babel-plugin-transform-hmr-runtime",
-    "website:dev": "npm start --prefix './website'"
+    "website:dev": "npm start --prefix './website'",
+    "clean": "lerna clean"
```
ronami: Really small, but how helpful is it? I think that running `npx lerna clean` is simple enough, and I prefer to only have scripts that we have to keep. Wdyt?

yurynix: I don't have a strong feelings about it,
however I do find it useful to have all "useful" (in the project context) scripts as npm scripts.

PR #1164, opened 2019-03-21
Update app-flow docs for client, add section about installing source-map-support

ronami:
### 🔦 Summary

This PR updates the docs with instructions on how to migrate a client project to app flow. It also adds a section that requires users to install `map-support-support`. We use this dependency to show correct stack traces and rely on it being installed as a production dependency.

I'm not entirely sure about this: We can also change https://github.com/wix/yoshi/blame/version_4.x/packages/yoshi/config/webpack.config.js#L819 to fail gracefully if it wasn't able to find `source-map-support`. This way, if someone forgot to add it, source maps won't work but the server will still be up.

P.S: Also updated generators.

Wdyt?


packages/yoshi/config/source-map-support.js
```diff
@@ -0,0 +1,7 @@
+try {
+  require('source-map-support').install();
+} catch (e) {
+  console.error(
+    'Cannot find "source-map-support", stack traces may appear without source maps',
```
ranyitz: can instruct to `npm i --save source-map-support`

docs/guides/app-flow.md
```diff
@@ -82,18 +92,10 @@ import wixExpressRequireHttps from '@wix/wix-express-require-https';
...
 ```
 
-Finally, opt-into app flow by changing your `package.json` to include:
+We use `source-map-support` internally so stack traces show locations in your source files. To work in production (with New Relic monitoring), please install `source-map-support` under `dependencies`:
 
-```diff
-{
-  "scripts": {
--   "start": "yoshi start --entry-point=index-dev.js"
-+   "start": "yoshi start --server=index-dev.js"
```
yanivefraim: just making sure - we removed this part because we are backward compatible?

ronami: Yes, we can deprecate it but I think having a simple migration, or at least, as simple as we can, is helpful.

packages/yoshi/package.json
```diff
@@ -109,6 +109,7 @@
     "stylelint": "9.10.1",
     "stylelint-config-yoshi": "4.1.0-rc.3",
     "stylelint-prettier": "1.0.6",
+    "source-map-support": "0.5.11",
```
yanivefraim: why is this needed?

PR #1176, opened 2019-03-25
Setup puppeteer env only for E2E project

### LINKED ISSUE #1134: Do not throw an error telling to run `npm start` when it's not actually needed.
# 🙋 Feature Request

### 😯 Current Behavior

When using `jest-yoshi-preset` we verify that the user has CDN up before running the tests and throw an error in order to give a descriptive message indicating on what the user should do.

https://github.com/wix/yoshi/blob/5e91a64e316dd425e5bcb32e5163be66c20bf64c/packages/jest-environment-yoshi-puppeteer/globalSetup.js#L80-L86

The problem with this behavior is that it happens in cases when there you don't run `e2e` tests, for example, when running `server` tests.

Currently, it also happens when running component tests, it will solve after solving [this issue](https://github.com/wix/yoshi/issues/1133)

### 🤔 Expected Behavior

Do not throw an error telling to run `npm start` when it's not actually needed.

### 💁 Possible Solution

Instead of throwing an error during `globalSetup` we will wait for our tests to call the CDN (`localhost:3200`) in case a not found error returns (The CDN is down) we can add an informative error.

We have the [user's page](https://github.com/wix/yoshi/blob/5e91a64e316dd425e5bcb32e5163be66c20bf64c/packages/jest-environment-yoshi-puppeteer/index.js#L32) and we can use it in order to listen to errors/intercept requests.

yurynix:
### 🔦 Summary
This PR moves from having a global `globalSetup` and `globalTeardown` calling the puppeteer ones,
instead it moves them to the `e2e` project, since jest 24 supports them.

Instead requiring the user to run `npm start` at `globalSetup`, the proxy, on error, will tell that to the user.

Closes #1134

### 🗺️ Test Plan
...


packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -34,114 +34,111 @@ const serverLogPrefixer = () => {
...
 
-    const webpackDevServerProcess = await getProcessOnPort(
-      servers.cdn.port,
-      false,
+  const webpackDevServerProcess = await getProcessOnPort(
+    servers.cdn.port,
+    false,
+  );
+
+  if (!webpackDevServerProcess) {
```
ronami: I think that this check will throw if the user didn't run `npx yoshi start` in another terminal.

packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -34,114 +34,111 @@ const serverLogPrefixer = () => {
...
+  }
 
-    if (!webpackDevServerProcess) {
-      throw new Error(
-        `Running E2E tests requires a server to serve static files. Could not find any dev server on port ${chalk.cyan(
-          servers.cdn.port,
-        )}. Please run 'npm start' from a different terminal window.`,
-      );
-    }
+  if (webpackDevServerProcess.cwd !== process.cwd()) {
```
ronami: I don't think we need this check if we also remove the previous one.

packages/yoshi-helpers/rewrite-forward-proxy.js
```diff
@@ -23,13 +23,20 @@ module.exports = async function startRewriteForwardProxy({
...
+      const originalTarget = target;
+
       if (target.startsWith(search)) {
         target = target.replace(search, rewrite);
       }
 
       regularProxy.web(req, res, { target }, err => {
         if (err) {
           res.statusCode = 500;
+          res.write(
```
ronami: This is going to show this error for every failed HTTP request. Even for HTTP requests that didn't get to `webpack-dev-server` or that didn't fail with 404.

I think that by checking that the URL the request was made to is our `webpack-dev-server` URL and that the request failed with `404`, only then print this helper error message.

Wdyt?

packages/jest-environment-yoshi-puppeteer/globalSetup.js
yurynix: You're right, thanks! 👍

packages/yoshi-helpers/rewrite-forward-proxy.js
yurynix: I've added a check to send out the message only if URL was indeed replaced.
This error handler is called only if the request is failing on the socket level, bad response code like `4xx` and `5xx` is considered not an error.

So I've added a code to handle that as well:
```
regularProxy.on('proxyRes', function(proxyRes, req, res) {...}
```

Can test with:
```es6
const startRewriteForwardProxy = require('./rewrite-forward-proxy');
async function start(port) {
  closeProxy = await startRewriteForwardProxy({
    search: 'https://static.parastorage.com/test',
    rewrite: 'http://localhost:3200',
    port,
  });
};

start(12345);
```

We'll get a connection refused error:

```shell
https_proxy=http://localhost:12345 curl -v --insecure https://static.parastorage.com/test/bla
```
-->
```
Proxy failed to get http://localhost:3200/bla: connect ECONNREFUSED 127.0.0.1:3200
You need to start Yoshi CDN to serve https://static.parastorage.com/test/bla?
```

To test the 404 scenario, we can run just some http server to return 404:
```shell
npx http-server -p 3200 .   
```

packages/yoshi-helpers/rewrite-forward-proxy.js
```diff
@@ -19,17 +19,45 @@ module.exports = async function startRewriteForwardProxy({
     cert: fs.readFileSync(path.join(__dirname, './server.cert')),
   };
 
+  // If we get a bad response, status codes 4xx or 5xx, something's wrong with CDN
+  // we should inform the user about it.
+  regularProxy.on('proxyRes', function(proxyRes, req, res) {
```
ronami: Why do we need this if we already add better errors in `proxyRequest `?

yurynix: It's not better, it's different, the error handle in `proxyRequest` https://github.com/wix/yoshi/pull/1176/files/bc2a5b4037ab5ea4c8224936cf9c9c292ae59ba4#diff-5160cf5881b0aeff8a0b990e1aa25849R50
Is called only on network level errors, like connection refused.
404 is not considered an error by it, since it's perfectly valid from a remote server to return 4xx/5xx, the proxy just needs to pass it trough.

PR #1187, opened 2019-03-29
feat: optional open browser

bildja:
### 🔦 Summary
Every time I run `npm start` the browser gets open in localhost. For some of our apps it does not make sense, since it requires different query params for developers and use-cases.
So as a solution there is a flag `openBrowser` which is on by default, but lets turn it off.

### 🗺️ Test Plan
...

### DIFF
--- a/docs/api/configuration.md
+++ b/docs/api/configuration.md
@@ -213,6 +213,12 @@ Set to `false` in order to disable hot module replacement. (defaults to true)
 
 If true, instructs the browser to physically refresh the entire page if / when webpack indicates that a hot patch cannot be applied and a full refresh is needed.
 
+### openBrowser
+
+`Boolean`
+
+Set to `false` if you don't want localhost to be opened after npm start. (defaults to true).
+
 ## performance
 
 Allows to use the Webpack Performance Budget feature.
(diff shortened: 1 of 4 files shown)


bildja:
@ranyitz as I said, the URL is not constant, it requires query params like fileId and siteToken, which are unique for every dev and for usecase, so passing one url as a parameter to package.json wouldn't work. Besides devs might want to work with different fileId for different cases and site-token tends to expire, so it will need to get changed.
Besides that it's not only my feedback, that it's just annoying behaviour :)

bildja:
@ranyitz developers say they just don't need this, because we usually have the tab already open for it.
One more use-case is for app-settings, tpa widget – where opening the tab is just not making sense, because it can only work in an iframe with editor/viewer environment and depends on postmessages it gets from environment.
Passing `BROWSER=none` seems to work, thanks. Explicit param would still be a cleaner option as for me, but it's up to you to decide, whether to just close this PR.

ranyitz:
@bildja That makes sense, Currently we have this PR as a reference, if we'll see that more people need this feature, we'll add it as part of the CLI.

Thanks for understanding.

PR #1190, opened 2019-04-01
Add ability to use an answers json file instead of prompt

Schniz:
### 🔦 Summary

In order to make testing of Perfer's Yoshi preset synced to every Yoshi change, I'd like to create a `create-yoshi-app` on the fly.
This PR adds an ability to provide `--answers-file` to `create-yoshi-app` so I could autogenerate fullstack projects on tests, instead of
having a generated (and unsynced) yoshi project on the repo.

### 🗺️ Test Plan

I have added a test that saves the "minimal-template" to disk, and then read it.
Also, I have extracted the `defineProperty` part of the tests to the `Answers` class.

cc @ranyitz


packages/create-yoshi-app/src/Answers.js
```diff
@@ -27,4 +27,26 @@ module.exports = class Answers {
 
     return this.projectType + typescriptSuffix;
   }
+
+  toJSON() {
```
Schniz: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior

@cowchimp

packages/create-yoshi-app/__tests__/createApp.spec.js
```diff
@@ -54,19 +54,30 @@ test('it should not create a git repo if the target directory is contained in a
   expect(() => fs.statSync(path.join(projectDir, '.git'))).toThrow();
 });
 
+test('it uses answers from a file', async () => {
+  const answersFile = tempy.file();
+  fs.outputFileSync(answersFile, JSON.stringify(minimalTemplateAnswers()));
```
ranyitz: I would use a fixture here (instead of generating it on the fly)

packages/create-yoshi-app/bin/create-yoshi-app.js
```diff
@@ -27,4 +29,4 @@ if (customProjectDir) {
   process.chdir(path.resolve(customProjectDir));
 }
 
-createApp(process.cwd(), customProjectDir);
+createApp(process.cwd(), customProjectDir, answersFile);
```
ranyitz: It makes sense to use an object for named properties sake, and because `customProjectDir` can be `undefined`.

Issue #1174, opened 2019-03-24
Consider upgrading JSDOM to v14


ronami:
### Summary

This is targeted specifically to Jest but can also be relevant for our Mocha/Jasmine setup.

We use the default Jest environment for JSDOM which use an old version of JSDOM (`^11`) and probably won't upgrade soon because JSDOM dropped support for Node 6 (https://github.com/facebook/jest/pull/7122#issuecomment-427893092). Those recent versions of JSDOM added many helpful features (__even shadow DOM support and `MutationObserver` API__).

I think we should bump JSDOM to version 14 (https://github.com/jsdom/jsdom/blob/master/Changelog.md) and I don't think that this version bump is going to break any of our users (needs to be checked).

To support in it our Jest preset, we can do something similar to https://github.com/ianschmitz/jest-environment-jsdom-fourteen (used in CRA).

ronami:
Closed in #1193. Thanks @nktssh and @ihork 🏆

PR #1193, opened 2019-04-01
Upgrade JSDOM to v14

nktssh:
### 🔦 Summary
JSDom dependency was updated according to #1174 

### 🗺️ Test Plan
**TL;DR version**: We checked, there is v14! 😎

**Long version**: we checked it by creating additional test on our local env with one of the features were added to JSDOM since v11 (v13 actually, function `insertAdjacentElement ` is present as specified in changelog)


jest.unit.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
   testMatch: ['**/__tests__/**/*.spec.js'],
   testResultsProcessor: 'jest-teamcity-reporter',
+  testEnvironment: 'jest-environment-jsdom-fourteen',
```
ronami: I think this is not needed. We can probably even change this to `node`. This is the setup for Yoshi's unit tests (no need for `jsdom`).

package.json
```diff
@@ -58,6 +58,7 @@
     "globby": "^9.0.0",
     "husky": "^1.0.0-rc.9",
     "jest": "^24.1.0",
+    "jest-environment-jsdom-fourteen": "^0.1.0",
```
ronami: Can probably be removed, see above.

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
     ...[
       {
         displayName: 'spec',
-        testEnvironment: 'jsdom',
+        testEnvironment: 'jest-environment-jsdom-fourteen',
```
ronami: I think we should use `require.resolve()` just to make sure we don't get a different version of that package.

This is a breaking change and I want to test it in version 4 **before** we release the v5. What do you think about adding this behind a configuration option? We'll use `jest-environment-jsdom-fourteen` only if it's turned on (opt-in).

This way, we can merge and release this feature under v4. Projects can opt-into it by turning this config on. In v5 we can remove the opt-in and use this for everyone (breaking). 

cc @yavorsky @ranyitz

nktssh: @ronami we made environment variable for this

nktssh:
@ronami is it good enough to merge?

Issue #1121, opened 2019-03-17
Add versioned documentation


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

Currently, we have multiple major versions. The docs are different regarding each feature. It can confuse users that use older versions.

### 🤔 Expected Behavior

The user can choose the version of the documentation (according to his version of Yoshi)

### 🔦 Context

I've heard about users that wasted time due to reading version 4 documentation while they have used version 3.

### 💁 Possible Solution 

Docusaurus has versions support, please see https://github.com/facebook/jest/pull/5541 for reference

ranyitz:
Closed on #1194

PR #1194, opened 2019-04-01
Add versioned documentation 🎉🐉

nktssh:
### 🔦 Summary
Refers to #1121 
Only thing missing is links to the change log because of version URL format which include dates, e.g. https://github.com/wix/yoshi/blob/master/CHANGELOG.md#411-2019-03-31. 
So for now it just redirects you to the CHANGELOG page.

versions page
current stable version
master version
one of the previous versions

### 🗺️ Test Plan
- `cd website && npm start`
- play with versions menu


yavorsky:
@ronami for Babel we are using https://babeljs.io/versions minor updates also.
But in general I think 99% of users are using `^` version range, so just updating docs for the current minor version should work for us.

yavorsky:
But looks really amazing! Thanks guys!

ronami:
This is really good.

I think we should fold the docs for version 4 to `4.x.x` or `4.1.x` too. I'm not sure if we need to create new docs for every patch version.

nktssh:
@ronami it could be confusing for users - which version is actual. Besides, Jest is using same approach - fold all previous minor versions into one but for current major they do exists.
E.g. if current version is 24, there are docs 24.2, 24.1, 24.0, 23.x, 22.x
So I think we nned to keep minor version al least, not sure about patch version though.

ranyitz:
@nktssh Thanks for your contribution, this is really awesome!

I agree that it's more correct to keep all versions, but we release patches/minors on a daily basis, we got to `yoshi@3.31.1` it means 31 versions of the docs.

We do not encourage people to stay on old patches/minors but to upgrade to the last minor possible, there should be no reason to stay on an old minor.

It's a tradeoff, the price of holding all of this history in the git repo is high, it's hard to search and replace things that are relevant. The gain is smaller, it means to support old minor versions.

I agree with @ronami that we should start with major versions only and if we'll feel the need we would be able to always start keeping a documentation version for each minor.

nktssh:
@ranyitz agree.
So I'll create `3.x` and `4.x` doc versions.
But in this case keep in mind that we will need to regenerate docs on each minor release under same name `4.x` (if there are any changes)

PR #1197, opened 2019-04-02
Simplify biz-mgr babel template

yurynix:
### 🔦 Summary
Simplify biz manager templates according to:
https://github.com/wix-private/shahata-generated/commit/42b36b756a72a4889791b79b5abdb3baaccabbb0

### 🗺️ Test Plan
- run `npm run create-yoshi-app:dev`
- Choose biz-mgr babel template
- assert created project's tests are passing `npm test`, and you are able to start the project `npm start` (you'll need `git init` for it)


packages/create-yoshi-app/templates/business-manager-module/typescript/src/config.ts
```diff
@@ -1,20 +1,17 @@
...
-// Note: should be in sync with the config inside your BM module ERB
-export const MODULE_ID: any = '{%PROJECT_NAME%}';
-export const LAZY_COMPONENT_ID: any = '{%projectName%}-lazy-component-id';
+import { ModuleId, PageComponentId, TModuleParams } from '@wix/business-manager-api';
 
-export const COMPONENT_ID: any = '{%projectName%}-component-id';
-export const BI_VIEW_ID = '{%projectName%}_APP_VIEW_ID';
+// Note: after submitting PR to BM you can remove the constant strings below
+// Note: should be in sync with the config inside your BM module ERB
+export const MODULE_ID = ModuleId.Realestate || '{%PROJECT_NAME%}';
```
yairhaimo: Why `Realestate`?
If i understand, the user can use an enum once BizMgr merge the PR?
I think using a fake enum is confusing.

yurynix: Good catch, a Copy&Paste mistake. I'll change it.

packages/create-yoshi-app/src/runPrompt.js
```diff
@@ -21,7 +21,8 @@ module.exports = async (workingDir = process.cwd()) => {
   }
 
   // use the basename of the current working directory if projectName wasn't supplied
-  answers.projectName = answers.projectName || path.basename(workingDir);
+  answers.projectName =
+    answers.projectName || 'tmp' + path.basename(workingDir);
```
yairhaimo: what is the purpose of `tmp`?

ranyitz: I think that it breaks the behavior of running `npx create-yoshi-app` inside of an empty directory

.eslintignore
```diff
@@ -3,4 +3,6 @@ node_modules
 dist
 packages/eslint-config-yoshi-base/__tests__/rules
 test/projects/*/src/errors
-packages/create-yoshi-app/templates/business-manager-module/test/e2e/drivers/app.driver.js
+packages/create-yoshi-app/templates/business-manager-module/javascript/__tests__/e2e/drivers/app.driver.js
+packages/create-yoshi-app/templates/business-manager-module/javascript/src/config.js
+packages/create-yoshi-app/templates/business-manager-module/typescript/src/config.ts
```
ranyitz: @yurynix Can you explain why we should ignore those files during the lint phase?

packages/create-yoshi-app/src/createApp.js
```diff
@@ -33,7 +33,7 @@ module.exports = async ({ workingDir, projectDirName, answersFile }) => {
 
   console.log(
     `\nCreating a new ${chalk.cyan(
-      results.projectType,
+      results.getTitle(),
```
ranyitz: Nice catch! 👍 Thanks for the fix

packages/create-yoshi-app/src/getValuesMap.js
```diff
@@ -13,6 +14,9 @@ module.exports = ({ projectName, authorName, authorEmail, organization }) => {
   for (const key in valuesMap) {
     // create CONSTANT_CASE entries for values map
     valuesMap[constantCase(key)] = constantCase(valuesMap[key]);
+
+    // create PascalCase entries for values map
+    valuesMap[pascalCase(key)] = pascalCase(valuesMap[key]);
```
ranyitz: Interesting feature! 👍

.eslintignore
yurynix: Because:
```
export const MODULE_ID = ModuleId.{%ProjectName%} || '{%PROJECT_NAME%}';
export const COMPONENT_NAME = PageComponentId.{%ProjectName%} || '{%projectName%}';
```

Is not a valid code until the project is generated, so the linter always complain here. On the other hand, I don't want to insert comment to ignore the line since after the template runs it'll be fine.

packages/create-yoshi-app/src/getValuesMap.js
yurynix: That's for:
```
export const MODULE_ID = ModuleId.{%ProjectName%} || '{%PROJECT_NAME%}';
```

packages/create-yoshi-app/src/runPrompt.js
yurynix: Good catch 👍 I'll address that differently

yairhaimo:
Removed the `@wix/react-test-wrappers` fix because BizMgr will fix it on their end.

PR #1196, opened 2019-04-02
Refactor create yoshi app

ranyitz:
### 🔦 Summary
This is a part of #1192 which I want to merge in order to prevent conflicts in the future.

* Refactor to use typescript & javascript directories instead of -typescript - While we don't want to support a template that provides only javascript/typescript, our current design enables it (We list them separately and only connect them during the questions phase). Also, when providing the extend API, it becomes cumbersome to require template developers to connect the template in two places and we want to enforce them to use the same base template for both typescript and javascript.

* ~~Refactor the e2e tests to _not publish the monorepo locally_ and instead link the node modules as we do in the app flow tests, which is much faster and much less flaky locally.~~ Instead, just refactor the symlinks mechanism on app flow

### 🗺️ Test Plan
1. All tests for CI are passing.
2. app flow tests are passing locally.
3. `npm run create-yoshi-app:dev` is working locally.

____________

app flow tests passing locally

image

____________

dev experience is working locally

image

PR #1199, opened 2019-04-02
Support extending templates

ranyitz:
### 🗒 TL;DR
The purpose of this PR is to support extending templates, meaning that if we or someone else in the community would like to extend one of our templates they will be able to do so easily. We want `create-yoshi-app` to scale while maintaining its velocity.

#### Code duplication between templates

We're starting to see emerging platforms that would like to provide a template for their users (plugin developers).
Currently, it requires the template creator to copy one of the official yoshi templates (`client`/`fullstack`) and make some changes. The hard part is to continue and keep all of those templates up to date. Every feature would need to be implemented in 10+ templates, which will essentially affect the **velocity** of adding new features to the templates.

#### Misplaced ownership

As a platform creator, I would like the template to be my bread an butter. Be able to make changes without asking permission and the first to know when it fails in the CI (because i've changed something in the APIs). A good way to give them those privileges is by transferring the "platformize" part of the code to their hands. Currently, it's all in Yoshi repository.

### 🔜 The solution

well, in progress...

_______________

### 🔦 Changes Summary
* [x] **Refactor to use `typescript` & `javascript` directories instead of `-typescript`** - While we don't want to support a template that provides only javascript/typescript, our current design enables it (We list them separately and only connect them during the questions phase). Also, when providing the extend API, it becomes cumbersome to require template developers to connect the template in two places and we want to enforce them to use the same base template for both typescript and javascript. (done via #1196)

* [x] **Template Definition will be specified on each template config file** Instead of templates being specified on one file `templates.js` for example, they are now being defined in their own directories, the intention here is that every template could specify if it extends any other template (something which is tightly coupled to the template itself and therefore should not be specified somewhere else)

* [x] **An extended template will override anything that has been already generated on the base template** this is the most basic thing we could do here. later on, we'll may get into merges between `json` files.


ranyitz:
Closing, let's reopen if we'll need this in the future.

Issue #800, opened 2018-11-18
Build and deploy the website on PRs


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior

When someone is doing a PR to the website/docs, there is no live example to see.

### 🤔 Expected Behavior

The website can be previewed from the PR. (A comment with a link?)

### 💁 Possible Solution

Use something like https://surge.sh/ to deploy a temporary site with the changes from the PR and post a link as a comment.

### 🔦 Context

It's easier and faster to review PRs that way, which should speed up the work on the documentation.

### ⚙️ Nice to have

It would be very nice if it would run only if there was a change to the website.

(We can check if there was a change in the website/docs directories using GIT and run the publish + comment only then)

ranyitz:
I've seen https://github.com/wix/surge-github-autorelease around, @odedcagan @shilomangam perhaps you can help?

shilomangam:
I can assist with this, will try to do it till the weekend.

ranyitz:
@shilomangam Awesome, thanks!

saarkuriel:
Work done so far:
PR #950 has the basic setup for this. It deploys the website to surge.sh for every PR.
What's left:
* Notifying in the PR that the deployment has succeeded with a link to surge.sh
* Running this job only when docs were changed (check the diff in the commit and if docs were changed - deploy the site)

@ranyitz @ronami

nktssh:
For notifying I think one of suitable approaches is to create Github app which will be listen to webhook: 
> **Check runs**
> Check run is created, requested, rerequested, or completed.

and if Travis check is completed and there are changes in `docs/` or `website/` folder - post a comment with link. Link should be created manually in the same way Travis does it.

cc @ranyitz @ronami

ranyitz:
@nktssh Thanks! I think that once we do that, we may consider adding the surge release process to the same place, so after it finishes the release it will notify the user using a comment/status check.

vozhekhovskyi:
Hi @ranyitz ,
I and @brumeregan would like to help with this task (I mean `Notifying in the PR...`), but seems we need permissions to change repo settings.
So is it possible to provide us with appropriate permissions?

ranyitz:
Hey @vozhekhovskyi That's great that you are taking this!

If it's ok, I prefer that you'll specify what needs to be changed and someone from the Yoshi team will make those changes (@yavorsky can you help?)

yavorsky:
@ranyitz I can, but I don't have permissions also.

PR #1200, opened 2019-04-02
Make deploy "website" to surge.sh conditional

nktssh:
### 🔦 Summary
Building "website" is now conditional, so Travis CI check would be faster! 🚀🚀🚀 
Refs #800 

### 🗺️ Test Plan
In PR checks, verify that "website" build deployed in this PR (since we changed spelling mistake in website)


nktssh:
I’ve tried to use my personal token but Github automatically revokes it 🙈

We need to create repository variable with name `GITHUB_TOKEN` and value which will be Github API token with access to scope `repo/public_repo` and `write:discussion`.

deploy_website_to_surge.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+if [type = pull_request && ${WEBSITE_CHANGED}]
+then
+  npx surge --project ./build --domain wix-yoshi-${TRAVIS_PULL_REQUEST}.surge.sh;
+
+  #inspired by https://damien.pobel.fr/post/github-api-from-travisci/
+  CURL_RESULT = curl -H "Authorization: token ${GITHUB_TOKEN}" -X POST \
```
ranyitz: About the message, maybe we can go with something similar to https://github.com/reactjs/reactjs.org/pull/1892#issuecomment-479347823

deploy_website_to_surge.sh
```diff
@@ -0,0 +1,11 @@
...
+    -d "{\"body\": \"You could check new website at http://wix-yoshi-${TRAVIS_PULL_REQUEST}.surge.sh/yoshi 😉\"}" \
```
ranyitz:
```suggestion
    -d "{\"body\": \"You could check new website at https://wix-yoshi-${TRAVIS_PULL_REQUEST}.surge.sh/yoshi\"}" \
```

Also I think that you can use html tags to style it a bit 😉

.travis.yml
```diff
@@ -16,6 +16,8 @@ branches:
     - /^version_\d+\.x$/
 cache:
   yarn: true
+env:
+  - WEBSITE_CHANGED = git diff --name-only $TRAVIS_COMMIT_RANGE | grep "website/"
```
ranyitz: Also changes in `/docs` directory are relevant

.travis.yml
```diff
@@ -16,6 +16,8 @@ branches:
...
 jobs:
   include:
     - name: "website"
```
ranyitz:
```suggestion
    - name: "website:publish"
```

.travis.yml
```diff
@@ -25,8 +27,8 @@ jobs:
       - git config --global user.email "${GH_EMAIL}"
       - echo "machine github.com login ${GH_NAME} password ${GH_TOKEN}" > ~/.netrc
       - cd website && yarn install && GIT_USER="${GH_NAME}" yarn run publish-gh-pages 
-    - name: "website - release to surge"
-      if: type = pull_request
+    - name: "website - build for release to surge if there are changes"
```
ranyitz:
```suggestion
    - name: "website:preview"
```

ranyitz: I think that's it's not the best place for a description. We could settle on short names

.travis.yml
```diff
@@ -53,5 +55,5 @@ jobs:
       if: type != pull_request
       before_script: npm install
       script: npm run lint 
-after_success: 
-  - npx surge --project ./build --domain wix-yoshi-${TRAVIS_PULL_REQUEST}.surge.sh
\ No newline at end of file
+after_success:
+  - ./deploy_website_to_surge.sh
```
ranyitz: When thinking about it, I'm not sure that we should preview the site just if the build passes, it means that if you've changed some documentation and want to see how it looks, you need to wait approximately 20 minutes with no reason.

@nktssh what do you think?

deploy_website_to_surge.sh
nktssh: Well, it's so opinionated...
Could you please provide some info how you would like to see the message?

.travis.yml
nktssh: Totally agree! It was here so didnt change this part. Moved to jobs

.travis.yml
nktssh: Added

deploy_website_to_surge.sh
nktssh: I've changed the comment text a little bit.
Also, we need to register a new Github user and get access token from this account with scopes `repo/public_repo` and `write:discussion`. Those token should be stored in this repository settings under name `GITHUB_TOKEN` (but we could rename it)

deploy_website_to_surge.sh
ranyitz: That's correct, let's start with the message provided and improve later on down the road if we'll feel that's important enough.

.travis.yml
ranyitz: Thanks a lot

Issue #1180, opened 2019-03-26
yoshi@4 incorrectly interprets the https setting


bt4R9:
# 🐛 Bug Report

### 😯 Current Behavior

It seems like there is a conflict with `https` and `servers.cdn.ssl` settings when it comes to open browser in `projectType=app`.
In my app I start yoshi using this command:
```
yoshi start --server=index-dev.js
```
(which is implicitly has `{ https: false }` setting)
but eventually it automatically opens `https://localhost:3000` which is wrong.
It happens because yoshi rely on this condition
```
const https = cliArgs.https || project.servers.cdn.ssl;
```
which is `true` for me because `project.servers.cdn.ssl` is `true`.

### 🤔 Expected Behavior

I expect that command `yoshi start --server=index-dev.js` should open `localhost` without a secure connection.

### 💁 Possible Solution 

I think we should rely on `cliArgs.https` setting not combination of `cliArgs.https || project.servers.cdn.ssl;` when it comes to open browser.

### 👣 Steps to Reproduce 

```
{
  "scripts": {
    "start": "yoshi start --server=index-dev.js"
  },
  "yoshi": {
    "servers": {
      "cdn": {
        "ssl": true
      }
    },
   }
}
```

### 🌍 Your Environment

```
  System:
    OS: macOS 10.14.1
    CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  Binaries:
    Node: 11.2.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.4.1 - /usr/local/bin/npm
    Watchman: Not Found
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
    Safari: 12.0.1
  npmPackages:
    @storybook/core:  4.0.0-alpha.14
    @storybook/react: 4.0.0-alpha.14 => 4.0.0-alpha.14
    babel-preset-yoshi:  4.1.0-rc.4
    eslint-config-yoshi:  4.1.0-rc.4
    eslint-config-yoshi-base:  4.1.0-rc.4
    jest-environment-yoshi-bootstrap:  4.1.0-rc.4
    jest-environment-yoshi-puppeteer:  4.1.0-rc.4
    jest-yoshi-preset: ^4.1.0-rc.4 => 4.1.0-rc.4
    storybook: Not Found
    tslint-config-yoshi:  4.1.0-rc.4
    tslint-config-yoshi-base:  4.1.0-rc.4
[...]

PR #1202, opened 2019-04-03
Different https settings for app server and cdn

bt4R9:
### 🔦 Summary
This PR changes the behaviour of `https` setting in `yoshi start` command in `projectType=app`.
Now `--https` flag correctly corresponds in yoshi`s server addresses output.
fixes https://github.com/wix/yoshi/issues/1180

### 🗺️ Test Plan
Tests are a bit difficult here, can somebody participate?

### DIFF
--- a/packages/yoshi/src/commands/start-app.js
+++ b/packages/yoshi/src/commands/start-app.js
@@ -45,7 +45,8 @@ const Server = require('../server-process');
 
 const host = '0.0.0.0';
 
-const https = cliArgs.https || project.servers.cdn.ssl;
+const appHttps = cliArgs.https;
+const cdnHttps = project.servers.cdn.ssl || false;
 
 function watchPublicFolder() {
   const watcher = chokidar.watch(PUBLIC_DIR, {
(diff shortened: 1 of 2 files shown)


ronami:
Thanks for the PR.

I know that it's confusing but our CLI flag (`--https`) and our JSON config (`servers.cdn.ssl`) both have the same effect: Start `webpack-dev-server` in HTTPS. We don't support starting an app server in HTTPS.

Changing on of them to only open the browser with HTTPS can break certain projects.

I want to remove the duplicated configuration options in the future, but for now I suggest that we always open the browser in HTTP. Not a lot of projects start their server in HTTPS, and if they do, they can use `--url=...` as a way around it.

In the future, I want to have the option for Yoshi to start your app server in HTTPS (maybe with a proxy?).

Wdyt?

ranyitz:
Agree with @ronami 
We should strive to a more coherent API, but currently, we can't control the user's server, so using the [`--url` option](https://wix.github.io/yoshi/docs/guides/app-flow#npx-yoshi-start) would be the best way to go.

closing

PR #1204, opened 2019-04-03
Run wix-wnp-version, when package.json is changed (in server related templates)

sidoruk-sv:
### 🔦 Summary

Usually after project is created users of yoshi update packages (for example with `npm-update`), after updating versions of some packages are specified in `package.json`, but for WNP modules the version should always be `"latest"`, so this   [wix-wnp-version](https://github.com/wix-platform/wix-node-platform/tree/master/support/wix-wnp-version) helps to keep correct versions of Wix Node Platform packages.

List of packages that will be affected (if ever be added to the project):
https://github.com/wix-platform/wix-node-platform/blob/master/support/wix-wnp-version/lib/modules.json


ranyitz:
@sidoruk-sv Thanks for suggesting this idea.

> CLI tool to verify/fix your wix-node-platform dependencies are on latest

Just to clarify, this tool is relevant only if you're using the node platfrom, so I guess we should leave `client` & `business-manager-module` out of the scope of the PR.

_______________________

Regarding the question of whether we should add this by default to the fullstack template I would like to hear some more opinions @ronami @yairhaimo what do you say?

_______________________

My opinion is that adding boilerplate to the templates is something that we're trying to limit as far as we can to leave only critical things.

I've never heard that putting `node_platform` dependencies with `latest` tag was a big problem that we should address (and worth the added boilerplate and the time of running it before each commit).

sidoruk-sv:
@ranyitz seems like yes, also thought that this is relevant only to server-related projects, but there are some dependencies that can be added to the client project (for example `wix-proto-codegen` to generate types for client application). And when such will be added, this utility starts to play its role.

sidoruk-sv:
Added `server` template && remove client template types from PR

sidoruk-sv:
> I've never heard that putting node_platform dependencies with latest tag was a big problem that we should address (and worth the added boilerplate and the time of running it before each commit).

@ranyitz  I used to see such in node_platform channel:
* https://wix.slack.com/archives/C0C89GRRB/p1555571585054500?thread_ts=1555571399.054000&cid=C0C89GRRB
* https://wix.slack.com/archives/C0C89GRRB/p1554789307167900?thread_ts=1554788973.167600&cid=C0C89GRRB
* https://wix.slack.com/archives/C0C89GRRB/p1555514195047800?thread_ts=1555514131.047600&cid=C0C89GRRB

About some of the solutions:
https://wix.slack.com/archives/C0C89GRRB/p1548852708492100

sidoruk-sv:
@ranyitz @ronami @yairhaimo any updates on this?

ranyitz:
@sidoruk-sv Thanks for the effort.

> My opinion is that adding boilerplate to the templates is something that we're trying to limit as far as we can to leave only critical things.

> I've never heard that putting node_platform dependencies with `latest` tag was a big problem that we should address (and worth the added boilerplate and the time of running it before each commit).

I think it adds boilerplate for every project and the problem it's solving is verifying that you are using `latest` for node platform dependencies (or not changing it from `latest` to something else).

Personally, I would not add this check to my project.

Maybe @yairhaimo or @ronami thinks otherwise.

sidoruk-sv:
Fresh one:
https://wix.slack.com/archives/C0C89GRRB/p1559035771054900?thread_ts=1559029090.044100&cid=C0C89GRRB

sidoruk-sv:
Also, comment from Alexander Vainshtein (from Wix Node Platform team):
> In Node Platform we do not maintain semver. The approach is that all platform’s modules *in the current state of `master` branch` are the *current* logical version of the Node Platform. We don’t support forward-backward compatibility between platform modules from “different logical versions*.
> Thus it is important to always have Node Platform dependencies with the `latest` version in package.json.

ranyitz:
@sidoruk-sv You are not bringing any new arguments to the table.

I think that it's clear to the maintainers of Yoshi that node-platform's dependencies should be with the `latest` tag.

[Here is a proof...](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/fullstack/typescript/%7B%25packagejson%25%7D#L23-L29)

I would not add a tool that **runs on each commit** and verifies that my node platform dependencies are using `latest`, while the equivalent is just to generate them like that.

___________________

If we want to enforce it somehow, we need to find another way.

We can, for example, run it on `postInstall` hook, so it won't be the concern of each project, but a concern of the module itself.

What do you think?

sidoruk-sv:
@ranyitz it wouldn't run on **each commit**, cause it will be run on pre-commit phase only if package.json is changed:
```
lint-staged": {
    "package.json": [
```

ranyitz:
> so the case when you add a new dependency from Wix Node Platform will also be handled correctly.

@sidoruk-sv yes, and also any other case when you change the version of any dependency or any other unrelated configuration in `package.json`.

Please note that while this is a disadvantage, also having this concern as a few lines in each project is a disadvantage.

Here I've offered a solution to this problem which **doesn't include those disadvantages**.

Run the check during the `postInstall` hook of the module itself.

Any reason not to do it?

sidoruk-sv:
> Here I've offered a solution to this problem which **doesn't include those disadvantages**.
> 
> Run the check during the `postInstall` hook of the module itself.
> 
> Any reason not to do it?

Agreed, moved to `postinstall` hook

ranyitz:
> Agreed, moved to postinstall hook

I actually meant to the post install hook of **one of the node platform's modules**.

> Run the check during the `postInstall` hook of the module itself.

I think that it's the node platform's concern and by changing it there, they will solve any future problems regarding versioning (regardless of what the developer do in its `package.json`).

@hugebdu Is there any plan on integrating this check as part of `wix-bootstrap-ng` post install hook?

hugebdu:
@ranyitz @sidoruk-sv `postinstall` indeed sounds like the proper place.
as for `wix-bootstrap-ng` that's not enough, since this one is the first one to be added (either by `yoshi` generated or manually). I guess we have to do something smarter about it.

sidoruk-sv:
Decided to not do this on yoshi side

ranyitz:
@hugebdu The fact that it's the first one should not be a problem.

In CI there are installs for each change, it means that it will be validated on CI and in case someone will try to push a node platform's module with a version that is different than `latest` their build will fail with a proper error.

hugebdu:
@ranyitz so your idea is to run validation on `postinstall`, my idea was actually to run the fix (update `package.json` on local (not CI) `postinstall`.

of course we can start only with the validation

ranyitz:
@hugebdu I think that validation, in this case, would be more predictable. The user should understand what they did wrong from the error message and what to do in order to fix it.

You can also add a link to GitHub that explains properly why you want them to work with `latest`.

If we'll see that fixing this kind of problems takes a lot of time, we can think of a different mechanism to auto fix it for them, but IMHO it won't be an issue.

hugebdu:
Fixing is easy - simply run our CLI tool. Always the same. IMHO mechanical, stupid and annoying.
But again, let's start with the validation

PR #1206, opened 2019-04-03
Pipe output to server.log in app-flow

### LINKED ISSUE #1107: Write server output to `target/server.log` in app flow
### Summary

Currently, Yoshi doesn't show the output of the server in the terminal and instead pipes it to a file: `target/server.log`. In app-flow, we changed it so it's shown in the terminal. Users have asked to also have it piped to a file for convenience as we clear the screen pretty often in app-flow.

ihork:
### 🔦 Summary
Pipe server output to server.log in app-flow.
Closes #1107 

### 🗺️ Test Plan
Tested manually on test project

### DIFF
--- a/packages/yoshi/src/commands/start-app.js
+++ b/packages/yoshi/src/commands/start-app.js
@@ -181,7 +181,7 @@ module.exports = async () => {
   try {
     await serverProcess.initialize();
     const serverLogStream = fs.createWriteStream(
-      path.join(process.cwd(), 'target', 'server.log'),
+      path.join(TARGET_DIR, 'server.log'),
     );
     [serverProcess.child.stderr, serverProcess.child.stdout].forEach(stream =>
       stream.pipe(serverLogStream),
--- a/packages/yoshi/src/commands/start-app.js
+++ b/packages/yoshi/src/commands/start-app.js
@@ -180,12 +180,6 @@ module.exports = async () => {
 
   try {
     await serverProcess.initialize();
-    const serverLogStream = fs.createWriteStream(
-      path.join(TARGET_DIR, 'server.log'),
-    );
-    [serverProcess.child.stderr, serverProcess.child.stdout].forEach(stream =>
-      stream.pipe(serverLogStream),
-    );
   } catch (error) {
     console.log();
     console.log(
--- a/packages/yoshi/src/commands/start-app.js
+++ b/packages/yoshi/src/commands/start-app.js
@@ -180,6 +180,12 @@ module.exports = async () => {
 
   try {
     await serverProcess.initialize();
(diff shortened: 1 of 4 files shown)


packages/yoshi/src/commands/start-app.js
```diff
@@ -180,6 +181,9 @@ module.exports = async () => {
 
   try {
     await serverProcess.initialize();
+    [serverProcess.child.stderr, serverProcess.child.stdout].forEach(stream =>
+      stream.on('data', writeToServerLog),
```
yurynix: Why not just pipe to the file stream?
```es6
const serverLogStream = fs.createWriteStream(path.join(base, 'target', 'server.log'));
// ...
[serverProcess.child.stderr, serverProcess.child.stdout].forEach(stream => stream.pipe(serverLogStream));
```

ihork: I used existed util. But you're right - pipe is better in this case

packages/yoshi/src/commands/start-app.js
```diff
@@ -180,6 +180,12 @@ module.exports = async () => {
 
   try {
     await serverProcess.initialize();
+    const serverLogStream = fs.createWriteStream(
+      path.join(process.cwd(), 'target', 'server.log'),
```
yurynix: Since we have `TARGET_DIR` above, maybe we can simplify to `path.join(TARGET_DIR, 'server.log')` ?

ihork: Great idea, thanks

packages/yoshi/src/commands/start-app.js
```diff
@@ -180,6 +180,12 @@ module.exports = async () => {
 
   try {
     await serverProcess.initialize();
+    const serverLogStream = fs.createWriteStream(
```
ronami: Can we move this over to [server-process.js](https://github.com/wix/yoshi/blob/master/packages/yoshi/src/server-process.js#L28)?

packages/yoshi/src/commands/start-app.js
```diff
@@ -180,6 +180,12 @@ module.exports = async () => {
...
+      path.join(TARGET_DIR, 'server.log'),
```
ronami: I think it would be nice to put the path to `server.log` in [https://github.com/wix/yoshi/blob/master/packages/yoshi-config/paths.js](paths.js). We can then refactor [app-server](https://github.com/wix/yoshi/blob/master/packages/yoshi/src/tasks/app-server/index.js#L141) to use it too.

yurynix:
@ronami could you advice on how best to create a test for that?

PR #1207, opened 2019-04-03
[yoshi-config] update unitTests glob pattern

skyiea:
We need to extend `unitTests` glob to support next folder structure:
```
- components/
-- Loader
--- __tests__
---- spec.ts <-- is not matched by current glob pattern
---- driver.tsx
--- index.ts
--- view.tsx
```

If good to go:
- [ ] update `wallaby.js` glob pattern as well
- [ ] document


skyiea:
The use-case, as I've mentioned, is the approach of structuring a project where module name is used to name module directory, while internals naming is based on the purpose of the file, without duplicating the module name itself (e.g. `spec.ts`, not `moduleName.spec.ts`). I've shared the example module structure in issue description. This approach is employed in one of our projects: https://github.com/wix-private/cashier-client/tree/master/wallet/wallet
Following existing glob pattern feels hacky in scope of such project structuring.

PR #1208, opened 2019-04-03
handle bad MATCH_ENV

orimwix:
### 🔦 Summary
background:
In Yoshi 4, there is no longer an env called `component`, yet projects migrating from Yoshi 3 might not know this, and set `MATCH_ENV=component`. 
If they do, jest is run with no configuration at all, and get a long and meaningless list of errors.

solution:
1) If no env is matched
- configure Jest to not run any test.
- Print a meaningful error to the console
2) Add a note to the migration docs

### DIFF
--- a/docs/migration/version-4.md
+++ b/docs/migration/version-4.md
@@ -141,7 +141,7 @@ To make migration easier, please start by running a helper script, which will ch
 curl https://gist.githubusercontent.com/ronami/1608dc49efc166bb6e15a21f7073cb79/raw | node
 ```
 
-note: `MATCH_ENV=component` should be replaced with `MATCH_ENV=spec`
+__Note:__ If you're using `MATCH_ENV=component` it should be replaced with `MATCH_ENV=spec`
 
 ### Puppeteer actions in Jest have a shorter default timeout
 
--- a/docs/migration/version-4.md
+++ b/docs/migration/version-4.md
@@ -141,6 +141,8 @@ To make migration easier, please start by running a helper script, which will ch
 curl https://gist.githubusercontent.com/ronami/1608dc49efc166bb6e15a21f7073cb79/raw | node
 ```
 
+note: `MATCH_ENV=component` should be replaced with `MATCH_ENV=spec`
+
 ### Puppeteer actions in Jest have a shorter default timeout
 
 (**Only relevant for projects using `jest-yoshi-preset`**)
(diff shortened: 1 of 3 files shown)


packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -100,7 +108,6 @@ module.exports = {
             '^.+\\.jsx?$': require.resolve('./transforms/babel'),
             '^.+\\.tsx?$': require.resolve('ts-jest'),
             '\\.st.css?$': require.resolve('@stylable/jest'),
-            '\\.(gql|graphql)$': require.resolve('jest-transform-graphql'),
```
yurynix: Why is that required?

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -115,6 +122,12 @@ module.exports = {
           setupFilesAfterEnv,
         };
       }),
+    // fallback if no env was matched
```
yurynix: Why we need two of those?, looks like a duplicate of the one below:
```
    // workaround for https://github.com/facebook/jest/issues/5866
    {
      displayName: 'dummy',
      testMatch: ['dummy'],
      modulePathIgnorePatterns,
    },
```

yurynix:
Thank you for taking the time to create a PR for that!
I had a few questions that I hope you could answer.

packages/jest-yoshi-preset/jest-preset.js
yurynix: That also fails the `test:app-flow` tests.

orimwix: looks like a keyboard accident, oops

packages/jest-yoshi-preset/jest-preset.js
orimwix: the original workaround was needed because of an issue in Jest that ignores the config if there is only one project.
the original workaround fixed a case where the user chose 1 project, and artificially increased the number to 2. Now I suggest a workaround for a case where the user chose 0 valid projects, therefore need to add 2 fake projects to satisfy Jest

orimwix:
Hi, any reason not to merge this?

docs/migration/version-4.md
```diff
@@ -141,6 +141,8 @@ To make migration easier, please start by running a helper script, which will ch
 curl https://gist.githubusercontent.com/ronami/1608dc49efc166bb6e15a21f7073cb79/raw | node
 ```
 
+note: `MATCH_ENV=component` should be replaced with `MATCH_ENV=spec`
```
ranyitz:
```suggestion
__Note:__ If you're using `MATCH_ENV=component` it should be replaced with `MATCH_ENV=spec`
```

PR #1211, opened 2019-04-04
fixing build race condition

deanshub:
### 🔦 Summary
There's a race condition in the build process between transpiling, copying and bundling.
I separated them to solve it


ranyitz:
@deanshub Can you explain what is the race of conditions?

Doing those tasks one after the other could possibly slow down the build, we need to make sure that there is a reference to the reason.

deanshub:
running `yoshi build` throws 
```dist/src/types/app-types.d.ts:3:28 - error TS2307: Cannot find module '../graphql/queries-schema'.```
the ts file is there (in src) but the transpiled js file will be there only after the transpile phase which runs in parallel to the bundle process, so sometimes it's not there in time

reproduced on https://github.com/wix-private/ecom/tree/yoshi4

**Comment by deanshub** [comment]
oh and it seems like I need to do it on the start command as well, will update the pr

**Comment by deanshub** [comment]
fixed the start command as well

**Comment by ranyitz** [comment]
@deanshub I'm sorry that it took some time to answer.

In client apps, TypeScript transpilation should be derived from a loader, as part of the bundling process (Webpack).

The [transpile javascript task](https://github.com/wix/yoshi/blob/9aa1128123a71c49da695a86cca2337833f95501/packages/yoshi/src/commands/build.js#L241) is in yoshi's codebase for libraries (That produces a transpiled output for other apps to use)

The fact that Yoshi supports both kinds of apps **together** leads to misuse of the original features in a hacky way to support other use cases.

Not that you're doing something improperly, it's just that we could give better support if we'll know the use case, by providing a specific solution.

_____________________________


What I'm trying to say is that the "race of conditions" might not be "the problem" but a suggested solution to support something, which I'm still not sure what it is.

Can you give an exact link to your repo, instead of [link to the whole monorepo](https://github.com/wix-private/ecom/tree/yoshi4) and explain what is the use case?

Or, why do you need to transpile something before you bundle?

**Comment by deanshub** [comment]
reproduce:

within the monorepo (inside the [yoshi4 branch](https://github.com/wix-private/ecom/tree/yoshi4))
```
cd client/wixstores-client/wixstores-client-product-widget
```
and just hit `yoshi build`

the use case:
our package.json has an npm script called `create-graphql-queries-schema` which generates a types file based on graphql schema (located in a different project)
this script should run pre-build and so added to the prebuild script

the issue:
`yoshi build` first removes the transpiled files and it runs the pre-build concurrently with the transpilation which seems to create a race condition since the file generated might not be created in time for yoshi's tranpilation phase

ranyitz:
Thanks @deanshub 

> yoshi build first removes the transpiled files and it runs the pre-build concurrently with the transpilation ...

It cleans the target and dist directories, that's correct.

https://github.com/wix/yoshi/blob/25555486fb37a016d34bed70dde80240c6ed0f4f/packages/yoshi/src/commands/build.js#L52

It [does not run any preBuild step](https://github.com/wix/yoshi/search?q=prebuild&unscoped_q=prebuild) that I know of.

____________________
> ... seems to create a race condition since the file generated might not be created in time for yoshi's tranpilation phase

Do you refer to the file generated during `preBuild` phase?

deanshub:
if it runs the build script, npm automatically runs pre-build script.
yap, I'm refering to the file generated in the pre-build phase.

ranyitz:
@deanshub ok, so if you are right and I understand you correctly, you can just wait for the file to be generated and then run `yoshi build`. You don't need to change anything in the inner implementation if the race of condition is between your custom task and `yoshi build`.

_______________

I do see that you're doing something different from every other project, I would like to understand what it is. Can you explain what [this script](https://github.com/wix-private/ecom/blob/master/client/wixstores-client/wixstores-client-core/scripts/generate-gql-types.js) is doing and why do you run it in a pre build step?

deanshub:
`yoshi build` doesn't run `npm run build` (the build script)?
if so then it needs to support this issue, because `npm run build` runs automatically `npm run prebuild`
<hr/>

our pre build step generates types file using apollo tolling, it takes the gql schema from another directory which is common to all projects and a gql query (which resides within the current project) and generates ts types file

deanshub:
race condition accrued because of a wrong typescript configuration, closing the PR

Issue #1218, opened 2019-04-04
[app flow] ejs fIles are not copied to statics directory in dev mode


maxaon:
# 🐛 Bug Report

### 😯 Current Behavior
If using regular flow all ejs files are copied to dist/statics. 
If **projectType** is **app** then ejs files are not watched and not copied to dist/statics. And app fall on first request.

...

### 🤔 Expected Behavior

During run ejs files should be copied to dist

### 💻 Code Sample

```javascript
   const templatePath = path.join(
      hadron.staticLocalPath(artifactId, './dist/statics'),
      'index.ejs',
    );
    const html = await renderEjsFile(templatePath, renderModel, {
      cache: isProduction,
      filename: templatePath,
    });

```
### 🌍 Your Environment

```
  System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  Binaries:
    Node: 10.15.3 - /usr/bin/node
    Yarn: 1.15.2 - /usr/bin/yarn
    npm: 6.8.0 - ~/.npm-global/bin/npm
    Watchman: Not Found
  Browsers:
    Chrome: 73.0.3683.86
    Firefox: 66.0.2
  npmPackages:
    @storybook/core: Not Found
    @storybook/react: Not Found
    babel-preset-yoshi: Not Found
    eslint-config-yoshi: Not Found
    eslint-config-yoshi-base: Not Found
    jest-environment-yoshi-bootstrap: Not Found
    jest-environment-yoshi-puppeteer: Not Found
    jest-yoshi-preset: Not Found
    storybook: Not Found
    tslint-config-yoshi: Not Found
    tslint-config-yoshi-base: Not Found
    typescript: Not Found
    webpack: Not Found
    yoshi-config: Not Found
    yoshi-helpers: Not Found
    yoshi-runtime: Not Found

 Yoshi config
{
  "projectType": "app",
  "entry": {
    "videoMakerModal": "./videoMakerModal",
    "videoMakerModalCritical": "./videoMakerModal.critical"
  },
  "externals": {
    "react": "React",
    "react-dom": "ReactDOM",
    "react-intl": "ReactIntl",
    "prop-types": "PropTypes",
    "bluebird": "Promise"
  },
  "servers": {
    "cdn": {
      "port": 3220
    }
  }
}
```

ronami:
[...]

PR #1220, opened 2019-04-07
Watch and copy templates in `start-app`

ronami:
### 🔦 Summary

This PR addresses #1218. We already copy template files to `dist/statics` in https://github.com/wix/yoshi/blob/master/packages/yoshi/src/commands/build-app.js#L41.

This PR watches and copies templates in `start-app` to `dist/statics` too.

### DIFF
--- a/packages/yoshi-config/globs.js
+++ b/packages/yoshi-config/globs.js
@@ -24,6 +24,7 @@ module.exports = {
   multipleModules: {
     clientDist: dist,
   },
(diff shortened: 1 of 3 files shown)


yanivefraim:
@ronami - can we merge this?

ranyitz:
There is no need to copy the templates during start, as we do it only in order to upload it to parastorage.

closing

PR #1224, opened 2019-04-07
Create manifest JSON file with mapping to initial assets

ronami:
### 🔦 Summary

This PR adds on `build-app` to create an assets manifest JSON file. The manifest should be used by tools and platforms to get the full mappings of all initial assets, per entry.

Currently, the manifest file will be created only on CI and will include the version in its name (and without any hash). That will allow tools and platforms to get the mappings for a specific version.


packages/yoshi-helpers/utils.js
```diff
@@ -250,7 +256,7 @@ module.exports.killSpawnProcessAndHisChildren = child => {
 
     const pid = child.pid;
 
-    psTree(pid, (err /*eslint handle-callback-err: 0*/, children) => {
+    psTree(pid, (err, children) => {
```
ronami: why did we have this lint override? 😅

ronami: why did we have this lint override? 😅

packages/yoshi/src/commands/build-app.js
```diff
@@ -128,6 +131,44 @@ module.exports = async () => {
 
   const clientOptimizedStats = webpackStats.stats[1];
 
+  // Generate `manifest.[version].json`
+  if (inTeamCity && experimentalBuildHtml) {
+    const assetsJson = clientOptimizedStats.compilation.chunkGroups.reduce(
```
netanelgilad: This piece of code seems a bit complicated to grok. What about some tests? even units could be nice.

packages/yoshi/src/commands/build-app.js
```diff
@@ -128,6 +131,56 @@ module.exports = async () => {
...
+          ...(acc[chunk.name] || []),
+          // Add files to the list
+          ...chunk.chunks.reduce(
+            (files, child) => [
+              ...files,
+              ...child.files
+                // Remove map files
+                .filter(file => !file.endsWith('.map'))
+                // Remove rtl.min.css files
+                .filter(file => !file.endsWith('.rtl.min.css'))
```
netanelgilad: not needed

PR #1235, opened 2019-04-18
Add an option to add a custom protractor config

yanivefraim:
### 🔦 Summary
We want to be able to run Protractor tests in parallel, in ci. Using Protractor's parallel features did not work, because of Applitools eyes issues (see [here](https://github.com/wix/wix-style-react/pull/3265)).

This feature will allow us to run Protractor tests in parallel on multiple machines, each with its own specs glob pattern.


packages/yoshi/config/protractor.conf.js
```diff
@@ -19,7 +19,10 @@ const project = require('yoshi-config');
 
 setupRequireHooks();
 
-const userConfPath = path.resolve('protractor.conf.js');
+const userConfPath = process.env.YOSHI_PC
```
yavorsky: maybe we can use the full name, so it will be more clear for everyone what's going on here.

ranyitz:
@yanivefraim Do we still need this PR?
_____________

I'll say my opinion in case we do want to keep it.

In general, I'm not pro adding a configuration, I can elaborate if needed.

But if it is necessary for someone I would not like to stop them from improving their build times.

I prefer not to add it as an official feature of Yoshi as I think it won't have many usages. Can we add it as an undocumented feature? Maybe an environment param? Something that won't be listed in the docs or when you run `yoshi test --help`.

packages/yoshi/config/protractor.conf.js
ranyitz: I have a totally different connotation to `PC` 😆

yanivefraim:
> I'll say my opinion in case we do want to keep it.
> 
> In general, I'm not pro adding a configuration, I can elaborate if needed.
> 
> But if it is necessary for someone I would not like to stop them from improving their build times.
> 
> I prefer not to add it as an official feature of Yoshi as I think it won't have many usages. Can we add it as an undocumented feature? Maybe an environment param? Something that won't be listed in the docs or when you run `yoshi test --help`.

OK, I changed it to be an env variable, so users cannot see it in `--help`. I also changed the name (:

PR #1242, opened 2019-04-22
Use `jest-yoshi-preset` without Puppeteer (WIP)

yanivefraim:
In case of a project without Puppeteer (server template?), `jest-yoshi-preset` will crash (cannot find Puppeteer). This PR will let you use `jest-yoshi-preset` without the need of Puppeteer

TODO: Add tests?


ranyitz:
@yanivefraim Do we still need this? If not, let's close it and reopen if needed

yanivefraim:
Closing. We decided that this feature might be confusing for client / fullstack projects that for some reason will not have Puppeteer installed (they will break and it will be hard to track for the reason)

Issue #1091, opened 2019-03-10
JS files in TS templates aren't being linted


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

When starting a new project, `yoshi lint` does not lint the `js` files

### 🤔 Expected Behavior

We supply `js` files like `jest-yoshi.config.js` and `index.js`. They should be linted by `yoshi lint`

### 💁 Possible Solution

1. Add `"allowJs": true` to tsconfig.
2. Add those root `js` files to the tsconfig files/include array.

### 👣 Steps to Reproduce

1. Generate a client typescript project
2. Change a JS file
3. Run `npx yoshi lint` or run `npx yoshi lint <js_file_name>`

Both should not raise the relevant lint problem.

### 🎶 Note
Important to say your IDE will also not lint those JS files.

ranyitz:
Also, It would be nice to add `"tslint.jsEnable": true` to the `settings.json` that we supply in the typescript templates (as it won't work otherwise...)

ranyitz:
Fixed on #1243

PR #1243, opened 2019-04-22
Fix issues with lint .js files in typescript environments

dmfilipenko:
### 🔦 Summary
1. Added a list of js files to include a rule to make it visible for a linter. 

https://github.com/wix/yoshi/issues/1091


yurynix:
Thank you for this PR!
Regarding adding:
```
"./src/**/*.js"
```
That sounds too broad to me, should we allow making arbitrary files in a TS project?
I think @ranyitz meant in his issue only include the configuration files that must remain in JS.
what do you think @dmfilipenko ?

ranyitz:
Thanks @yurynix, that's correct. Furthermore, it's important to say that most of those files are not in the `src` directory, and therefore won't be linted. (`index.js`, `jest-yoshi.config.js` etc ...)

packages/create-yoshi-app/templates/fullstack/typescript/.vscode/settings.json
```diff
@@ -35,5 +35,6 @@
   "html.format.extraLiners": "",
   "files.insertFinalNewline": true,
   "eslint.autoFixOnSave": true,
```
ranyitz: I know that it's not related to the PR, but it hurts looking at it.
```suggestion

```

dmfilipenko:
@ranyitz, please review

ranyitz:
@dmfilipenko That's the error from the CI. It's not really emitting files there, but it seems that TypeScript is not allowing it, so it makes sense to just add the `outDir` configuration anyway.

image

Issue #1092, opened 2019-03-10
Running e2e tests with `ssl:true` requires extra configuration


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior
I'm using [`ssl:true`](https://github.com/wix/yoshi/blob/version_4.x/packages/yoshi-config/index.js#L29) in my yoshi config, running `npx yoshi start` and then on another terminal `npx jest e2e` to run my e2e tests.

My tests fail, and when debugging it a bit, I see that the assets that I'm trying to load from `https://localhost:3200` have errors due to an unsigned certificate.

![image](https://user-images.githubusercontent.com/11733036/54088502-45029380-4367-11e9-946a-40cdb92d22a6.png)

I've added `args: ['--ignore-certificate-errors']` to my `jest-yoshi.config.js` file under the `puppeteer` section. And that solved the problem.

### 🤔 Expected Behavior

I won't have these errors in the first place. The preset will take care of them for me.

Because **ZeroConfig™** & **JustWorks™**

### 🔦 Context

I'm writing an app that during development needs to be integrated inside of a platform that runs on `https`. My assets are served on https because of that reason.

### 💁 Possible Solutions

1. Always pass `'--ignore-certificate-errors'` flag. ([In here](https://github.com/wix/yoshi/blob/e093336d475a21f77808521e5b90b5fb5ff3a71a/packages/jest-environment-yoshi-puppeteer/globalSetup.js#L54))
2. Pass the above flag only when using `ssl` in the yoshi config.
3. Run cdn on both `http` and `https, always. If possible it would be really cool.

### 🌍 Your Environment

```
yoshi
└── 4.1.0-alpha.5
```

yurynix:
Closed in #1245

PR #1245, opened 2019-04-22
Make puppeteer ignore ssl errors if user run with flag ssl:true

dmfilipenko:
### 🔦 Summary
Fix for #1092 

### 🗺️ Test Plan
...


packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -52,6 +52,7 @@ module.exports = async () => {
       // defaults
       args: [
         '--no-sandbox',
+        servers.cdn.ssl ? '--ignore-certificate-errors' : '',
```
yurynix: Maybe we should use the same style as below?
```
        ...(servers.cdn.ssl
          ? ['--ignore-certificate-errors']
          : []),
```

yurynix:
Thank you for this PR, I had a small comment, let me know what you think =)

dmfilipenko:
@yurynix I've changed it

PR #1255, opened 2019-04-30
OOI generator best practices

jonathanadler:
### 🔦 Summary
Following todays meeting, I thought there were a few changes that should be made to the OOI generator:
1. The new docs state that using `wix-ui-tpa` is a must, so I thought it would be best to change the `wix-ui-core` button that was used to the `wix-ui-tpa` button (commit [3bc8abf](https://github.com/wix/yoshi/commit/3bc8abfe9ea277b19f808195f4cb60bd12837649)).
2. After adding the `wix-ui-tpa` button, I thought it would be nice to show how the component's styling can be changed from the settings panel (commit [b9f2d76](https://github.com/wix/yoshi/commit/b9f2d7600dec3155b1a72ed21c3f9e78543f72df))
3. It was stated today that `pageReady` returns a promise so maybe it should be `async` by default? Not sure about this... (commit [6031d4e](https://github.com/wix/yoshi/commit/6031d4e00c9abd2f1ffbbfd4e85c03bdd444590f))

### 🗺️ Test Plan
Not yet

### DIFF
--- a/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.js
+++ b/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.js
@@ -19,7 +19,7 @@ export async function createAppController(controllerConfig) {
   const experiments = await getExperimentsByScope(EXPERIMENTS_SCOPE);
 
   return {
-    pageReady() {
+    async pageReady() {
       setProps({
         name: 'World',
         cssBaseUrl: appParams.baseUrls.staticsBaseUrl,
--- a/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/App.js
+++ b/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/App.js
@@ -4,7 +4,7 @@ import {
   ExperimentsProvider,
   withExperiments,
 } from '@wix/wix-experiments-react';
-import { ButtonNext as Button } from 'wix-ui-core/button-next';
+import { Button } from 'wix-ui-tpa/Button';
 import i18n from '../../config/i18n';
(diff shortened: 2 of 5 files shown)


packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.js
```diff
@@ -19,7 +19,7 @@ export async function createAppController(controllerConfig) {
   const experiments = await getExperimentsByScope(EXPERIMENTS_SCOPE);
 
   return {
-    pageReady() {
+    async pageReady() {
```
ranyitz: @gileck What do you think?

Should it return a promise?

gileck: sure, we use `Promise.resolve` on the returned value of pageReady so both promise and primitive value are fine. I assume that you want to make sure the user is aware of the fact that pageReady returns a promise. it makes sense

jonathanadler:
> Looks good, overall.
> 
> The settings panel is getting quite complex.
> I know it's not necessarily related to this PR, but it would be good if it would be possible to generate less code there for the basic behavior.
> 
> @gileck, what do you think?

Regarding the complexity of the settings panel, I totally agree, and I also felt this.

However, I do believe it would be a really great opportunity to show how to extend a `wix-ui-tpa` component.
Maybe instead of changing the button's color, it would be better to change its font size or something like that? and then add a slider instead of the color palette.
Or maybe have the one color palette control both the header and the button.

What do you think?

ranyitz:
@jonathanadler I'm not sure, let's merge it because it's better than what we have now.

We can rethink it later on in improve it along the way

Issue #1173, opened 2019-03-24
Generated project templates should look better


ronami:
### Summary

Currently, projects that are generated from `create-yoshi-app` look old and not very good. It would be nice to make them look polished and more professional.

I did a little bit of trying around in #1027.

PR #1258, opened 2019-04-30
Add `yoshi-template-intro` package to import it for all templates.

yavorsky:
ezgif com-video-to-gif (3)

### 🔦 Summary
This PR updaes fullstack project for CYA.
It contains 2 things that could help to make development fun:
1) Animation with yoshi.
2) Some description with useful links.

I'll update the typescript project if this approach looks good to you.

Fixes https://github.com/wix/yoshi/issues/1173


yavorsky:
> The image is a bit heavy, I would downgrade it a lot.
Good catch! I'll minify it or use svg with different yoshi.

>This change adds around 180 lines of code
It was my concern also, so I moved everything below the comment `You can remove everything below` and grouped it to the component called ComponentYouProbablyWantToRemove.
Maybe moving it to a separate module should be faster... Or maybe we can add a button that will send a message to our local server and call a script that removes everything except root div? Not sure about it also..

What if we'll upload JS and css files with animation and description somewhere and lazy-load it. Then the user just needs to remove 3-4 LOC to make a project clean. We also can reuse this approach for all templates and don't copy paste everything to every project. Should be easier to maintain.

ranyitz:
@yavorsky The idea of having this in one place makes sense for me.

The code reuse part in all templates would be amazing.
The 3 lines of code also sound very appealing to me.

packages/yoshi-template-intro/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "jest-yoshi-preset": "^4.0.0",
+    "lint-staged": "^7.2.2",
+    "node-sass": "^4.12.0",
+    "puppeteer": "^1.15.0",
+    "react-test-renderer": "~15.6.0",
+    "yoshi": "^4.0.0",
+    "yoshi-style-dependencies": "^4.0.0"
+  },
+  "dependencies": {
+    "babel-preset-yoshi": "^4.1.0",
```
ranyitz: Why do we need it?

packages/yoshi-template-intro/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "jest-cli": "^24.8.0",
```
ranyitz: Why do we need it?

Issue #1171, opened 2019-03-24
"yoshi build" breaks after being installed with yarn


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

Generating a new project, removing node modules and using yarn instead resulting in `yoshi build` that raise errors.

![image](https://user-images.githubusercontent.com/11733036/54880585-befa4880-4e4e-11e9-893c-06a6a2187cf7.png)

### 🤔 Expected Behavior

Installing using yarn should work.

> It's important to say that Yoshi is not tested with yarn, but unless stated otherwise, there is no reason for it to not work.

### 📽 Possible problem explanation

After a small research, I've seen that yarn installs `babel-preset-yoshi` inside of `yoshi` directory in `node_modules`.

![image](https://user-images.githubusercontent.com/11733036/54880569-92463100-4e4e-11e9-9ca0-52d6dda9e97a.png)

While the `require.resolve` is actually from the `yoshi-helpers` directory.

https://github.com/wix/yoshi/blob/5612331bed1ef6e6daab1dc5cad37c317991b6f1/packages/yoshi-helpers/utils.js#L37

Means, it will never find it using regular node modules resolution algorithm.

### 💁 Possible Solution
1. `babel-preset-yoshi` should be a dependency of `yoshi-helpers` which requires it.
2. Move the resolution (`require.resolve`) to yoshi.

* I'm sure that there are better solutions to that. I haven't checked all the edge cases of it, it's just a suggested way to start and solve the issue.

### 👣 Steps to Reproduce 

1. Create a new project
2. Remove node_modules
3. Install using yarn

### 🌍 Your Environment

`yoshi@4.1.0-rc.4`

PR #1263, opened 2019-05-01
Fix babel-preset-yoshi resolve path.

yavorsky:
Fixes https://github.com/wix/yoshi/issues/1171
Please check the issue for more info. @ranyitz made incredible research.

### DIFF
--- a/packages/yoshi-helpers/utils.js
+++ b/packages/yoshi-helpers/utils.js
@@ -34,7 +34,7 @@ function logIfAny(log) {
 
 module.exports.createBabelConfig = (presetOptions = {}) => {
   return {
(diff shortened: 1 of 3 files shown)


ranyitz:
Thanks @yavorsky This is really cool!

I think that now we have another solution which I think it worth to reuse in this case, just so we handle the same problems with similar solutions.

The solution is to create a file in the root of the repo that routes the proper package, and then call `require.resolve` to this package as `root/my-package`.

You can look at [this PR](https://github.com/wix/yoshi/pull/1103) for an example to what I've just described.

What do you think?

ranyitz:
image

@yavorsky Are those reverts intentional?

PR #1274, opened 2019-05-06
Add React Hooks eslint rules

### LINKED ISSUE #1265: support linting react hooks
# 🙋 Feature Request

### 😯 Current Behavior

Currently, if you are using react hooks in your codebase, the [rules of hooks](https://reactjs.org/docs/hooks-rules.html) aren't enforced. Which means it's easy to make mistakes.

### 🤔 Expected Behavior

eslint will help us not to make those known mistakes.

### 💁 Possible Solution 

Integrate [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) into eslint-config-yoshi.

### 🔦 Context

@yavorsky already did a PR to integrate it - https://github.com/wix/yoshi/pull/673 It became outdated because of a lot of changes to the codebase.

aarnoldaas:
Fixes #1265
### 🔦 Summary
Currently, if you are using react hooks in your codebase, the rules of hooks aren't enforced. Which means it's easy to make mistakes.
This PR add `eslint-plugin-react-hooks` which enforces not making mistakes when using React Hooks. 
### 🗺️ Test Plan
Rules that were added has unit tests


packages/eslint-config-yoshi-base/package.json
```diff
@@ -24,8 +24,6 @@
     "babel-eslint": "^10.0.0",
```
ranyitz: I think that as we moved all tests into `eslint-config-yoshi` it worth to also move the dev dependencies as well.

ranyitz: My mistake! actually, it's all being used from the root of the monorepo. I think you can remove all of the dev dependencies as they are specified on the root

PR #1278, opened 2019-05-07
ts-lint react-hooks

### LINKED ISSUE #1265: support linting react hooks

aarnoldaas:
Fixes #1265

### 🔦 Summary
Add tslint rule for react-hooks

### 🗺️ Test Plan
Lint rule has unit tests


packages/tslint-config-yoshi/index.js
```diff
@@ -20,5 +21,8 @@ module.exports = addJsRules({
 
     // https://github.com/wix/wix-ui/tree/master/packages/tslint-plugin-wix-style-react
     'no-full-wsr-lib': true,
+
+    //https://github.com/Gelio/tslint-react-hooks
```
ranyitz:
```suggestion
    // https://github.com/Gelio/tslint-react-hooks
```

packages/tslint-config-yoshi/package.json
```diff
@@ -12,12 +12,15 @@
   "license": "ISC",
   "dependencies": {
     "tslint-config-yoshi-base": "4.1.0",
-    "tslint-react": "^3.6.0"
+    "tslint-react": "^3.6.0",
+    "tslint-react-hooks": "^2.1.0"
```
ranyitz: You add `tslint-react-hooks` in both `tslint-config-yoshi` and `yoshi`. There is no need to specify it twice.

We have to options:

1. Add it only on `tslint-config-yoshi` (here) and remove the dependency from `yoshi`, then we'll need to add `tslint-reat-hooks` as a dev dependency of yoshi. So we'll need to maintain those in 2 places.

2. Add it as a peer dependency here, and add it as a dependency of `yoshi`.

I think the second is a bit easier maintenance-wise.

packages/tslint-config-yoshi/package.json
```diff
@@ -12,12 +12,15 @@
...
   },
   "devDependencies": {
+    "glob": "^7.1.3",
```
ranyitz: I think that as we moved the infrastructure of the testing to the root we don't need any dev dependencies here :) 

So feel free to remove all of them

packages/tslint-config-yoshi-base/package.json
```diff
@@ -23,7 +23,6 @@
     "tslint": ">=5.10.0"
   },
   "devDependencies": {
-    "glob": "^7.1.3",
     "tslint": "^5.10.0",
```
ranyitz: Worth removing all the dev dependencies from here as we don't test anything here, and anyways it's from the root of the monorepo :)

aarnoldaas: tslint do not let me remove all deps :/

packages/tslint-config-yoshi/package.json
aarnoldaas: tslint do not let me remove all deps :/

aarnoldaas:
I don't think adding all dependencies to root of the monorepo is good idea. I will need to add all deps: 
tslint, tslint-plugin-wix-style-react, ...

PR #1282, opened 2019-05-08
Log full URL to CDN, including scheme and port

danielagreen:
### 🔦 Summary
Log a full path to the cdn instead of only the port. example:
Running cdn on {port_number} => Running cdn on http://localhost:{port_number}


danielagreen:
@ranyitz on it

danielagreen:
@ranyitz fixed: 
https://github.com/danielagreen/yoshi/pull/1/commits/921b175a3615bba29fb25adc1696fc211426e30e

(not sure if i should create a different pr for this commit)

PR #1285, opened 2019-05-12
Basic monorepo support

ronami:
### 🔦 Summary

This PR adds an experimental config option (`experimentalMonorepo`) to Yoshi that adds basic Lerna monorepo support.

See https://github.com/wix/yoshi/blob/monorepo-support/MONOREPO_SUPPORT.md on the MVP and what it includes.


packages/yoshi/src/commands/release-monorepo.js
```diff
@@ -0,0 +1,54 @@
...
+  const [, libs] = await splitPackagesPromise;
+
+  // Patch libraries' `package.json` main field to point to `dist`
+  await Promise.all(
+    libs.map(async lib => {
+      const packageJsonPath = path.join(lib.location, 'package.json');
+      const json = await fs.readJSON(packageJsonPath);
+
+      // Point to js version of the file in the `dist` directory
+      json.main = path.join('dist', json.main.replace('.ts', '.js'));
```
yanivefraim: Do we want to document this? (maybe not...)
Also, we'll have to remove this and update all projects (breaking change) when we move to use `ts-loader` to transpile project references (https://github.com/TypeStrong/ts-loader/issues/851),

PR #1286, opened 2019-05-13
add wix-ui-tpa provider - for mobile mode

jonathanadler:
### 🔦 Summary
We've added a Provider component in the `wix-ui-tpa` library, which receives a `mobile` boolean parameter. This should be part of the generator I believe.

### DIFF
--- a/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/App.js
+++ b/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/App.js
@@ -4,18 +4,21 @@ import {
   ExperimentsProvider,
   withExperiments,
 } from '@wix/wix-experiments-react';
+import { TPAComponentsProvider } from 'wix-ui-tpa/TPAComponentsConfig';
 import { Button } from 'wix-ui-tpa/Button';
 import i18n from '../../config/i18n';
 import styles from './App.st.css';
 
 export default class AppRoot extends React.Component {
   render() {
-    const { name, locale, experiments } = this.props;
+    const { name, locale, experiments, mobile } = this.props;
 
     return (
       <I18nextProvider i18n={i18n(locale)}>
         <ExperimentsProvider options={{ experiments }}>
-          <App name={name} />
+          <TPAComponentsProvider value={{ mobile }}>
+            <App name={name} />
+          </TPAComponentsProvider>
         </ExperimentsProvider>
       </I18nextProvider>
     );
--- a/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.js
+++ b/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.js
@@ -5,6 +5,10 @@ function getLocale({ wixCodeApi }) {
   return wixCodeApi.window.locale || 'en';
 }
 
+function isMobile({ wixCodeApi }) {
+  return wixCodeApi.window.formFactor === 'Mobile';
+}
+
 async function getExperimentsByScope(scope) {
   const experiments = new Experiments({
     scope,
@@ -16,6 +20,7 @@ async function getExperimentsByScope(scope) {
 export async function createAppController(controllerConfig) {
   const { appParams, setProps } = controllerConfig;
   const locale = getLocale(controllerConfig);
+  const mobile = isMobile(controllerConfig);
   const experiments = await getExperimentsByScope(EXPERIMENTS_SCOPE);
 
   return {
@@ -24,6 +29,7 @@ export async function createAppController(controllerConfig) {
         name: 'World',
         cssBaseUrl: appParams.baseUrls.staticsBaseUrl,
         locale,
+        mobile,
         experiments,
       });
     },
--- a/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.spec.js
+++ b/packages/create-yoshi-app/templates/out-of-iframe/javascript/src/components/app/appController.spec.js
@@ -20,7 +20,9 @@ describe('createAppController', () => {
       },
     };
     const locale = 'locale';
+    const formFactor = 'Desktop';
     const experiments = { someExperiment: 'true' };
+    const mobile = formFactor === 'Mobile';
 
     const controller = await createAppController({
       appParams,
@@ -28,6 +30,7 @@ describe('createAppController', () => {
       wixCodeApi: {
         window: {
           locale,
+          formFactor,
         },
       },
     });
@@ -39,6 +42,7 @@ describe('createAppController', () => {
       cssBaseUrl: appParams.baseUrls.staticsBaseUrl,
       locale,
       experiments,
+      mobile,
     });
   });
 });


ranyitz:
@jonathanadler if that's correct (and everybody needs that provider) maybe we can talk to the platform, bring the component library and the provider once through the platform.

By that, we can benefit from multiple advantages:
1. The `wix-ui-tpa` configuration being in one place (not spread on multiple projects)
2. `wix-ui-tpa` being external, ensures only one version at a time (performance).
3. The conditions like `is-mobile` being calculated once instead of in every projects.
4. Easier to test (no need for mocking providers).
5. Less code in the template lowering the learning curve of the platform.

jonathanadler:
> So if I understand correctly, you have to use this provider to inject the `mobile` boolean to all of `wix-ui-tpa` components?
> 
> @jonathanadler there is no other way to work with the library?

Since the `mobile` property will be needed for many components, and we might also want some more common configuration parameters to be passed to the components (`displayMode` or `rtl` for instance), we believed it would be best to have on provider that will pass all these configuration params to the components, without the need to pass them manually to each component.
At the moment, this is how the library works.

jonathanadler:
> @jonathanadler if that's correct (and everybody needs that provider) maybe we can talk to the platform, bring the component library and the provider once through the platform.
> 
> By that, we can benefit from multiple advantages:
> 
> 1. The `wix-ui-tpa` configuration being in one place (not spread on multiple projects)
> 2. `wix-ui-tpa` being external, ensures only one version at a time (performance).
> 3. The conditions like `is-mobile` being calculated once instead of in every projects.
> 4. Easier to test (no need for mocking providers).
> 5. Less code in the template lowering the learning curve of the platform.

Very very interesting idea.

However, there's one major issue with this:
It will be hard to introduce changes to the provider, and even almost impossible to introduce breaking changes, since we'll have to orchestrate the tpas' and editor's updates to be simultaniously.
This means, for instance, that tpa's won't be able to rollback if they need to (all the tpas and the editor will have to rollback together).

Regarding testing, i'm not really sure I see the difference between the two approaches.

And about the learning curve, i'm not sure there will be a great impact because of this. 
Shlomi suggested that leaving the provider without the `mobile` property will be sufficient.
I think it would be better to have it, since at some point the developers will have to add it anyway.

ranyitz:
@jonathanadler Thanks for the detailed response!

I'll to explain the high-level idea and then go down to the details.
_____________________

`isMobile`, `displayMode` & `rtl` are all specific to a **page**, not to an **app**. 

For example, all of the apps in the page will share the same `isMobile` property and therefore it's an inherent code duplication to run [this function](https://github.com/wix/yoshi/pull/1286/files#diff-3161deb3f14230b7d201388986d0144aR8) on every app, and add [this provider](https://github.com/wix/yoshi/pull/1286/files#diff-2bc694e551703d2f02775938f15ee26fR7) in every app.

I'm sure that the logic will grow bigger and bigger as we'll progress, and action will be required from all of the apps in order to add more properties.

The idea is to add the provider from the page perspective and have the logic in one place. Let's decouple the specific version of `wix-ui` from this idea because it shouldn't be.
____________________

**Note:** We don't necessarily have to use a provider consumer contract for that, we can choose any other way (Like putting a property on the window object for example). The idea here is that it will be transparent to the developers that work on the apps.

jonathanadler:
@ranyitz 

I see what you're saying, but again, I can't see how it will be possible to synchronise deployments between all the apps.

In addition, it is true that `isMobile` and `displayMode` are specific to the page, however, sometimes apps need to force a different value (for instance if they want to have a mobile design for tablets, or no mobile design at all). 
`rtl` is a bit different, since the editor doesn't support it, and doesn't pass a param representing it.

Regarding the note, The provider is already used, so we can't change it. In addition, we want to support ssr and so we can't rely on the window object.
Using the provider is much more readable, and maintainable, and also will be easier for testing IMO. It gives the developers using the library, complete control of how the components will behave in their app, regardless of what the editor's state will be.

The way I like to think of these things is, let's say, theoretically, that the editor decides to change its inner implementation, maybe OOI components are rendered back in an iframe, or maybe even the editor will dump React (idk 🤷‍♂ ), we don't want it to affect the users' app, meaning the app could still work and be released as is.

@gileck what do you think?

ranyitz:
@jonathanadler Thanks for the response. Here are my notes.
____________________

> developers using the library, complete control of how the components will behave in their app, regardless of what the editor's state will be.

We can't ignore the fact that those apps run in the viewer. I don't think that we should prepare for a situation that those apps could work without it.
________________

> The way I like to think of these things is, let's say, theoretically, that the editor decides to change its inner implementation, maybe OOI components are rendered back in an iframe, or maybe even the editor will dump React (idk 🤷‍♂ ), we don't want it to affect the users' app, meaning the app could still work and be released as is.

There is an inherent **tradeoff** between isolation (which improves developer velocity) and performance. For example, working with iframes gives very good **isolation** but it is bad for **performance** (that's the main reason to move out of it)

I think we can't maintain what you suggest here while still go into the path of a more performant platform.

______________

Giving every app to configure its own `isMobile`/`rtl` property is a decision, which I'm not sure that we want to support. It would give apps the ability to use properties in a not expected way (I would expect a component to work on a "mobile" mode or design when rendered in a mobile device and not differently on each app).

_____________________________

> I see what you're saying, but again, I can't see how it will be possible to synchronize deployments between all the apps.

Why do we need to synchronize? Can you elaborate?

jonathanadler:
@ranyitz I really understand what you're saying, and I agree with your approach. However, moving the provider to the editor will restrict future development, whereas adding this to each app is much more flexible. At least for now.

> I think we can't maintain what you suggest here while still go into the path of a more performant platform.

I don't think that **just** the `wix-ui-tpa` will effect the performance in such a way.
TPA's are using tons of 3rd party libraries. When `wix-ui-tpa` becomes one of the heaviest libraries, we'll obviously have to optimise. 

> Giving every app to configure its own `isMobile`/`rtl` property is a decision, which I'm not sure that we want to support. It would give apps the ability to use properties in a not expected way (I would expect a component to work on a "mobile" mode or design when rendered in a mobile device and not differently on each app).

We might not want to support it, but what we decide here won't change that. 
Companies will design and create whatever and however it suites them best.
This is something I saw in my time in the Restaurants company.

> Why do we need to synchronize? Can you elaborate?

Because if we make changes to the provider, we'll have to have the editor and the apps GA together. Then, if a certain app needs to rollback because of a bug, it will be a big mess.

Two more notes:
1. The OOI development api is defined by the props that the OOI app component receives from the editor. Adding a provider, is changing that api.
2. The provider already exists, and is used in production. I don't believe we'll change that soon. So at the moment developers will have to use it. 
The question is should we add it to the generator or not.
Using something other than the provider means relying on conventions, which is something I don't like so much.

With all that being said, I think this should be discussed and decided with @gileck.

Sorry for the "digging", and thanks for all the feedback!

ranyitz:
> Sorry for the "digging", and thanks for all the feedback!

Don't worry about it. I think this is an important discussion.
_________
> I don't think that just the wix-ui-tpa will affect the performance in such a way.

It's about the path that we choose, how our decisions aligned with what we're trying to achieve.

> We might not want to support it, but what we decide here won't change that.

I think it will, it will make it harder and "hacky" to do so, and therefore you'll see less of that. That's one of the powers of good APIs.

> Because if we make changes to the provider, we'll have to have the editor and the apps GA together. Then, if a certain app needs to rollback because of a bug, it will be a big mess.

I'm not sure that I understand when you'll need to do such a thing. If you are talking about breaking changes, there is also the possibility of making backward compatible changes and deprecations.

> The OOI development api is defined by the props that the OOI app component receives from the editor. Adding a provider, is changing that api.

I'm not sure that I understand, can you explain?

> The question is should we add it to the generator or not.

I agree. This is the main question here. And I'll give my answer to it:

I think that **we should add it only once, in the platform level**. 

So it will be:

* **Learning curve** - Fewer things you should learn when creating OOI app. (**JustWorks™**)
* **Fewer code duplications** - Being able to change things in one place (instead of in every OOI app)
* **Less boilerplate** - Less code = easier maintainability (and fewer places that the app developer can make mistakes in)

jonathanadler:
I have to say I really don't agree.

Using providers is a well known practice, and will have almost no impact on learning curve.
It's really basic, and pretty self explanatory.

Fewer code duplication, well that's true, and that's why we want it in the generator.

I think having an "invisible" provider is not an api that will force good coding. 
This is a convention the developers will have to know about, in order to understand why their components are acting the way they are.
If the components won't act the way they want them to, they'll just write other components, or open pull requests blowing up our components props. I know, because i've done that many times as a consumer.

Anyways, @gileck , @shlomitc and I talked about this, and none of us want to add the component to the platform.
I suggest you talk to them, and see if you can convince them otherwise.
If it's up to me, then i'm all against it, sorry.

Maybe we should set a meeting for all of us, and talk about this f2f?

shahata:
Didn't read the whole thing, but after hearing the tl;dr just want to give my personal opinion: It is true that generators become more and more bloated with boilerplate, but this is what we want to solve when doing the Wext project. The way to get rid of boilerplate is to create better abstractions. Most of the time we don't need to move the code to the platforms since platforms also need to be slim just as much and even more than the apps need to be slim. cc @shlomitc @yairhaimo @asafyonaywix

PR #1248, opened 2019-04-24
#899 change mocha to jest

dmfilipenko:
### 🔦 Summary
Change ~mocha~ -> **Jest** in templates

PR #1289, opened 2019-05-13
Replace mocha with jest for templates

yavorsky:
During the guild-week, we had an intention to migrate all projects from mocha to jest, and @dmfilipenko opened a great PR.
I'm going to just finish it and moved everything into yoshi origin.

Original PR: https://github.com/wix/yoshi/pull/1248


yavorsky:
@ranyitz @ronami 
I think current jest-yoshi-preset is not configured to test node.js apps (like server app).
It proposes to have `spec` files and runs it inside `jsdom` environemnt, which is not correct. We want `node` environment here. Also, e2e will start puppetteer and run the browser, which is also not expected. We want just node js env here.

For example, node.js has some specific methods of timeouts. And node-platforms is using it:
[`this._timer.unref();`](https://github.com/wix-platform/wix-node-platform/blob/master/composer/wnp-bootstrap-composer/lib/experiments/platform-experiments.js#L12)

So we don't have any chance to test node apps with `jest-yoshi-preset`. We can use extend it or get rid of it for server projects.

Please correct me if I'm wrong.

packages/create-yoshi-app/templates/library/javascript/{%packagejson%}
```diff
@@ -35,6 +37,9 @@
   "yoshi": {
     "hmr": "auto"
   },
+  "jest": {
```
yanivefraim: You now have two `jest` configurations in this file... 
We decided not to use  the `jest-yoshi-preset` for library generators, because it is not really needed here (we do not need any of its features, correct me if I'm wrong)

packages/create-yoshi-app/templates/business-manager-module/javascript/{%packagejson%}
```diff
@@ -22,6 +22,7 @@
   "devDependencies": {
     "@wix/business-manager": "latest",
     "@wix/business-manager-api": "latest",
+    "@wix/react-test-wrappers": "^1.0.94",
```
yanivefraim: why do we need this in our generator?

packages/create-yoshi-app/templates/library/javascript/{%packagejson%}
```diff
@@ -24,7 +24,9 @@
   "devDependencies": {
     "husky": "~0.14.0",
     "lint-staged": "^7.2.2",
-    "yoshi": "^4.0.0"
+    "yoshi": "^4.1.0",
+    "puppeteer": "^1.1.0",
```
yanivefraim: We do not really need `Puppeteer` here (if we choose to stay with  `jest-yoshi-preset` (see my comment below), we can remove this when this [PR](https://github.com/wix/yoshi/pull/1242) is merged. If we just use plain `jest`, we do not need `Puppereer` anyway.

packages/create-yoshi-app/templates/server/javascript/test/it/server.spec.js
```diff
@@ -1,4 +1,3 @@
-import { expect } from 'chai';
```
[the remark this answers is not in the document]
ranyitz: We've decided that the usage of `jest-yoshi-preset` in server/node projects would not be a good way to go - https://github.com/wix/yoshi/pull/1242#issuecomment-492963412

PR #1293, opened 2019-05-15
Verify typescript references before build/start

ronami:
### 🔦 Summary

This PR adds a script that uses Lerna to infer the local dependencies of packages in an `experimentalMonorepo` project and update its `tsconfig.json` files.

See https://github.com/Microsoft/TypeScript/issues/25376.


packages/yoshi/src/commands/utils/index.js
```diff
@@ -1,14 +1,65 @@
+const os = require('os');
+const fs = require('fs-extra');
+const path = require('path');
+const chalk = require('chalk');
 const { partition } = require('lodash');
 const importCwd = require('import-cwd');
+const { isEqual } = require('lodash');
```
yairhaimo: Lets import it together with `partition` two rows above

PR #1301, opened 2019-05-20
Ignore monorepo hoisted dependencies in server bundle

ronami:
### 🔦 Summary

__This PR is aimed specifically to help with monorepo support.__

Currently, importing modules (from `node_modules`) in a server bundle will not treat them as external if they are hoisted (or just installed in root).

This PR adds a rule to ignore monorepo-level dependencies if the experimental flag is being used.


netanelgilad:
@ronami  Maybe late to the party, but did you consider [`modulesFromFile`](https://www.npmjs.com/package/webpack-node-externals#optionsmodulesfromfile-false). 
I think it's a better solution because it doesn't rely on heuristics on where the modules are actually located. (and stuff like yarn-pnp totally break the current solution).

PR #1302, opened 2019-05-20
Upgrade react to 16.8

yanivefraim:
This will enable the use of [React Hooks](https://reactjs.org/docs/hooks-intro.html)


ranyitz:
@yanive the label `PR: chore` is not listed in our changelog, and I think that even though it's just a bump in a version I would like to see it in the changelog :)

yanivefraim:
> Please note that

Good catch. My search and replace missed them (and I forgot, although I knew about them)

PR #1305, opened 2019-05-21
Update generators to use Webpack's dynamic import

yanivefraim:
This fix will make the generator use Webpack's dynamic import for `yoshi-template-intro`.


packages/create-yoshi-app/templates/client/javascript/src/components/App/App.js
```diff
@@ -14,14 +14,10 @@ class App extends React.Component {
   };
 
   /* <-- Feel free to remove this lifecycle hook */
+  /* <-- Please also remove `yoshi-template-intro` fom package.json */
```
ronami: fom -> from
maybe change to: "Please also remove `yoshi-template-intro` from your `package.json`"?

packages/create-yoshi-app/templates/client/javascript/src/components/App/App.js
```diff
@@ -30,10 +26,10 @@ class App extends React.Component {
 
     return (
       <div className={s.root}>
-        {/* <-- Feel free to remove h2 and TemplateIntro */}
         <h2 className={s.title} data-testid="app-title">
           {t('app.title')}
         </h2>
+        {/* <-- Feel free to remove h2 and TemplateIntro */}
```
ronami: No need to ask to remove the h2 in my opinion

packages/yoshi-template-intro/package.json
```diff
@@ -24,7 +25,7 @@
     "yoshi": "4.6.1",
     "yoshi-style-dependencies": "4.6.0"
   },
-  "dependencies": {
+  "peerDependencies": {
```
ronami: I think we should also add these to dev deps

Issue #1177, opened 2019-03-25
Need to refresh jest's watch mode when updating jest-yoshi-config


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

During jest's watch mode, if you make changes in `jest-yoshi.config.js`, you need to exit jest's watch mode and rerun it so the config will be reevaluated.

### 🤔 Expected Behavior

Change's in `jest-yoshi.config.js` will not require a jest rerun.

### 🔦 Context
When working on fullstack/client projects, we find ourselves changing puppeteer `decvtools` options from `false` into `true` in order to debug our code.

![image](https://user-images.githubusercontent.com/11733036/54920558-317d2e00-4f0c-11e9-93e6-64012f2ab61c.png)

While this is just an example, it happens many times and the required refresh to the watch mode slows the work flow.

### 💁 Possible Solution 

Do not cache the config load and call it within `jest-environment-yoshi-puppeteer`'s global setup function:

https://github.com/wix/yoshi/blob/f2b95b83f3628bc51ac0988cfd672b1760b8f56d/packages/jest-environment-yoshi-puppeteer/globalSetup.js#L25

### 👣 Steps to Reproduce

1. Create a new project.
2. Run `npm start`
3. Run the tests in watch mode `npx jest --watch`
4. Change the `devtools` option in `jest-yoshi.config.js` into true.

(The browser would not be opened. After a refresh to the jest process it will be) 

### 🌍 Your Environment
```
yoshi v4.1.0-rc.4
```

brumeregan:
Hi,
I and @vozhekhovskyi are going to resolve this issue at our guild week.

ranyitz:
Thanks @brumeregan and @vozhekhovskyi! Good luck :)

brumeregan:
@ranyitz 
Looks like this can not be resolved easily, because the whole file is invoked once by jest setup, and do not detect any updates. --no-cache and --clearCache didn't help.
Also we didn't find any information that jest allows us to update config in watch mode

ranyitz:
Fixed on #1307

PR #1307, opened 2019-05-22
issue-1177: refresh jest's watch mode when updating jest-yoshi-config

brumeregan:
PR for issue https://github.com/wix/yoshi/issues/1177

We found that `require` cached file, so to update config file in watch mode we've to clean cache with `delete require.cache[require.resolve(module)]`

### DIFF
--- a/packages/jest-environment-yoshi-bootstrap/environment-setup.js
+++ b/packages/jest-environment-yoshi-bootstrap/environment-setup.js
@@ -1,6 +1,6 @@
-const loadYoshiConfig = require('yoshi-config/jest.js');
+const loadJestYoshiConfig = require('yoshi-config/jest');
 
-const jestYoshiConfig = loadYoshiConfig();
+const jestYoshiConfig = loadJestYoshiConfig();
 
 const bootstrapConfig = jestYoshiConfig.bootstrap;
 
--- a/packages/jest-environment-yoshi-bootstrap/environment-setup.js
+++ b/packages/jest-environment-yoshi-bootstrap/environment-setup.js
@@ -1,4 +1,6 @@
-const jestYoshiConfig = require('yoshi-config/jest');
+const loadYoshiConfig = require('yoshi-config/jest.js');
+
+const jestYoshiConfig = loadYoshiConfig();
 
 const bootstrapConfig = jestYoshiConfig.bootstrap;
 
--- a/packages/jest-environment-yoshi-puppeteer/globalSetup.js
+++ b/packages/jest-environment-yoshi-puppeteer/globalSetup.js
@@ -17,6 +17,7 @@ const { shouldDeployToCDN } = require('yoshi-helpers/queries');
 const { getProcessOnPort } = require('yoshi-helpers/utils');
 const { setupRequireHooks } = require('yoshi-helpers/require-hooks');
 const cdnProxy = require('./cdnProxy');
+const loadJestYoshiConfig = require('yoshi-config/jest');
 
 // the user's config is loaded outside of a jest runtime and should be transpiled
 // with babel/typescript, this may be run separately for every worker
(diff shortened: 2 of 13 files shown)


packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -33,7 +33,15 @@ const serverLogPrefixer = () => {
   });
 };
 
+const requireUncached = module => {
+  delete require.cache[require.resolve(module)];
+  return require(module);
+};
+
 module.exports = async () => {
+  const configPath = path.join(process.cwd(), 'jest-yoshi.config.js');
```
yavorsky: Why do we need to change approach here and not use `'yoshi-config/jest'` path like beofre?

brumeregan: because this file (`yoshi-config/jest`) doesn't react on changes, but we need up-to-date user config file

ranyitz:
@yavorsky @brumeregan Good points!

We don't want to reimplement the functionality of reading the jest config in multiple places.

I think that we don't really need the cached nature of our API in the places where we load the config.

I'll elaborate.

__Those are the usages in the jest config:__

https://github.com/wix/yoshi/blob/1af7bf7479c5868da50346133b94c5cea4f53642/packages/jest-environment-yoshi-puppeteer/globalSetup.js#L25
https://github.com/wix/yoshi/blob/10346e6c2fb074190af8ce5b35fc8168811f40bf/packages/jest-environment-yoshi-bootstrap/environment-setup.js#L1
https://github.com/wix/yoshi/blob/69e8940575e586feb4c5793fda82bd3ee6481925/packages/jest-environment-yoshi-puppeteer/index.js#L10

We would like to load the config every time we call `globalSetup` or load the environment (So it will be responsive to changes during watch mode).

We can not do it using the current API (not in a hacky way 😄). So I'll suggest a new one:

```js
const loadJestConfig = require('yoshi-config/jest');

const jestConfig = loadJestConfig();
```

We could use [`import-fresh`](https://github.com/sindresorhus/import-fresh) to ignore require-cache [in the load jest config function](https://github.com/wix/yoshi/blob/master/packages/yoshi-config/jest.js#L19)

I don't mind breaking this API as we are the only ones that use it.

What do you say?

packages/jest-environment-yoshi-bootstrap/environment-setup.js
```diff
@@ -1,4 +1,6 @@
-const jestYoshiConfig = require('yoshi-config/jest');
+const loadYoshiConfig = require('yoshi-config/jest.js');
```
ranyitz: maybe `loadJestYoshiConfig` (because we do have `yoshiConfig` which is something else)?

packages/jest-environment-yoshi-puppeteer/globalSetup.js
```diff
@@ -34,6 +32,8 @@ const serverLogPrefixer = () => {
 };
 
 module.exports = async () => {
+  const loadYoshiConfig = require('yoshi-config/jest.js');
```
ranyitz: This line could move to the top like any other `require` statement

packages/jest-environment-yoshi-puppeteer/index.js
```diff
@@ -7,7 +7,9 @@ const { setupRequireHooks } = require('yoshi-helpers/require-hooks');
 // with babel/typescript, this may be run separately for every worker
 setupRequireHooks();
 
-const jestYoshiConfig = require('yoshi-config/jest');
+const loadYoshiConfig = require('yoshi-config/jest.js');
```
ranyitz: Can be moved to the top

brumeregan:
@ranyitz looks like it works =)

packages/jest-environment-yoshi-bootstrap/environment-setup.js
vozhekhovskyi: fixed

packages/jest-environment-yoshi-puppeteer/globalSetup.js
vozhekhovskyi: fixed

packages/jest-environment-yoshi-puppeteer/index.js
vozhekhovskyi: fixed

PR #1312, opened 2019-05-22
Revert testkit server

yanivefraim:
We had issues with Ambassador while running server on forked process, reverting for now.


ranyitz:
@yanivefraim Can you explain what were the problems?

yanivefraim:
@ranyitz `ambassador-testkit` puts environment variables which ambassador reads. This all thing is breaking when starting the server as a fork.

We can fix ambassador, but for now it is important that things will work.

PR #1313, opened 2019-05-22
Added better SSL certificate support for yoshi local server

amiryonatan:
### 🔦 Summary
Added a new certificate to `yoshi` with `altNames`.
This will allow everyone to work with the same certificate, and once trusted, all local SSL communication will be trusted.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.11.0",
+  "lerna": "3.15.0",
   "packages": [
     "packages/*"
   ],
@@ -13,15 +13,6 @@
       "PR: Polish :nail_care:": ":nail_care: Polish",
       "PR: Docs :memo:": ":memo: Documentation",
       "PR: Internal :house:": ":house: Internal"
-    }
-  },
-  "command": {
-    "publish": {
-      "allowBranch": [
-        "master",
-        "version_*",
-        "hotfix_*"
-      ]
     }
   },
   "version": "4.9.3"
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "jest-environment-puppeteer": "^3.4.0",
     "jest-puppeteer": "^3.4.0",
     "jest-teamcity-reporter": "^0.9.0",
-    "lerna": "^2.11.0",
+    "lerna": "^3.15.0",
     "lerna-changelog": "^0.8.0",
     "lint-staged": "^7.2.0",
     "lodash": "^4.17.10",
--- a/scripts/createVersion.js
+++ b/scripts/createVersion.js
@@ -39,7 +39,7 @@ Promise.resolve()
   })
   // Create new tag/bump versions in `package.json` files
   .then(() => {
-    execa.shellSync(`${lernaPath} publish --exact --skip-npm`, {
+    execa.shellSync(`${lernaPath} version --exact`, {
       stdio: 'inherit',
     });
   })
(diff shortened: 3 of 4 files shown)


amiryonatan:
@ronami let me know what u think.
Thanks!

yurynix:
@amiryonatan thanks for working on that!

IMO, we should add fields that would make it clear where this certificate is came from,
currently it's not the case:

```
openssl x509 -in server.crt -text         

Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 12090694944137314152 (0xa7cac6fe10ea7f68)
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=localhost
        Validity
            Not Before: May 22 13:21:30 2019 GMT
            Not After : May 20 13:21:30 2024 GMT
        Subject: CN=localhost
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Subject Alternative Name: 
                DNS:localhost, DNS:local.wix.com

```

So maybe something like
```
[ dn ]
C=IL
ST=Tel Aviv
L=Tel Aviv
O=FED Infra
OU=Yoshi CDN
emailAddress=devex@wix.com
CN = localhost
```

Also we shouldn't have one certificate that is trusted by all users and it's private key exposed on the internet (this repo), `yoshi` should generate a new cert/key pair on each machine.

IMO, the path that we should take is:
1. on startup check if we have cert, if we don't we generate it and save in somewhere like `~/.yoshi-certs`
2. If user has certs generated, we check if user trusts it via:
```
security verify-cert -c ~/.yoshi-certs/server.cert
```
3. If user doesn't trust it, we ask him if he wants to, and add it via:
```
sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ~/.yoshi-certs/server.cert

```

packages/yoshi/src/tasks/cdn/certificate.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+};
+
+const createAndTrustCertificate = () => {
+  if (!fs.existsSync(certsFullDirName)) {
+    fs.mkdirSync(certsFullDirName);
+  }
+
+  if (!fs.existsSync(certFile)) {
+    const configPath = path.join(
```
amiryonatan: I don't know what's the correct way to build this path

packages/yoshi/src/tasks/cdn/certificate.js
```diff
@@ -0,0 +1,60 @@
...
+      `openssl req -config ${configPath} -new -x509 -sha256 -newkey rsa:2048 -nodes -keyout ${keyFile} -days 1460 -out ${certFile}`,
+    );
+  }
+
+  const isTrusted = isCertificateTrusted(certFile);
+  if (!isTrusted) {
+    console.log(`An SSL certificate was generated, and must be trusted.`);
+    console.log(`Please enter your sudo password to trust it.`);
+    execSync(
+      `sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ${certFile}`,
```
amiryonatan: Supporting only mac machines now

amiryonatan:
@yurynix @ronami 
Thanks for the comments.
Please take a look now :)

ronami:
Putting this here for future reference: https://github.com/davewasmer/devcert.

PR #1316, opened 2019-05-22
Refactor src file structure for ooi TypeScript template

ranyitz:
### 🔦 Summary
Following https://github.com/wix/yoshi/pull/1230 only on the JavaScript version this is the TypeScript one.

### 🗺️ Test Plan
ooi test suite is passing


packages/create-yoshi-app/templates/out-of-iframe/javascript/src/viewerApp/viewerWidget.js
```diff
@@ -3,8 +3,6 @@ import { withStyles } from '@wix/native-components-infra';
 
 export default {
   component: withStyles(App, {
-    cssPath: ['editorApp.stylable.bundle.css'],
+    cssPath: ['viewerWidget.stylable.bundle.css'],
```
ranyitz: I've also found this, and it didn't make sense 😅

packages/create-yoshi-app/templates/out-of-iframe/javascript/src/viewerApp/viewerWidget.js
```diff
@@ -3,8 +3,6 @@ import { withStyles } from '@wix/native-components-infra';
...
   }),
 };
-
-// This file must export a default export object with "component" key
```
ranyitz: This API should be self explendatory

ranyitz:
@gileck Hey! I've done the `src` file structure arrangement also in the TypeScript project.

I moved a lot of files, and I hope I did well. Your review would be great here. In case everything looks good, let's merge and continue to move from there.

About the types themselves. I think we should treat that as a separate task.

ranyitz:
Thanks for the review @gileck!

I'll merge and release a version, let's try to be aware of the following usages of `ooi` typescript template.

PR #1323, opened 2019-05-23
Simplify server rendering logic in full-stack templates

ronami:
### 🔦 Summary

Removes a redundant (and somewhat confusing) function in the rendering logic of full-stack servers.


ranyitz:
@ronami TypeScript seems to be angry about it

image

Issue #1319, opened 2019-05-23
Preserve cache per template (create-yoshi-app:dev)


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior
When working on one template (for example the `out-of-iframe` typescript) we use `create-yoshi-app:dev` which creates a project in a temp directory.

When we close the watch process we can open it again and Yoshi will ask us if we want to use the old session (In order to jump right back to development and not to run `npm install` again).

![image](https://user-images.githubusercontent.com/11733036/58248922-26a51500-7d66-11e9-81cc-6717ea5a9ec8.png)

The thing is that sometimes we would like to move on check the javascript version or a completely different template, today it will clear the cache completely as we preserve only the last session.
 
### 🤔 Expected Behavior + 💁 Possible Solution

The cache will be project-based, which means that when we go back to choose a project to work on we'll have a list of projects to choose from (without duplicates)

for example:
```
Choose one of the following old sessions, press "n" to start a new session.

> out-of-iframe-javascript
> out-of-iframe-typescript
> client-typescript
```

[The projects should be sorted by date, most recent first 😎]

### 🔦 Context

* Developers may work on multiple templates at the same time.
* `npm install` takes a lot of time and make developers less productive.
* Also answering questions regarding your name organization and such can happen once one first creation.

PR #1326, opened 2019-05-26
reuse createApp from dev and move relevant parts to bin

ranyitz:
### 🔦 Summary
`createApp` function was duplicated between `dev` and production usages.

It's better when the logging is being done via the bin file (`create-yoshi-app`) or the `dev` script, rather in the `createApp` function itself.

For that to work I've needed to add the `install: false` and `lint: false` capabilities, that works well for the unit tests.

These are preparations that needed to be done before we add [this feature](https://github.com/wix/yoshi/issues/1319) to enhance the dev experience.

### 🗺️ Test Plan
All tests are passing


packages/create-yoshi-app/scripts/dev.js
```diff
@@ -150,10 +101,35 @@ async function init() {
     }
   }
 
-  const { templateModel, workingDir } = await createApp({
-    useCache,
-    cacheData,
-  });
+  if (useCache) {
```
Schniz: `useFromCache` or `readFromCache`

ranyitz: Thanks!

PR #1346, opened 2019-06-02
Bump Axios to v0.19.0

ronami:
### 🔦 Summary

Just a maintenance PR. Currently, our generators are using an old and out-dated version of Axios that misses some security fixes.

Here's the [changelog](https://github.com/axios/axios/blob/master/CHANGELOG.md#0170-oct-21-2017) for reference. I didn't find any significant breaking changes other than https://github.com/axios/axios/pull/950.


packages/create-yoshi-app/templates/server/typescript/{%packagejson%}
```diff
@@ -20,7 +20,7 @@
     }
   },
   "devDependencies": {
-    "axios": "~0.16.0",
+    ~0.16.0",
```
ranyitz:
```suggestion
	"axios": "^0.19.0",
```

PR #1351, opened 2019-06-02
support startUrl option and accept multiple urls for `--url` flag

ranyitz:
### 🔦 Summary
This PR adds 2 things:
1. The ability to use multiple urls when running `yoshi start` and open multiple tabs with multiple different urls.

2. Use this option through the config (instead of as a cli option `yoshi start --url`)

____________

__Byproducts:__

1. I've noticed that there are cases during the regular `start` flow that we don't really need to open the browser (because we don't serve anything), the original `openBrowser` function may have handled an `undefined` argument and didn't open the browser, but I thought it would be clearer to show this from the command point of view.

2. I wanted to add unit tests for the `open-browser` wrapper function. Our `test:unit` suite caught my tests when I've used `open-browser.spec.js` but unfortunately, also the `test:integration` suite caught it and failed. My solution was to support both `spec` and `test` extensions for the `test:unit` suite. I think that it makes sense to move all of the unit tests to this format, so it will be clearer from the file name itself that this is a unit test and not an integration test. (I'll open a dedicated issue for that)

3. Added documentation for the `--url` option as well.

4. Added a schema validation for the new option in Yoshi config. I've realized that there is no test for the schema validation. It means that I could have missed it and users would notice it after it has been merged. I'll open an issue regarding adding a way to automatically test all of the newly added options in the config.

## 🎊 Context
When working on `out-of-iframe` app, you would like to open both the editor and the viewer when running `npm start`. This PR enables it.
You would also like to move the opened url to a variable and use it across tests inside a project or through different projects. Moving it to the config opens the door for using it in a js config and later on in a Yoshi preset.

### 🗺️ Test Plan
Added unit tests for `open-browser` wrapper module.


packages/yoshi-config/schema/yoshi-config-schema.js
```diff
@@ -183,6 +183,23 @@ const schema = {
       description: 'An experimental way to enable monorepo specific features.',
       type: 'boolean',
     },
+    startUrl: {
+      description:
+        'opens the browser on the specified url/s after running yoshi start',
```
ronami: Super small but I'd start the description with a capital letter.

PR #1353, opened 2019-06-03
allow hmr over production site

deanshub:
### 🔦 Summary
allowing HMR over production site

### Issue
currently we replace assets in production sites in order to live test our features\bugs.
for that we use either [requestly](https://chrome.google.com/webstore/detail/requestly-redirect-url-mo/mdnleldcmiljblolnjhpnblkcekpdkpa?hl=en) or [clientspecmap experiments](https://bo.wix.com/client-spec-map-experiment-server/experiments/specs.stores.ProductWidgetOOI_local/1380b703-ce81-ff05-f115-39571d94dfcd) or [pacman extension](https://chrome.google.com/webstore/detail/wix-statics-override/fhaehbcdbkccakpjgokgppjkggkmkmbl)

when we enable yoshi's HMR those hacks doesn't work, because the bundled asset first tries to establish web socket to the site it's requested from

### Solution
I configured webpack dev server to establish the websocket to the local CDN
https://github.com/webpack/docs/wiki/webpack-dev-server

### DIFF
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -784,7 +784,9 @@ function createClientWebpackConfig({
   if (isHmr) {
     addEntry(clientConfig, [
       require.resolve('webpack/hot/dev-server'),
-      `${require.resolve('webpack-dev-server/client')}?${https?'https':'http'}://localhost:${port}`,
+      `${require.resolve('webpack-dev-server/client')}?${
+        https ? 'https' : 'http'
+      }://localhost:${port}`,
     ]);
   }
 
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -784,7 +784,7 @@ function createClientWebpackConfig({
     addEntry(clientConfig, [
       require.resolve('webpack/hot/dev-server'),
       // Adding the query param with the CDN URL allows HMR when working with a production site
-      // because the bundle is requested from parastorage we need to specify to open the scoket to the local cdn
+      // because the bundle is requested from "parastorage" we need to specify to open the socket to localhost
       `${require.resolve('webpack-dev-server/client')}?${
         project.servers.cdn.url
       }`,
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -567,7 +567,7 @@ function createClientWebpackConfig({
   isDebug = true,
   isHmr,
   withLocalSourceMaps,
-  cdn: { port = 4200, https = false } = {},
+  cdn: { port = 3200, https = false } = {},
 } = {}) {
   const config = createCommonWebpackConfig({
     isDebug,
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -567,6 +567,10 @@ function createClientWebpackConfig({
   isDebug = true,
   isHmr,
   withLocalSourceMaps,
+  cdn: {
+    port = 4200,
+    https = false,
+  } = {},
 } = {}) {
   const config = createCommonWebpackConfig({
     isDebug,
@@ -783,7 +787,7 @@ function createClientWebpackConfig({
   if (isHmr) {
     addEntry(clientConfig, [
       require.resolve('webpack/hot/dev-server'),
-      require.resolve('webpack-dev-server/client'),
+      `${require.resolve('webpack-dev-server/client')}?${https?'https':'http'}://localhost:${port}`,
     ]);
   }
 
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -783,6 +783,8 @@ function createClientWebpackConfig({
   if (isHmr) {
     addEntry(clientConfig, [
       require.resolve('webpack/hot/dev-server'),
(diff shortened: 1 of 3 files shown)


packages/yoshi/config/webpack.config.js
```diff
@@ -567,6 +567,10 @@ function createClientWebpackConfig({
   isDebug = true,
   isHmr,
   withLocalSourceMaps,
+  cdn: {
+    port = 4200,
```
yurynix: I'm wondering, why this default?

yurynix:
Thank you for the PR @deanshub 
Could you please fix the lint?

packages/yoshi/config/webpack.config.js
deanshub: that's the yoshi default port for CDN, isn't it?

yurynix: 3200 AFAIK, https://github.com/wix/yoshi/blob/2d3989511f43c7da0dee82251cdbfaabf06c76ff/packages/yoshi-config/loadConfig.js#L40

deanshub: oh, thanks!

packages/yoshi/src/webpack-utils.js
```diff
@@ -168,6 +168,7 @@ function createDevServer(clientCompiler, { publicPath, https, host }) {
     overlay: true,
     // https://github.com/wix/yoshi/pull/1191
     allowedHosts: ['.wix.com'],
+    disableHostCheck: true,
```
ranyitz: Why is that needed?

packages/yoshi/config/webpack.config.js
```diff
@@ -783,7 +784,9 @@ function createClientWebpackConfig({
   if (isHmr) {
     addEntry(clientConfig, [
       require.resolve('webpack/hot/dev-server'),
-      require.resolve('webpack-dev-server/client'),
+      `${require.resolve('webpack-dev-server/client')}?${
```
ranyitz: 1. Maybe you can reuse the [`servers.cdn.url`](https://github.com/wix/yoshi/blob/8b0af07b65b05cab396f8c1fefb42ef981b9b611/packages/yoshi-config/loadConfig.js#L75) variable?

2. I think that there is no need to supply `port` and `https` as variables to the `createClientWebpackConfig` function, just require the config and use it directly from this file. (You will also support both `start` and `start-app` for free)

packages/yoshi/src/webpack-utils.js
deanshub: I got `"Invalid Host/Origin Header" warning` and it failed the request to the localhost and this is what I found as a solution https://github.com/webpack/webpack-dev-server/issues/1604

deanshub:
Though it's a nice idea for a test, it sounds pretty flaky and a bit complex because of all the moving parts (puppeteer + test site + fs change)
you sure you want it?

packages/yoshi/src/webpack-utils.js
ranyitz: > this is what I found as a solution

This is a **workaround**. It can open Yoshi's users to DNS rebinding. Please see: https://github.com/wix/yoshi/pull/1191#issue-266161062

@ronami What do you think?

ranyitz:
@deanshub We have many tests like this one in Yoshi's codebase.

Why would it be flaky?

deanshub:
the timeout thingy... I'll write one and we'll see how it goes

packages/yoshi/src/webpack-utils.js
ranyitz: @deanshub Which production site are you working on? Could it be a `*.wixsite.com`? if so we could add it to the `allowedHosts`.

packages/yoshi/config/webpack.config.js
ronami: I checked Webpack Dev Server's client code and this is not needed. They find the script tag of the bundle and use its protocol and URL.

https://github.com/webpack/webpack-dev-server/blob/dc81e23318043938333847963b9049435f48ecde/client-src/default/utils/createSocketUrl.js#L17.

packages/yoshi/src/webpack-utils.js
ronami: > > this is what I found as a solution
> 
> This is a **workaround**. It can open Yoshi's users to DNS rebinding. Please see: [#1191 (comment)](https://github.com/wix/yoshi/pull/1191#issue-266161062)
> 
> @ronami What do you think?

Unless it's necessary I think we should try to whitelist common stuff.

packages/yoshi/config/webpack.config.js
```diff
@@ -783,7 +783,11 @@ function createClientWebpackConfig({
   if (isHmr) {
     addEntry(clientConfig, [
       require.resolve('webpack/hot/dev-server'),
-      require.resolve('webpack-dev-server/client'),
+      // adding the query param with the cdn url allows HMR when working with a production site
```
```suggestion
      // Adding the query param with the CDN URL allows HMR when working with a production site
```

packages/yoshi/config/webpack.config.js
```diff
@@ -783,7 +783,11 @@ function createClientWebpackConfig({
...
+      // because the bundle is requested from parastorage we need to specify to open the scoket to the local cdn
```
```suggestion
      // because the bundle is requested from "parastorage" we need to specify to open the socket to localhost
```

PR #1358, opened 2019-06-04
Server template - improvements

hugebdu:
@ronami @ranyitz guys, a bit more according to what we've agreed.


packages/create-yoshi-app/templates/server/typescript/src/config.ts
```diff
@@ -0,0 +1,18 @@
...
+  // return your app-specific context - an object which contains your domain singletons
+  // (eg DB connections, kafka producers/consumers) shared by all your express, websocket and RPC services.
+  // you may return a Promise which resolves to an object, in that case bootstrap startup sequence will wait for
+  // that promise to resolve before bringing up the listeners (HTTP/kafka/RPC).
+  return {
+    petriScopes
+  } as {%ProjectName%}Context;
+};
+
+export interface {%ProjectName%}Context {
```
ronami: Will it make sense to call it `ProjectContext`? It's convenient that code doesn't have template variables unless it's mandatory so you can open it in your cod editor and not get a lot of syntax errors.

hugebdu: I'm fine with having a static name for this type, don't expect any ambiguity nor collisions.
However `ProjectContext` doesn't seem sound in that case.

@alexandervain suggested `ApplicationContext` or shortened `AppContext`.

WDYT?

packages/create-yoshi-app/templates/server/javascript/index.js
```diff
@@ -1,5 +1,6 @@
 const bootstrap = require('@wix/wix-bootstrap-ng');
 
 bootstrap()
+  .config('dist/src/config') // https://github.com/wix-platform/wix-node-platform/tree/master/bootstrap/wix-bootstrap-ng#wixbootstrapngconfigfileexportingfunction-this
```
ranyitz: I know that it's rather opinionated, but we usually put comments above the lines

```suggestion
  // https://github.com/wix-platform/wix-node-platform/tree/master/bootstrap/wix-bootstrap-ng#wixbootstrapngconfigfileexportingfunction-this
  .config('dist/src/config')
```

packages/create-yoshi-app/templates/server/typescript/src/config.ts
ranyitz: `AppContext` sounds great to me! 👍

packages/create-yoshi-app/templates/server/typescript/src/external-types.d.ts
```diff
@@ -1,3 +1,12 @@
 declare module '@wix/wix-run-mode';
 declare module '@wix/wix-express-csrf';
 declare module '@wix/wix-config-emitter';
+
+declare module '@wix/wix-bootstrap-ng' {
```
ranyitz: Just a note. I think that those types should probably be in the node platform (instead of in every project).

It will add completions for people that already generated projects in the past and it will add completions for JavsScript projects as well.

packages/create-yoshi-app/templates/server/typescript/src/config.ts
ronami: Sounds great 👍 

It being static will make it easier for us to maintain.

packages/create-yoshi-app/templates/server/typescript/src/external-types.d.ts
hugebdu: indeed, that's our plan (mentioned that during our last sync)
@alexeevg is on it, hopefully a matter of few weeks

packages/create-yoshi-app/templates/server/typescript/src/config.ts
hugebdu: [done](https://github.com/wix/yoshi/pull/1358/commits/d05ac3d30acfc6944cfb341aa3e9cca8dcfc0af0)

packages/create-yoshi-app/templates/server/typescript/src/external-types.d.ts
hugebdu: let's keep it for now due to our forthcoming course if you don't mind and then drop it?

ronami: Sounds good to me, this is PR is a big improvement, @ranyitz?

ranyitz: Of course 👍

PR #1364, opened 2019-06-05
feat(create-yoshi-app): change templates playground working directory

AndriiTsarenko:
### 🔦 Summary
Improving the development experience of a project's templates by adding `templatesPlayground` folder inside the `Yoshi` repository.
Previosly "development template" was creating in folder like that: 
`/private/var/folders/3x/jf5977fn79jbglr7rk0tq4d00000gn/T/4f504b9edb5ba0e89451617bf9f971dd` 
And for side by side comparation, you should open this folder in IDE.
For now, the template is creating in the folder `packages/create-yoshi-app/templatesPlayground` inside `Yoshi` repository.
Screen Shot 2019-06-05 at 3 04 03 PM

 

### 🗺️ Test Plan
...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist
 .DS_Store
 lerna-debug.log
 .changelog
+templatesPlayground
--- a/packages/create-yoshi-app/scripts/dev.js
+++ b/packages/create-yoshi-app/scripts/dev.js
@@ -4,7 +4,6 @@ process.on('unhandledRejection', error => {
 
 const fs = require('fs-extra');
 const path = require('path');
-const tempy = require('tempy');
 const chalk = require('chalk');
 const map = require('lodash/map');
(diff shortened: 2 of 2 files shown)


AndriiTsarenko:
I've tried this locally. And it works for me well. Can you provide a test case that will fail? That will highlight the problem that you afraid?

ranyitz:
> I've tried this locally. And it works for me well. Can you provide a test case that will fail? That will highlight the problem that you afraid?

It's not that something will fail now, it's that something might fail in the future which could take us some time to figure out.

The problem here is that there is a `node_modules` directory in the root of yoshi package, it consists of a bunch of dev dependencies. All generated templates will resolve to those modules in case they won't find a module which can create unexpected behavior.

Today we have some kind of **isolation**, I would like to keep it that way.

PR #1367, opened 2019-06-05
create-yoshi-app: custom entry point for server project templates

hugebdu:
@ranyitz @ronami please.
this is the last one from us for now.


packages/create-yoshi-app/templates/server/javascript/index-dev.js
```diff
@@ -0,0 +1,16 @@
...
+
+// tear down hooks
+function stopAndExitWith(code) {
+  return async () => {
+    await stop();
+    process.exit(code);
+  };
+}
+
+process.on('SIGHUP', stopAndExitWith(128 + 1));
```
ronami: Does it make sense to move those into the testkit? It's boilerplate code that may be confusing for some users.

If you want to allow developers to change this default we can have an option for it in the testkit. Wdyt?

hugebdu: So it is about the environment, not the testkit. Server project will have more than one testkit, plus any other tear-down logic, which is app specific (FS cleanups, ...).
All our testkits provide test framework hooks, but the environment is app specific.

We have [plans](https://github.com/wix-platform/wix-node-platform/issues/1138) and some [WIP](https://github.com/wix-platform/wix-node-platform/pull/2295) to provide a unified abstraction for IT/E2E testing environment. I guess moving process hooks there makes more sense than for a single testkit.

Once we have it, I'd make a PR to use it in the template.

WDYT?

PR #1369, opened 2019-06-10
Run minimal number of tests on PR CI

ronami:
### 🔦 Summary

This PR adds an experimental feature that tries to understand which files have changed in a PR and run only the minimal amount of tests.

In a nutshell: Run only relevant unit/component tests, similar to how `jest --changedSince` works ([see here](https://jestjs.io/docs/en/cli#changedsince)) and run all end-to-end tests.

I used somewhat undocumented APIs from Jest which may break without major versions and have somewhat clunky APIs (HasteMap). If they break often we can also run Jest twice on CI and adjust the TeamCity reporting to work with multiple runs.


packages/yoshi/src/commands/test.js
```diff
@@ -214,6 +219,48 @@ module.exports = runner.command(
           jestCliOptions.push('--runInBand');
       }
 
+      // Run minimal tests on PR CI
+      if (
+        inTeamCity() &&
+        process.env.agentType === 'pullrequest' &&
```
ranyitz: inPRTeamCity

packages/yoshi/src/commands/test.js
```diff
@@ -214,6 +219,48 @@ module.exports = runner.command(
...
+        // Run only if this project is using `jest-yoshi-preset`
+        projectConfig.jestConfig.preset === 'jest-yoshi-preset'
+      ) {
+        const { changedFiles } = await getChangedFilesForRoots(['.'], {
```
ranyitz: pass ROOT_DIR

packages/yoshi/src/commands/utils/dependency-resolver.js
```diff
@@ -0,0 +1,36 @@
...
+
+async function getDependencyResolver(cwd = process.cwd()) {
+  // Basic Jest config to generate HasteMap
+  const config = {
+    haste: {
+      providesModuleNodeModules: [],
+    },
+    moduleFileExtensions: ['js', 'ts', 'tsx', 'jsx'],
+    modulePathIgnorePatterns: [],
+    cacheDirectory: path.resolve(tmpdir(), 'yoshi'),
```
ranyitz: tempy

PR #1376, opened 2019-06-11
create-yoshi-app | feat (i18next with ICU format): i18next & ICU integration.

etaybarzilay89:
Using ICU format in i18next.

- adding i18next-icu to package.json of client/fullstack templates (both js and ts).
- enriching i18next with i18next-icu plugin.

### DIFF
--- a/packages/create-yoshi-app/templates/client/javascript/src/i18n.js
+++ b/packages/create-yoshi-app/templates/client/javascript/src/i18n.js
@@ -1,7 +1,9 @@
 import i18next from 'i18next';
+import ICU from 'i18next-icu';
 
 export default function i18n(locale) {
   return i18next
+    .use(ICU)
     .use({
       type: 'backend',
       read: (language, namespace, callback) => {
--- a/packages/create-yoshi-app/templates/client/javascript/{%packagejson%}
+++ b/packages/create-yoshi-app/templates/client/javascript/{%packagejson%}
@@ -35,6 +35,7 @@
   "dependencies": {
     "axios": "^0.19.0",
     "i18next": "^11.6.0",
+    "i18next-icu": "^1.1.2",
     "prop-types": "~15.6.0",
     "react": "16.8.0",
     "react-dom": "16.8.0",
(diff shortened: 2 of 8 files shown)


itaibenda:
This adds around *5.4kb gzip* to the bundle including `intl-messageformat` which is *4.79kb gzip*

yairhaimo:
Thanks for the contribution!

Two notes:
- Can you please explain what this PR does? What is the effect on the users? Why do we need it?
- This PR updates only some of the templates (missing bizmgr, ooi)

etaybarzilay89:
Yes of course:
- This PR adds an i18next-ICU plugin to i18next (our translation tool), which lets us use ICU syntax in our translations json (https://formatjs.io/guides/message-syntax/).
It means that whenever we use the "t" function, it evaluates the translation value in an "ICU way", similar to what happens today when we want to inject arguments to the translation value but with more powerful abilities.

It will help us solve existing translations problems (such as pluralization, gender, dates and so on).
You can read about it here (documentation is another PR we have):
https://github.com/wix-private/fed-handbook/pull/224

or play with it here:
https://github.com/wix-private/translations_icu_example_project

(It should also be on production, but CI has a bug so it's not yet there):
https://apps.wix.com/translations-icu-examples-project
.

- This PR should not effect our users without them using the new abilities of ICU syntax explicitly, this is because ICU is a super set of i18next syntax.
More over, because it only applies to new projects made by yoshi with i18next, there is no risk of not integrating correctly with not-yoshi or not-i18next projects.


- This PR updates only fullstack and client templates since in other templates the i18next version does not yet support ICU format. 
We saw that there is a PR to bump up i18next version on other templates as well. once they'll merged, it will be possible to apply the plugin on them as well.

yairhaimo:
Thank you.
Can you point me to the PR that bumps i18next in the other templates?
I rather that we first bump the versions across all templates and then update this PR to match, instead of doing it gradually.

itaibenda:
@yairhaimo https://github.com/wix/yoshi/pull/1344

yairhaimo:
That PR only bumps in the client and fullstack templates

itaibenda:
Because the other templates are using i18next version which isn't supported with ICU
The remaining templates could be updated if the following PR will bump their versions: https://github.com/wix/yoshi/pull/1344

itaibenda:
better solved here: https://github.com/wix/yoshi/pull/1407

PR #1377, opened 2019-06-11
Support web-worker bundle in Yoshi

ranyitz:
## 🔦 Summary

This PR should add the option to create bundles for a target of `web-worker` using Yoshi.

This will use a limited webpack configuration that will allow a specific set of things that are needed for web worker bundles.

The feature would be non-breaking and should let the user choose if a certain entry point should be bundled to run in a web worker.

An example of how the API would look like:

```js
// yoshi config

{
	entry: {
		app: "./app.js"
	},
	webWorker: {
		entry: {
			worker: "./worker.js"
		},
		externals: {
			react: "React",
		}
	}
}
```

## Features

1. Create a specific bundle for web worker (target `webworker` in webpack).
2. Support specific `externals` for the web worker bundle.
3. Refresh the browser after a change in the worker bundle.
4. Show an error overlay if there is a compilation error in the worker.
5. Recover from compilation errors and refresh the browser with the working state.

ooi-live-reload-low-res

## Concerns

~* Should both bundles have the same extensions? (`*.bundle.js`) If they will it will be predictable but there would be a possibility that two entries by the same name would override each other, while that's an edge case, the former API (only entry) didn't allow it.~

~* Webpack dev-server currently enables only one compiler at a time, so I configured the web worker config to watch for changes similar to the server bundle.~

~* If we'll want changes in the viewer script to trigger a refresh/hmr we'll need to specify it similarly to the server bundle, which means it won't be a general worker anymore, but a specific to the client app.~

_____________________________________

## Things to do

* [x] Add tests
* [x] Add live reload logic
* [x] Add error overlay
* [ ] Add to docs


packages/yoshi/config/webpack.config.js
```diff
@@ -937,9 +939,98 @@ function createServerWebpackConfig({ isDebug = true, isHmr = false } = {}) {
...
+
+    name: 'web-worker',
+
+    target: 'webworker',
+
+    entry: isSingleEntry(webWorkerEntry)
+      ? { app: webWorkerEntry }
+      : webWorkerEntry,
+
+    optimization: {
```
ranyitz: This is duplicated, should we extract it to a function? (We've tried to avoid doing that in the past, in order to keep the webpack config readable)

packages/yoshi/config/webpack.config.js
```diff
@@ -937,9 +939,98 @@ function createServerWebpackConfig({ isDebug = true, isHmr = false } = {}) {
...
+      ],
+
+      // https://webpack.js.org/plugins/split-chunks-plugin
+      splitChunks: useSplitChunks ? splitChunksConfig : false,
+    },
+
+    output: {
+      ...config.output,
+
+      // Bundle as UMD format if the user configured that this is a library
```
ranyitz: This is quite weird to use the same `umd` configurations for both the client app and the web-worker.

Another option would be to just configure it as follows:

```js
	  library: '[name]',
      libraryTarget: 'umd',
      globalObject: 'self',
```

And see in the future if we have more use-cases that require this configuration.

packages/yoshi/config/webpack.config.js
```diff
@@ -937,9 +939,98 @@ function createServerWebpackConfig({ isDebug = true, isHmr = false } = {}) {
...
+      // https://webpack.js.org/configuration/output/#output-umdnameddefine
+      umdNamedDefine: project.umdNamedDefine,
+    },
+
+    externals: project.externals,
+
+    plugins: [
+      ...config.plugins,
+      // https://webpack.js.org/plugins/define-plugin/
+      new webpack.DefinePlugin({
```
ranyitz: I don't know if those are needed, Looking at it back, maybe it's better to just remove them and add in case someone will ask for them.

packages/yoshi/config/webpack.config.js
```diff
@@ -937,9 +939,98 @@ function createServerWebpackConfig({ isDebug = true, isHmr = false } = {}) {
...
+        ),
+        'process.env.IS_MINIFIED': isDebug ? 'false' : 'true',
+        'window.__CI_APP_VERSION__': JSON.stringify(
+          artifactVersion ? artifactVersion : '0.0.0',
+        ),
+        'process.env.ARTIFACT_ID': JSON.stringify(getProjectArtifactId()),
+      }),
+
+      // https://webpack.js.org/plugins/loader-options-plugin
+      new webpack.LoaderOptionsPlugin({
```
ranyitz: I'm not sure if we still need this, this is needed for **webpack 1** loaders.

@ronami maybe you have input regarding why this was needed in the client config.

I planned on removing it if I'll see that all our loaders function without it.

netanelgilad:
> Should both bundles have the same extensions? (*.bundle.js) If they will it will be predictable but there would be a possibility that two entries by the same name would override each other, while that's an edge case, the former API (only entry) didn't allow it.

What do you mean by predictable? what actual benefits do you see for having the same extension? The disadvantage for having the same extension makes a lot of sense, trying to understand what the tradeoff here really is.

> If we'll want changes in the viewer script to trigger a refresh/hmr we'll need to specify it similarly to the server bundle, which means it won't be a general worker anymore, but a specific to the client app.

Since this is the only use case right now, seems correct to me. Since we haven't got  a request for general worker support, I see no need to try and accommodate it. IMO this point also stands for the issue about the target webpack config.

packages/yoshi/config/webpack.config.js
netanelgilad: I would try to avoid duplication, it would be a pitfall to forgetting to update one of the copies... but that's just me.

packages/yoshi/config/webpack.config.js
netanelgilad: If it's a kind of public API we established maybe users will expect it? I'm for waiting for a real use-case but if we keep it in the app entries and not have it here, it would be a weird discrepancy

ranyitz:
Thanks @netanelgilad, Awesome input!

I'm all into making this implementation as slim as possible, and suitable for the specific needs of ooi apps, at least for the MVP. The usage of web worker seems general enough so supporting it generally doesn't seem to be a bad thing, but that's ok to start with a naive implementation and evolve through use-cases.

> What do you mean by predictable? what actual benefits do you see for having the same extension? The disadvantage of having the same extension makes a lot of sense, trying to understand what the tradeoff here really is.

That's a good point. The only benefit is that we don't need to change anything, today we still configure `viewerScript.bundle.js` in some places, not something we can't change.

Would you prefer `viewerScript.worker.js`?

netanelgilad:
> Would you prefer `viewerScript.worker.js`?

Yeah, good as any I guess. It should be a hidden implementation detail, so as long there is no collision with `bundle.js` then 🤷‍♂

packages/yoshi/config/webpack.config.js
ronami: I think we should, it's becoming quite a bit of code 👍

packages/yoshi/config/webpack.config.js
ronami: Sounds good. Maybe in the future, we can use similar defaults in library flow.

packages/yoshi/config/webpack.config.js
ronami: I think this was needed for an old version of a loader. Not sure if we still need it since we bumped a lot of versions in v4.

Let's remove and see if all tests pass.

packages/yoshi/config/webpack.config.js
```diff
@@ -937,9 +939,98 @@ function createServerWebpackConfig({ isDebug = true, isHmr = false } = {}) {
...
+            libraryTarget: 'umd',
+            globalObject: 'self',
+          }
+        : {}),
+
+      // https://webpack.js.org/configuration/output/#output-umdnameddefine
+      umdNamedDefine: project.umdNamedDefine,
+    },
+
+    externals: project.externals,
```
ronami: Not sure how externals work for a Web-Worker but are we sure we want the same externals for both?

packages/yoshi/config/webpack.config.js
ronami: I think that supporting `DefinePlugin` with `process.env.NODE_ENV` is important. The others are debatable but also in the context of a standard app (not a web-worker).

I think that having the same `DefinePlugin` env vars for both types of entries will make it a lot less confusing. We can probably move it up to the common config so it's not duplicated.

packages/yoshi/config/webpack.config.js
```diff
@@ -937,9 +939,98 @@ function createServerWebpackConfig({ isDebug = true, isHmr = false } = {}) {
...
+            output: {
+              // support emojis
+              ascii_only: true,
+            },
+            keep_fnames: project.keepFunctionNames,
+          },
+        }),
+
+        // https://github.com/NMFR/optimize-css-assets-webpack-plugin
+        new OptimizeCSSAssetsPlugin(),
```
ronami: I don't think we need it here.

packages/yoshi/config/webpack.config.js
ranyitz: I don't think we would like to have the same externals for both.

I'm removing the main externals from the `webworker bundle` for now. I'll do some research and we may add something like `webWorkerExternals` if needed.

ranyitz:
Thanks @netanelgilad & @ronami for the great review!

I've fixed the PR according to the requested changes.

About the `.worker.js` suffix, doing the change in the PR is not problematic:

image

The problem is that tests won't pass, as the production site assumes there is a `viewerScript.bundle.js`. Changing this would be a breaking change that I rather do only after I know that the PR is green.

In any case, I don't find this issue critical, we can also just throw an error if 2 names are the same.

packages/yoshi/config/webpack.config.js
```diff
@@ -109,6 +109,23 @@ function prependNameWith(filename, prefix) {
   return filename.replace(/\.[0-9a-z]+$/i, match => `.${prefix}${match}`);
 }
 
+function createTerserPlugin() {
```
ranyitz: We use the terser plugin in both the worker and the web configurations, therefore export it to a function

packages/yoshi/config/webpack.config.js
```diff
@@ -941,9 +944,52 @@ function createServerWebpackConfig({
...
+
+      // https://webpack.js.org/plugins/split-chunks-plugin
+      splitChunks: useSplitChunks ? splitChunksConfig : false,
+    },
+
+    output: {
+      ...config.output,
+
+      // Bundle as UMD format
+      library: '[name]',
```
ranyitz: This is a specific configuration for the `ooi` setup, we can make it configuration in the future if needed.

packages/yoshi/src/webpack-utils.js
```diff
@@ -167,7 +167,7 @@ function createDevServer(clientCompiler, { publicPath, https, host }) {
     host,
     overlay: true,
     // https://github.com/wix/yoshi/pull/1191
-    allowedHosts: ['.wix.com', '.wixsite.com'],
+    allowedHosts: ['.wix.com', '.wixsite.com', '.ooidev.com'],
```
ranyitz: Most users will have `.wixsite.com` sites, but the initial site will be `www.ooidev.com`. Therefore I'm adding it here as an exception. (So HMR would work)

test/projects/kitchensink-typescript/integration/hmr.test.js
```diff
@@ -11,9 +13,11 @@ const clientFilePath = path.join(
 
 const serverFilePath = path.join(process.env.TEST_DIRECTORY, 'src/server.tsx');
 
+const workerFilePath = path.join(process.env.TEST_DIRECTORY, 'src/worker.ts');
+
 const originalServerContent = fs.readFileSync(serverFilePath, 'utf-8');
 
-describe.skip('hmr', () => {
+describe('hmr', () => {
```
ranyitz: That was skipped for no reason, not related, but let's enable it

test/projects/kitchensink-javascript/src/server.js
```diff
@@ -7,6 +8,16 @@ app.get('/other', async (req, res) => {
   res.send(await ejs.renderFile('./src/other.ejs'));
 });
 
+app.get('/web-worker-bundle.js', (req, res) => {
+  res.sendFile(path.join(__dirname, 'statics/worker.bundle.js'));
+});
+
+app.get('/web-worker.js', async (req, res) => {
+  res.send(
+    `self['worker-externals'] = 'Some external text';\n importScripts("/web-worker-bundle.js")`,
```
ronami: `res.sendFile()` so it is linted/highlighted, easier to maintain.

packages/yoshi/src/commands/start-app.js
```diff
@@ -47,6 +48,7 @@ const detect = require('detect-port');
 const host = '0.0.0.0';
 
 const https = cliArgs.https || project.servers.cdn.ssl;
+const isWebWorkerBundle = !!project.webWorkerEntry;
```
ronami: Export `isWebWorkerBundle` somewhere.

PR #1392, opened 2019-06-20
Open editor after npm start

yairhaimo:
### 🔦 Summary
This PR is targeted to reduce frustration. Many times you run `npm start` in a project and only then remember you want to open the code editor on this project too. You have to kill the app, open the code editor and run the app again. This PR lets you open your code editor after you start your app without killing it.

There are two options:
1. **A code editor is already open.** 
The following message will be displayed:
image

The project will be opened in a new instance of the current open editor.

2. **No code editor is currently open.**
When no code editor is open, the project will be opened in `Vim` by default in MacOS (also in CRA). I didnt think it was a good DX and @netanelgilad suggested the following solution.

The following shortlist will be displayed:
image

### Note: I am still checking whether to add IntelliJ in addition or instead of Webstorm.

------------- 

### Code wise:
1. I used `react-dev-utils`'s code in order to guess/open the code editor. I had to copy the code into Yoshi in order to export private functions. I took advantage of this in order to simplify and refactor the code.
2. I created an `onKeyPress` utility function and I was wondering where to put it.
I thought about `yoshi/packages/yoshi/src/commands/utils/index.js` but the name is misleading as it is a utility file for monorepos.

Thanks @ronami for the cool idea.


packages/yoshi/src/commands/start-app.js
```diff
@@ -172,6 +174,9 @@ module.exports = async () => {
     process.exit(1);
   }
 
+  console.log(chalk.white.bold(`To open this project in your code editor press ${chalk.cyan('E')}\n`));
+  onKeyPress('e', () => launchEditor(process.cwd(), 1, 1));
```
ronami: Can we use `ROOT_DIR` instead?

ronami: Also, since it's not immediately obvious, what about adding a comment mentioning that the 1s being passed are the row/column?

packages/yoshi/src/commands/start-app.js
```diff
@@ -172,6 +174,9 @@ module.exports = async () => {
     process.exit(1);
   }
 
+  console.log(chalk.white.bold(`To open this project in your code editor press ${chalk.cyan('E')}\n`));
```
ronami: I think this should be logged in `webpack-utils.js` since we clear the screen and print it again on every change. Wdyt?

packages/yoshi/src/commands/start-app.js
```diff
@@ -205,3 +210,17 @@ module.exports = async () => {
     persistent: true,
   };
 };
+
+
+function onKeyPress(key, cb) {
```
ronami: Can we move this to a different file? Like a utils file of some sort. Really minor tho, we can do it later on if this function grows.

PR #1395, opened 2019-06-23
Rtl css

liatash:
### 🔦 Summary

Adding PostCSS plugin that instead of creating 2 different css files (one for LTR and one for RTL) it creates a single one - and scope the LTR/RTL specific properties internally (as: `[dir=rtl] {selector}`

Cons:
1. Bigger bundle size (17% increase in our case, 17kb -> 20kb, minified gzipped)
2. A breaking change, requires adding a `[dir=ltr]` / `[dir=rtl]` on `<html>`

### 🗺️ Test Plan
...


AdirAmsalem:
@ronami is there anything else we should do to get it merged?

PR #1420, opened 2019-06-30
lock eslint-plugin-jsx-a11y

ronami:
### 🔦 Summary

See #1414.


ronami:
See v6.2.3 https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/CHANGELOG.md.

PR #1407, opened 2019-06-27
I18n generators update: use @wix/wix-i18n-config

vybu:
### 🔦 Summary
Use all new [@wix/react-i18n-config](https://github.com/wix-private/fed-infra/tree/master/wix-i18n-config) in all of the generators.
All changes made with as little of a footprint as possible.

Business manager generator was modified to load messages with dynamic imports.
Full stack example was modified to attach messages to templates during server time and use it synchronously during client side initialization, instead of loading them dynamically in client side.

PR #1426, opened 2019-07-01
Fixing i18n singleton issue for biz-mgr modules

yaelz:
When creating a biz-mgr app, i18n singleton was used and that caused the translations to not be loaded. Changed to use i18n instance.
Thanks @zivl for your help

Are the files generated by the generator tested?

### DIFF
--- a/packages/create-yoshi-app/templates/business-manager-module/javascript/src/i18n.js
+++ b/packages/create-yoshi-app/templates/business-manager-module/javascript/src/i18n.js
@@ -1,7 +1,7 @@
 import i18next from 'i18next';
 import i18nextXHRBackend from 'i18next-xhr-backend';
 
-const instance = i18next.createInstance()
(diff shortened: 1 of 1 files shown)


zivl:
@ronami if you recall, I've encountered this issue before... 
maybe it would be good to have it here ("by default") in `yoshi` as more and more wix apps are integrating with each other and `i18n` collisions might appear even more in the future

ranyitz:
@yaelz Thanks for opening this PR.

You are technically correct, and I think that I understand the problem.
There is another PR that replaces all `i18next` usage to wix's own abstraction. https://github.com/wix/yoshi/pull/1407

It also does exactly what you requested here (Use a specific instance), but if we'll merge this PR, it will clash with the other PR.

yaelz:
If it's a known issue and WIP I'm happy :)
Closing this...

PR #1440, opened 2019-07-03
update unit test files pattern to `*.test.js`

### LINKED ISSUE #1354: Move the unit test glob patterns from `*.spec.js` to `*.test.js`
## The Problem
Today we use `(src|test)/**/*.spec.js` in the `yoshi` package for `test:integation`.

We also use `**/__tests__/**/*.spec.js` for the unit tests.

This makes it possible to write a test that will run in both suites. This is also not clear whether a test is a unit test or an integration test.

_________________________

## The solution

We can use `**/__tests__/**/*.test.js` for unit tests, which is a different extension (`spec` => `test`) which makes it clear that those are from a different suite, and essentially enables us to write both unit and integration tests for the Yoshi package.

## What needs to be done?

I've needed to add unit tests on #1351 and therefore I've added support for both `*.test.js` and `*.spec.js` extensions.

It means that we can move all of the current unit tests from `*.spec.js` to `*.test.js` and remove the ability to add new unit tests under the `spec` extension.

agankarin:
### 🔦 Summary
Change unit test file's extension to be`*.test.js`. 
Only integration test will match the`*.spec.js` pattern.
Closes #1354 

### 🗺️ Test Plan
Verify that CI runs the same number of tests:
* Unit - 146 ✔️
* Integration - 310 ✔️
* Templates - 30 ✔️


ranyitz:
Thanks @agankarin Fixed on #1484

PR #1445, opened 2019-07-04
update mocha version to 6.0.0

rostyslavP:
update mocha version to 6.0.0

### DIFF
--- a/packages/yoshi/package.json
+++ b/packages/yoshi/package.json
@@ -102,7 +102,7 @@
     "mini-css-extract-plugin": "0.5.0",
     "minimist": "1.2.0",
     "mkdirp": "0.5.1",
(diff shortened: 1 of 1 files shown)


ranyitz:
@rostyslavP Updating mocha from `3.5.3` to `6.0.0` includes a lot of breaking changes.

It means that we should introduce it **only in a major version** (to prevent our users from breaking their builds without a notice).

Having said that, we should understand what are the benefits before we introduce breaking changes. We chose `Jest` as the main test runner, so it worth explaining the value in updating mocha before enforcing all of the developers that use mocha to update.

It would be great if you'll open an issue explaining why it's needed, and we'll have an open discussion of whether we should include it as part of yoshi@5.

PR #1447, opened 2019-07-05
Migrate `bootstrap-hot-loader` to TypeScript

ronami:
### Summary

Pretty straight-forward. See #1401 for details.

### DIFF
--- a/packages/bootstrap-hot-loader/tsconfig.json
+++ b/packages/bootstrap-hot-loader/tsconfig.json
@@ -1,0 +1,11 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "build"
+  },
+  "include": [
+    "src/**/*",
+    "src/**/*.json"
+  ]
+}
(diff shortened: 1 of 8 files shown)


packages/bootstrap-hot-loader/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "build"
+  },
+  "include": [
+    "src/**/*",
+    "src/**/*.json"
```
ranyitz: I've asked myself: why is this needed?

Answer to myself:

> If a segment of a glob pattern includes only `*` or `.*`, then only files with supported extensions are included (e.g. `.ts`, `.tsx`, and `.d.ts` by default with `.js` and `.jsx` if `allowJs` is set to true).

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

ronami: Also: https://github.com/Microsoft/TypeScript/issues/25636

ranyitz: @ronami That was the next question 😆

PR #1457, opened 2019-07-14
When using jest, do not ignore the transformation of svg assets from node_modules

amiryonatan:
### 🔦 Summary
Added support for loading svgs from 3rd party libraries


ranyitz:
@amiryonatan 

> **Added svg loader support for node modules**
> Added support for loading svgs from 3rd party libraries

Do you mean, **When using jest, do not ignore the transformation of `svg` assets which are located in node_modules directory**?

It would be nice if you could share your use-case.
__________________________

@ronami This line is getting incomprehensible:

https://github.com/wix/yoshi/blob/386269120ab3539bef8c50e57aedcd06b444d6c0/packages/jest-yoshi-preset/jest-preset.js#L101-L102

Is there a reason not to transform `md` files `jpg` files or any other asset (from `node_modules`)?

https://github.com/wix/yoshi/blob/386269120ab3539bef8c50e57aedcd06b444d6c0/packages/jest-yoshi-preset/jest-preset.js#L113

If that's true we can use a blacklist instead of a whitelist:
```regex
/node_modules/(.*?\.(js|ts|tsx)$)
```

So it will be clear it ignores the transformation of only `js` `ts` and `tsx` files.

What do you say?

amiryonatan:
@ranyitz 
There are 2 packages in play here:
1. responsive-editor-symbols - a module that contains svgs, and exports them as react components
2. responsive-editor-packages - renders the svgs.

When we run jest tests with components coming from `responsive-editor-symbols`, we get this error:
```
Warning: <%s /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.%s
    ReactComponent

        in ReactComponent
        in div
        in div
        in Unknown (created by WrapperComponent)
        in WrapperComponent

       5 |
       6 | global.console.error = (...args) => {
    >  7 |   throw new Error(args.join('\n'));
         |         ^
       8 | };
       9 |
      10 | global.testKitResolver = () => {
```

Because the transform function doesn't work on items from `node_modules`.
therefor, we added the same functionality as for stylable - to svg's as well

ranyitz:
@amiryonatan Ok thanks!

amiryonatan:
@ronami @ranyitz 
this PR is blocking the responsive-editor alpha version.
We can't push icons to the project since it's breaking our tests.

Is there any way we can rush this plz?

PR #1458, opened 2019-07-15
remove webpack lock from yoshi

matveyok:
### 🔦 Summary
Remove lock from webpack version due to this issue that was solved:
https://github.com/webpack/webpack/issues/8656

### DIFF
--- a/packages/yoshi/package.json
+++ b/packages/yoshi/package.json
@@ -137,7 +137,7 @@
     "tslint-react-hooks": "^2.1.0",
     "url-loader": "1.1.2",
     "wait-port": "0.2.2",
-    "webpack": "4.28.4",
+    "webpack": "^4.0.0",
     "webpack-bundle-analyzer": "3.3.2",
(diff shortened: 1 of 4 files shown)


matveyok:
I hope it will fix our problem and we will be able to migrate to yoshi 4.x.x :)

matveyok:
About the failing tests, the bundle is a bit different but still have what the tests wants:
image

What do you think, is it ok and change the test to have `\\` ?

@ranyitz

ranyitz:
@matveyok Yes! Nice digging into that.

Those tests are **not good** because they are too aware of the implementation.
In the future, we'll ditch them in favor of the kitchen sink tests.

________________________________

For now, just adjust the test, let's make it pass and I'll merge it :)

Again, sorry for the hassle, and thanks for the effort! 👍

matveyok:
@ranyitz I don't understand the lint build and how it passed before, I agree with the error... there is multiple tests with the same title 🤷‍♂️

PR #1463, opened 2019-07-17
Migrate `yoshi-helpers` to TypeScript

ronami:
### 🔦 Summary

Pretty straight-forward. See #1401 for details.


packages/yoshi-config/src/config.ts
```diff
@@ -72,6 +74,7 @@ export type InitialConfig = {
 export type Config = {
   name?: string;
   unpkg: string | unknown;
+  pkgJson: PackageJson;
```
ranyitz: @ronami I wanted to ask about the `pkgJson` addition.

Not that it's any problem, but why it was added?

PR #1467, opened 2019-07-22
Jest yoshi preset overrides

### LINKED ISSUE #861: Making jest-yoshi preset configurable
# 🙋 Feature Request

Jest has a lot of configuration options, but when using `jest-yoshi-preset` we only allowed a few of them.

# 💭 Thoughts

We can look at the approach presented in CRA - https://github.com/facebook/create-react-app/blob/49e258b4a6d04fb7e3542d7a060c6dcad4093564/packages/react-scripts/scripts/utils/createJestConfig.js#L72-L81
Which is, to whitelist a specific set of configuration options, that we know won't break the usage in the preset.

While it makes sense to control which options are whitelisted from the preset maintenance point of view, It really limiting the users.

It means that if I want to add coverage reports, my options are to ask yoshi maintainers to add a specific configuration option. Usually, when encountering of this kind of problems, I'm minded to create a project related feature and not to find the best API for everyone.

As a result, instead of using `jest-preset-yoshi`, I configured the environments myself, in a custom way ([example](https://github.com/wix-private/vmaf/blob/4b81d6aecb5f3141678ebc4298335ddcd7b203b2/vmaf-client/jest.config.js)).

I prefer to use `jest-preset-yoshi` instead of my own config, so I'll get updates.

### 🔦 Context 

These are the specific things that I needed to configiure in my project:
1. Disable `diagnostics` from `ts-jest`- It's an opinionated matter, that I prefer to run typechecking in precommit instead of during tests - https://kulshekhar.github.io/ts-jest/user/config/diagnostics
2. `collectCoverage`/`collectCoverageFrom`/`CoverageReports` - I want coverage reports and i need to configure them separatly for each project.
3. `testEnvironment`: [`enzyme`](https://github.com/FormidableLabs/enzyme-matchers/tree/master/packages/jest-environment-enzyme) instead of `jsdom`
4. `testEnvironmnetOptions: { "enzymeAdapter": "react16" }`

### 💁 Possible Solution

In order to create a predictable API we can use the following convention:

```js
{
  options:{
    // will be passed to the global config
    collectCoverage: true
  },
  component: {
    options: { 
      // will be passed to the component project config
      testEnvironment: 'enzyme'
      testEnvironmentOptions: {
        enzymeAdapter: 'react16'
      },
      setupTestFrameworkScriptFile: 'jest-enzyme'
    }
  },
  bootstrap: {
    options: {
      // will be passed to the bootstrap project config
    }
  },
  puppeteer: {
    options: {
      // will be passed to the puppeteer project config
    }
  }
}

```

yanivefraim:
Fixes #861.
For now, the only whitelisted value is `globals`. 
TODO: 
- update docs
- refactor `overrideJestConfig` method
- more tests?


packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -14,7 +21,25 @@ if (envs && envs.some(env => !supportedEnvs.includes(env))) {
...
 }
 
-module.exports = {
+const jestYoshiConfig = loadJestYoshiConfig();
+
+const projectOverrideMapping = {
+  e2e: 'e2eOptions',
+  spec: 'specOptions',
+};
+const supportedEnvOverrideKeys = ['globals'];
```
ranyitz: env -> project

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -56,6 +82,13 @@ module.exports = {
         return true;
       })
       .map(project => {
+        const projectOverrides =
+          jestYoshiConfig[projectOverrideMapping[project.displayName]];
```
ranyitz: maybe extract to 
```js
const projectOverrideKey = projectOverrideMapping[project.displayName]
```

packages/yoshi-config/src/jest/config.ts
```diff
@@ -26,4 +26,12 @@ export type Config = {
...
     port: number;
   };
+  specOptions?: any;
+  e2eOptions?: any;
+  collectCoverage?: boolean;
+  collectCoverageFrom?: [string];
+  coverageReporters?: [string];
+  coverageDirectory?: string;
+  coveragePathIgnorePatterns?: [string];
+  coverageThreshold?: any;
```
ranyitz: Maybe it's possible to just use the types from `@types/jest`

docs/guides/jest-setup.md
```diff
@@ -94,6 +94,24 @@ Has a default test timeout of 10 seconds.
 
 ## Configuration
 
+#### API
+
+- puppeteer - a configuration object to be passed as [options](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) to Puppeteer's launch method
```
ranyitz: Use backticks/bold for the names of the options

PR #1469, opened 2019-07-23
babel preset option: useLegacyDecorators

goloveychuk:
### 🔦 Summary
...

### 🗺️ Test Plan
...

### DIFF
--- a/packages/babel-preset-yoshi/index.js
+++ b/packages/babel-preset-yoshi/index.js
@@ -55,11 +55,13 @@ module.exports = function(api, opts = {}) {
       // Enable stage 2 decorators.
       [
         requireDefault('@babel/plugin-proposal-decorators'),
-        {
-          // Enable export after decorator syntax. It's also a part of the spec and tc39 is not made a decision about it.
-          // Read more https://github.com/tc39/proposal-decorators/issues/69
-          decoratorsBeforeExport: true,
-        },
+        options.useLegacyDecorators ? 
+          { legacy: true } : 
+          {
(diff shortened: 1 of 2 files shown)


ronami:
Thanks for the PR.

Since legacy decorators are deprecated and most likely won't be supported in JavaScript I don't think we should support them. What's the use-case? Can you use standard decorators instead?

goloveychuk:
I'm trying to migrate all packages in monorepo to one yoshi dep, some of packages use legacy decorators. (they expect [target, name, descriptor]), new babel passes only descriptor.
Also we cannot use typescript decorators in our project. 
I fixed my builds with `patch-package`, so you can close the pr, since the proper way is to migrate to latest decorators spec.
But maybe some other teams want to migrate to yoshi, they won't do this seamlessly, if they use old decorators.
Also, there is no way to override babel config, since you have
https://github.com/wix/yoshi/blob/879c989652053d6091a0a038c16d0302ac3b3ed7/packages/yoshi-helpers/src/utils.ts#L67
So no way to use whole yoshi workflow and add some custom plugin, e.g.

ranyitz:
@goloveychuk 

> But maybe some other teams want to migrate to yoshi, they won't do this seamlessly if they use old decorators.

Why it has to be seamless? Legacy decorators were (and still) on stage 2. If those people used any other JS features which are not in the spec yet, they would need to have some work done in order to use up to date syntax.

When enabling this option we are opening a door for divergent between our codebases. The same divergent that we are working hard to diminish.

Understanding two proposals for new syntax has a cost. Also switching between them when you go from one project to the other.

____________________

How hard would be the migration work so it justifies the concerns above?

PR #1470, opened 2019-07-23
added jest preset types pointer

amiryonatan:
### 🔦 Summary
Set the types property to point to your declaration file


ronami:
Thanks for the PR 🙏 

I'm not sure if this is necessary: The `types` prop is only used when a package is imported and you never directly import `jest-yoshi-preset`. We could try to have the types go into `node_modules/@types/...` so they are included automatically.

Does this solve a problem you encountered?

amiryonatan:
@ronami yeah:
right now I'm importing the types like that:
/// <reference types="../../../node_modules/jest-yoshi-preset/types" />

I don't want to expose page as a global, since we have tests that run either in sled / local e2e.
I have a class that supplies the relevant page object depending on the environment It's running

amiryonatan:
I would like to do:
/// <reference types="jest-yoshi-preset" /> which is cleaner and nicer :)

amiryonatan:
@ronami ?

PR #1476, opened 2019-07-30
[RFC] Yoshi server

ronami:
## Yoshi Server

Currently, setting up a simple [Node Platform](https://github.com/wix-platform/wix-node-platform) server requires a lot boilerplate: Setting up an [index.js](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/fullstack/javascript/index.js), [index-dev.js](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/fullstack/javascript/index-dev.js), [environment.js](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/fullstack/javascript/environment.js) and a somewhat complex [src/server.js](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/fullstack/javascript/src/server.js). We generate a lot of files that we don't expect developers to change or understand.

With `yoshi-server`, we take a step closer to including a runtime framework along with Yoshi as a build tool.

### Goals

- Reduce the amount of boilerplate required to write a server.
- Abstract the communication between the client and its server.
- Improve the developer experience with type completions for client-server interactions.
- Show clear error messages from the server in both, the terminal and the browser.
- Minimize the differences between full-stack and client projects.
- Make it easier for us to make non-breaking changes in the future.

### Proposal

Yoshi will provide its APIs, based on the [Node Platform](https://github.com/wix-platform/wix-node-platform), for creating a server.

#### Server functions

Yoshi will provide an abstraction for defining and consuming server APIs from the client (Inspired by [http-functions](https://github.com/wix-incubator/http-functions) by [shahata](https://github.com/shahata)). 

The API and concept are similar to [workerize-loader](https://github.com/developit/workerize-loader):

- Server functions can be defined as named exports from files with an `*.api.js` extension:

```js
export const greeting = function(name) {
  return {
    greet: `hello ${name}!`,
  };
};
```

- They can be invoked from the client by creating an API instance, importing a server function, and calling it with arguments. This will trigger an HTTP request to an endpoint on the server that will run the function with the correct arguments and return its response:

```js
import create from 'yoshi-server/client';
import { greet } from './greeting.api';

const api = create({ baseUrl: 'http://wix.com' });

api.request(greet, 12).then(data => {
  console.log(data.name);
});
```

- An [Express](http://expressjs.com) request and response objects are available on the server function context (`this`):

```js
export const greeting = function() {
  return this.req.query;
};
```

- Since server functions don't have type completions for `this`, and to prevent mistakes such as returning non-serializable values from them, they can be optionally wrapped with a helper function:

```js
import { wrap } from 'yoshi-server';

export const greeting = wrap(function(age: number) {
  // Adds type completions for `this`
  console.log(this.req);

  // Verifies that the returned value is serializable
  return {
    name: `world! ${age}`,
    age,
  };
});
```

- Invoking multiple server functions at once can be done with a single call to the server:

```js
import create from 'yoshi-server/client';
import { greet, bye } from './greeting.api';

const api = create({ baseUrl: 'http://wix.com' });

api.batch([greet, 12], [bye, 10]).then(([data1, data2]) => {
  console.log(data1, data2);
});
```

- Server functions support type completions for both, JavaScript and TypeScript projects, from defining server functions to consuming them and using their results. With [Ambassador](https://github.com/wix-private/ambassador), client code can even have type completions for RPC/Proto responses.
- For better developer experience, we can show a small indicator on the client if a call to a server function failed, possibly with some verbose error information that can be returned from a failed function. Similar to https://github.com/zeit/next.js/pull/6526.
- Since server functions are an abstraction for the client-server communication, we can switch the transport protocol (HTTP) later on without breaking any users. 
- Async return values should be supported: Returning a promise from a server function will wait until it's resolved before sending the response back.
- In component/unit tests, since we control the client library, we can either ask developers to fake it with `jest.mock` or provide our own abstraction for mocking server functions. This will require a custom [Jest](https://github.com/facebook/jest) `transform`  since server functions shouldn't run on the client.

#### Route functions

Since server functions are consumed from the client, we'll use route functions to expose routes to the outside world. Route functions are similar to server functions and support expressing routes (along with URL parameters) via the filesystem:

- Files in `src/routes` will be mapped to a route on the server with a URL that matches the filesystem. For example, `src/routes/users/create.js` will translate into `/users/create`.
- Named parameters can be used by wrapping the filename or directory name with `[]` and are available to the route function as `this.params`. For example: `src/routes/users/[pid].js` will map into `/users/:pid`.
- Similar to server functions, route files export a single function that returns a response and they can access [Express's](http://expressjs.com) request and response objects from the function context. It's recommended to also wrap them with `wrap` for type completions:

```js
import { wrap } from 'yoshi-server';

export default wrap(async function() {
  return {
    name: 'world!',
  };
});
```

- Runtime and 404 errors from route functions should show verbose responses in development. Currently, a blank page is shown. We should show verbose and pretty error pages, similar to [youch](https://github.com/poppinss/youch).

#### Common use cases

I want to discuss common use cases and how they should be used with `yoshi-server`:

- [wix-bootstrap-ng](https://github.com/wix-platform/wix-node-platform) initializes some APIs and makes them available on the [context](https://github.com/wix-platform/wix-node-platform/tree/master/bootstrap/wix-bootstrap-ng#context). Accessing the context should be available from every server or route function as `this.context`:

```js
import { wrap } from 'yoshi-server';

export const greeting = wrap(async function(age: number) {
  const experiments = await this.context.petri
    .client(this.req.aspects)
    .conductAllInScope('foo');

  return {
    experiments,
  };
});
```

- Rendering [EJS](https://github.com/mde/ejs) templates should be done by importing and calling `render()`. It accepts a template name and data, looks for it in `src/templates`, and returns the resulting HTML:

```js
import { render } from 'yoshi-server';

export default async function() {
  const html = await render('app', {
    title: 'hello world',
  });

  return html;
}
```

- Some server code requires users to be logged in. Limiting access to a server or route function only to logged-in users should be done by importing and calling `requireLogin()`. It should return a 403 (forbidden) error if the user isn't logged in:

```js
import { wrap, requireLogin } from 'yoshi-server';

export const greet = wrap(async function(age: number) {
  await requireLogin(this.req);

  return {
    name: `world! ${age}`,
    age,
  };
});
```

- Most [Node Platform](https://github.com/wix-platform/wix-node-platform) apps use an ERB configuration file to get runtime information from Chef. To avoid having to load the config in many server/route functions, `yoshi-server` will load the app's ERB config file and provide it as `this.config` (Separate config files can still be loaded with `this.context.config.load()`):

```js
import { wrap } from 'yoshi-server';

export const greet = wrap(async function(age: number) {
  const { clientTopology } = this.config;

  return {
    name: `world! ${age}`,
    clientTopology,
  };
});
```

#### Custom server and routing

Normally, you start your server with `yoshi start`, but you can override it and customize it with your own routes, middleware, [wix-bootstrap-ng](https://github.com/wix-platform/wix-node-platform) plugins, etc.

To use a custom server, start by creating an `index.js` file that starts up [wix-bootstrap-ng](https://github.com/wix-platform/wix-node-platform) and calls `express()` with your custom express function:

```js
import bootstrap from '@wix/wix-bootstrap-ng';

bootstrap()
  // Custom `wix-bootstrap-ng` plugins
  .use(require('@wix/wix-bootstrap-greynode'))
  .use(require('@wix/wix-bootstrap-hadron'))
  // A path to a custom express function
  .express(require.resolve('./your-express-function'))
  .start();
```

Your custom express function can add custom middleware and routing and eventually delegate to `yoshi-server`:

```js
import server from 'yoshi-server';

export default async (app, context) => {
  const handle = await server(context);

  // Use custom middleware, routing, db connection
  // or even mount `yoshi-server` on a different path
  app.get('/foo', (req, res) => {
    res.send('bar');
  });

  app.all('*', handle);

  return app;
};
```

#### Server functions' implementation

- Yoshi will use a [Webpack loader](https://webpack.js.org/loaders) to transform `*.api.js` files into files that contain metadata about how to consume those functions. For example, consider the following server function:

```js
export const greet = function() {
  return 'hello';
};
```

With the loader, it would be transformed into the following:

```js
export const greet = { fileName: '...', methodName: '...', };
```

- This DSL structure transformation is somewhat similar to the transformation [GraphQL](https://github.com/apollographql/graphql-tag) has.
- Similarily to [graphql-request](https://github.com/prisma/graphql-request) and [apollo-client](https://github.com/apollographql/apollo-client), we can create different and more specialized clients that consume server function metadata and make requests to the server. An added benefit is that we get full type completion for both, the arguments of every function and its return type.
- This RFC suggests creating an instance and passing it around to consume server functions. I think that this is required since most apps need some runtime configuration (baseUrl, headers, timeouts). Using a global approach will create a very misleading API. We can create neat APIs that help pass this instance around, similar to [@apollo/react-hooks](https://www.apollographql.com/docs/react/essentials/get-started/#connect-your-client-to-react).
- Since the loader only exposes metadata to consume server functions, other clients/abstractions can be built on top of it, similar to the different clients that consume [GraphQL](https://graphql.org/learn) queries.

### Final notes

- Now that most server boilerplate is provided as a library/framework it can be tested thoroughly. And because it has a very generic API (server and route functions), we can run a fake implementation of the [Node Platform](https://github.com/wix-platform/wix-node-platform) in local development or testing to speed it up. This *really* builds on the idea described in [overreacted.io](https://overreacted.io/how-does-the-development-mode-work/). 
- Hot reloading server and route functions should work in development without reloading the entire process or requiring wrapping functions with `hot` (Similar to [bootstrap-hot-loader](https://github.com/wix/yoshi/tree/master/packages/bootstrap-hot-loader)).
- Different return values from server and route functions can be supported, similarly to how it's done in [polkadot](https://github.com/lukeed/polkadot).


netanelgilad:
This is a super cool feature, and it's the beginning of a bright future! This will give `yoshi` superpowers, and as an app developer on top of `yoshi` it would help me focus on my business logic and less about infra and boilerplate.

As a general concern for me, I'd be interested in seeing a before and after example, where it would be easier to measure the value of the changes against the goals that were defined here. But maybe it's clear to everyone else.

Here are a few more specific stuff:

> Custom server and routing
Normally, you start your server with yoshi start, but you can override it and customize it with your own routes, middleware, wix-bootstrap-ng plugins, etc.

In our current generators, we are creating all projects with `greynode` and `hadron` plugins. If all projects need those plugins, that means that all apps will have to opt out to a custom server, which means we are keeping some of the boilerplate. I think we should address those plugins out of the box (and still keep them customizable, cause `greynode` can be configured with a list of topics, etc..)

> Currently, setting up a simple Node Platform server requires a lot boilerplate: Setting up an index.js , index-dev.js , environment.js  and a somewhat complex src/server.js . We generate a lot of files that we don't expect developers to change or understand.

As mentioned here, there is a lot of boilerplate with all the index files, environment, etc. This was not addressed, mostly the client-server communication was addressed. If I look at the list of goals, IMO reducing this boilerplate is more important than client-server communication, but that's just me I guess.

> With the loader, it would be transformed into the following:
`export const greet = { fileName: '...', methodName: '...', };`

I think this DSL is to aware of build tool concerns (mainly the `fileName` field). The DSL is important for composability, and avoiding `fileName` would help a lot.

Schniz:
It looks very good. I think that this whole idea is amazing. Throw in a runtime type checker (like io-ts) and I'm sold!

Probably everything I'm going to write is a bikeshed. Let's go!

> * Server functions can be defined as named exports from files with an `*.api.js` extension

`*.api.js` is not descriptive enough: it doesn't explain the context, which is a server file. I believe `*.server.js` would be a better fit.

> Invoking multiple server functions at once can be done with a single call to the server

I suggest have some kind of a builder pattern for the batch operations, for easier type checking/inference and to talk in "higher level" operations. Something like:

```ts
client.batch(
  batch(greet, 1, 2, 3) // Batch<[GreetFn]>
    .and(bye, 1, 2, 3) // Batch<[GreetFn, ByeFn]>
);
```

Binary operations are easier to statically type, and easier to understand

> * we can switch the transport protocol (HTTP) later

Or provide more functionality like performance monitoring, tracing, etc, and also, support multiple clients with multiple configurations! 👏 👏 

> * we can either ask developers to fake it with `jest.mock` or provide our own abstraction for mocking server functions.

I think it is good to emphasize how great this experience is: no more mocking HTTP calls or anything regarding the transport layer. You mock a function. If you use TypeScript, it will also be type-safe. amazing.

> * Optional named parameters can be used by wrapping the filename or directory name

what about required named parameters?

> ```js
> export const greeting = function(name) {
>   return {
>     greet: `hello ${name}!`,
>   };
> };
> ```

what about the following code:

```js
export const greeting = () => { throw new Error("Hello world!"); }
```

or

```js
export class MyError extends Error {}
export const greeting = () => { throw new MyError("This is a custom error!"); }
```

How would we address these errors on the client?

> * [wix-bootstrap-ng](https://github.com/wix-platform/wix-node-platform) initializes some APIs and makes them available on the [context](https://github.com/wix-platform/wix-node-platform/tree/master/bootstrap/wix-bootstrap-ng#context). Accessing the context should be available from every server or route function as `this.context`:

The `context` object is often untyped. It depends on the plugins you add. How can we type them correctly? How would we support bootstrap plugins?

> With the loader, it would be transformed into the following:
> 
> ```js
> export const greet = { fileName: '...', methodName: '...', };
> ```

Exporting a different type than the expected types (typescript will assume some kind of a function, but in reality, it will be an object). What about forcing the `wrap` function that will create an object on runtime as well? This is exactly what happens in GraphQL (`gql` template tag works on runtime, and works with babel/webpack plugins too)

deanshub:
A few questions,
1) why do we need the api creation? base URLs change often? Can't we take it from webpack public path and just import the api function and use it?
2) why inject the request to the scope and not get it as an argument?

packages/yoshi-server/src/server/render.ts
```diff
@@ -0,0 +1,16 @@
+import path from 'path';
+import ejs from 'ejs';
+import { TEMPLATES_BUILD_DIR } from 'yoshi-config/paths';
+
+export default async (templateName: string, data: any = {}) => {
+  const absolutePath = path.resolve(
+    TEMPLATES_BUILD_DIR,
+    process.env.NODE_ENV === 'development'
+      ? `${templateName}.debug.ejs`
+      : `${templateName}.prod.ejs`,
```
ranyitz: I'm wondering about `NODE_ENV === 'test'` case, currently it gets the `.prod.ejs`, is this the wanted behavior?

packages/yoshi-server/src/server/server.ts
```diff
@@ -0,0 +1,166 @@
...
+
+        socket.send(JSON.stringify({ success: true }));
+      };
+    }
+  }
+
+  private createRouter(): Router {
+    const router = new Router();
+
+    router.add('/_api_', async (req, res) => {
```
ranyitz: The `/_api_` route is also used in the `HttpClient` I think it makes sense to export it to a variable.

packages/yoshi-server/src/server/server.ts
```diff
@@ -0,0 +1,166 @@
...
+      const functions = this.createFunctions();
+
+      const body = await json(req);
```
ranyitz: It's not clear that the function `json` returns the body. Maybe it worth changing its name or adding a comment to https://github.com/zeit/micro#jsonreq--limit--1mb-encoding--utf8-

packages/yoshi-server/src/server/server.ts
```diff
@@ -0,0 +1,166 @@
...
+      };
+    });
+  }
+
+  private createFunctions(): {
+    [filename: string]:
+      | { [functionName: string]: DSL<any, any> | undefined }
+      | undefined;
+  } {
+    const serverChunks = globby.sync('**/*.api.js', {
```
ranyitz: Do we want to also start with `.ts` support?

packages/yoshi-server/src/server/server.ts
```diff
@@ -0,0 +1,166 @@
...
+      return send(res, 500, {
+        errors: [serializeError(error)],
+      });
+    }
+
+    return send(res, 404, '404');
+  };
+
+  private createDynamicRoutes(): Array<Route> {
+    const serverChunks = globby.sync('**/*.js', {
```
ranyitz: Same questions regarding `.ts` support

test/projects/yoshi-server-javascript/jest-yoshi.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  server: {
+    command: 'node index.js',
+    port: 3100,
+  },
+  specOptions: {
+    moduleNameMapper: {
```
ranyitz: it is worth explaining why it's here in a comment

test/projects/yoshi-server-javascript/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@types/jest": "^24.0.0",
+    "@types/node": "^8.9.4",
+    "@types/puppeteer": "1.12.1",
+    "@types/react": "^16.7.20",
+    "@types/react-dom": "^16.0.11",
+    "@wix/wix-bootstrap-testkit": "latest",
+    "@wix/wix-config-emitter": "latest",
+    "axios": "^0.19.0",
+    "ejs": "^2.6.1",
+    "express": "^4.15.0",
```
ranyitz: The user doesn't suppose to use `express`/`ejs` directly, maybe it worth reflecting it in the test project

yavorsky:
Sorry for delay with review :)
It looks really awesome! And a lot of cool things already suggested.
---
I would add few things:

Apollo-client is using small throttling to not send request immediately. For performance reasons it makes somethine like `batch` you proposed, but under the hood.

This example will actually send 2 requests:

```js
api.request(greet, 12);
// ...some code on the same thread
api.request(bye, 3)
```

but if we won't send requests immediately, we can gather everything for the lat 50ms (for example) and send 1 request, the same as using batch.

Also it can be optional. So you can use direct calls if you need, like: `api.requestNow(bye, 3)`
Moreover we can configure it for client:
```
const api = create({  ...,  throttlingDelay: 100  });
```
--- 
`api.request(greet, 12)`, `api.batch([a, 1], [b, 2])` looks good. But as for me, sometimes it could be not so obvious to run function w/o actually running it and pass arguments just next by as an arguments to a different method.

I think with webpack loader we can understand that `greet(12)` call is a call of http-function and convert direct call to the metadata we need:

`api.request(greet(12))` -> `api.request({ methodName: 'greet', fileName: 'somePath' , arguments: [12] })`
The same with batches: `api.batch(a(1), b(2))`.

---

Last thing: with some middleware or just single `onAction` callback for methods, we can has a super cool tool for debugging requests (like `redux-devtools`).
```
import { create, actionLogger }  from 'yoshi-server';
const api = create({ onAction: actionLogger })
// or
const api = create({ logLevel: process.env.NODE_ENV === 'development' ? 'debug' : 'error' })
```

```
REQUEST SENT:
  id: 1,
  methodName: 'greet',
  fileName: 'somePath/greet.js',
  arguments: [1, 2, 3]

REQUEST SENT:
  id: 2,
  methodName: 'bye',
  fileName: 'somePath/bye.js',
  arguments: [5],

RESPONSE RECEIVED:
  requestId: 1,
  statusCode: 200,
  time: 120ms

REQUESTS BATCH SENT:
  id: 3,
 . ..
```

PR #1477, opened 2019-07-30
added resolve symlinks flag

amiryonatan:
### 🔦 Summary
Added the option to use webpack resolve symlinks as exists in Yoshi@4


ronami:
Thanks for the PR @amiryonatan 🙏

Can you please explain what's the use-case for resolving symlinks?

amiryonatan:
@ronami 
In the ADI monorepo we have these dependencies:

A --> C
A --> B --> C

We use lerna to manage the dependencies, and therefor, all the project are linked.
A is our application, when we start / build it locally we get module C bundled twice inside, because webpack doesn't resolve the linked directories to the source directory.
When turning the flag on, the issue is fixed.

ranyitz:
We don't create new features for Yoshi@3 so I rather present it as an undocumented environment parameter.

Also, the `experimental` prefix is saved for testing new experimental features, which is not the case.

I suggest something like:

```js
WEBPACK_RESOLVE_SYMLINKS=true
```

Which will default to `false`, no need for documentation or tests in my opinion.

What do you think @ronami?

amiryonatan:
@ranyitz better ?

PR #1482, opened 2019-08-02
Create-yoshi-app: Better email input experience

yavorsky:
### 🔦 Summary
This PR is an intention to provide a bit better experience for new project's email enering.

I have not wix email in my Mac's git config. So I'm entering `artemy@wix.com` for any new project. `@wix.com` is good, I really love it, but I would enhance the autoformat mechanism a bit.

Here we're going to check:
- email from the git config (current solution)
- username for current user + @wix.com. Everyone I asked has the same username for wix and for wix's mac they have. (But it could confusing for non-wix mac experience. Let's discuss it)
- simple value w/o that has length, but doesn't match email pattern. It's enough to just write `username` and the promt will add `@wix.com` for you.

Also added some tests for email question logic.


packages/create-yoshi-app/src/getQuestions.js
```diff
@@ -1,6 +1,30 @@
 const getGitConfig = require('parse-git-config');
 const templates = require('../templates');
 
+const WIX_EMAIL_PATTERN = '@wix.com';
```
ranyitz: `WIX_EMAIL_DOMAIN`?

packages/create-yoshi-app/__tests__/getQuestions.spec.js
```diff
@@ -0,0 +1,81 @@
...
+describe('authorEmail', () => {
+  const mockEnvUser = value => (process.env.USER = value);
+  const currentUser = process.env.USER;
+  afterEach(() => {
+    process.env.USER = currentUser;
+  });
+  const EMAIL_VALIDATION_ERROR = 'Please enter a @wix.com email';
+
+  describe('initial', () => {
+    it('it should use email from git config', async () => {
```
ranyitz:
```suggestion
    it('it should use email from git config if it has a `@wix.com` domain name', async () => {
```

packages/create-yoshi-app/src/getQuestions.js
```diff
@@ -1,6 +1,30 @@
...
+  if (gitEmail.endsWith(WIX_EMAIL_PATTERN)) {
+    return gitEmail;
+  } else if (processUser) {
+    return `${processUser}@wix.com`;
+  }
+  return '';
+};
+
+// Format `value` to `value@wix.com` or use original value if it's already contains @wix.com.
+const formatEmail = email => {
```
ranyitz: const ensureWixEmail = maybeEmail => { ...

packages/create-yoshi-app/src/getQuestions.js
```diff
@@ -1,6 +1,30 @@
...
+  if (!email.endsWith(WIX_EMAIL_PATTERN)) {
+    return `${email}@wix.com`;
+  }
+  return email;
+};
+
+// Check if string is an email format.
```
ranyitz:
```suggestion
// Check if string is not in an email format.
```

PR #1486, opened 2019-08-06
Update business-manager-module

ronenst:


packages/create-yoshi-app/templates/business-manager-module/typescript/src/config.ts
```diff
@@ -1,9 +1,13 @@
...
+  ModuleId,
+  PageComponentId,
+  TModuleParams,
+} from '@wix/business-manager-api';
 
 // Note: after submitting PR to BM you can remove the constant strings below
 // Note: should be in sync with the config inside your BM module ERB
-export const MODULE_ID = ModuleId.{%ProjectName%} || '{%PROJECT_NAME%}';
-export const COMPONENT_NAME = PageComponentId.{%ProjectName%} || '{%projectName%}';
+export const MODULE_ID = /*ModuleId.{%ProjectName%} ||*/ '{%PROJECT_NAME%}' as ModuleId;
```
ranyitz: Why is this change needed?

If it's important, can we also do it on the javascript template? (`config.js`)

ronenst: Moving the loading code to `App.tsx` added `config.ts` to the import tree, which then caused the build to fail on a typescript failure:
http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=4513555&buildTypeId=Yoshi_TestBizmgr&tab=buildLog&branch_Yoshi=1486%2Fmerge

In the JS project there is no such validation, but I could do the same in the interest of consistency. WDYT?

I researched why this failure didn't happen initially and then discovered that `module.ts` and `client.tsx` where not being validated at all in BM projects, so I added the correct entry point to `tsconfig.json`.

Sorry if it's a bit out of scope for the original PR, I think I can try and split the technical changes to a new PR if you think I should.

ronenst: Clarification: I believe the reason that it did get validated once in `App.tsx` is because the test file is included in `tsconfig.json`.

ranyitz: Thanks for explaining in detail.

> In the JS project there is no such validation, but I could do the same in the interest of consistency. WDYT?

Yes, please.

ronenst: Maybe also remove the or (`||`) part?

ronenst: FYI: I'm trying another fix on the ts file, if that doesn't work or isn't good, I'll do the other thing.

ranyitz: awesome, thanks

ronenst: The other thing didn't work, it was a long shot. I fixed the JS templates as well, thanks.

PR #1492, opened 2019-08-09
fix migration guide v4 for yarn workspaces

zemlanin:
`yarn` workspaces can place package either in `{REPO_ROOT}/node_modules` or in the package directory (e.g. `{REPO_ROOT}/packages/{PACKAGE_NAME}/node_modules`)

Because it depends on version resolutions, dependency upgrade in one package could break another's relative paths to the `node_modules`. `import "yoshi/types"` solves this problem by offloading path resolution to `node`

---

related [PR](https://github.com/wix-private/site-search/pull/217) and [commit](https://github.com/wix-private/site-search/pull/217/commits/a3aa78e534e3e969a5ef85e16363be0c1c5f51d5)


docs/migration/version-4.md
```diff
@@ -87,6 +83,8 @@ Finally, remove any old types you had on assets handled by Yoshi (normally in `s
 ```diff
 -declare module '*.scss';
 -declare module '*.json';
++import "wix-ui-core/dist/src/globals";
```
bodia-uz: looks like `import "wix-ui-core/dist/src/globals"` should not be here.

docs/migration/version-4.md
```diff
@@ -75,10 +75,6 @@ Then, change your `tsconfig.json`:
 +    "resolveJsonModule": true,
     "..."
   },
-  "files": [
-+    "./node_modules/yoshi/types.d.ts",
-    "..."
-  ]
 }
 ```
```
bodia-uz: Need to update text about "remove any old types" to smth like "replace old types with types from yoshi"

docs/migration/version-4.md
zemlanin: Yeah, confused typings for `*.st.css` with the ones for `*.scss`

docs/migration/version-4.md
zemlanin: https://github.com/wix/yoshi/pull/1492/commits/82dd4fbf812560a848307e541d8b7015ced9d83e

docs/migration/version-4.md
zemlanin: https://github.com/wix/yoshi/pull/1492/commits/82dd4fbf812560a848307e541d8b7015ced9d83e

docs/migration/version-4.md
```diff
@@ -75,18 +75,15 @@ Then, change your `tsconfig.json`:
...
 
-Finally, remove any old types you had on assets handled by Yoshi (normally in `src/external-types.d.ts`):
+Finally, replace old types you had on assets (normally in `src/external-types.d.ts`) to [the ones provided by Yoshi](https://github.com/wix/yoshi/blob/v4.11.1/packages/yoshi/types.d.ts):
```
bodia-uz: Not sure, but mb relative link will be better. Pros - latest source. Cons is that if it moved to another path, link will be broken.
```suggestion
Finally, replace old types you had on assets (normally in `src/external-types.d.ts`) to [the ones provided by Yoshi](/packages/yoshi/types.d.ts):
```

bodia-uz: Ah. If it used outside github(yoshi site), direct link is better.

zemlanin: Con outweighs pro, imo

PR #1494, opened 2019-08-12
create server bundle for development entry

netanelgilad:
### 🔦 Summary
A simple approach to bundling the development entry when running `yoshi start`. The entry given to `yoshi start` will now be bundled into the `dist` folder under `devEntry.js` and then this file will be executed by `yoshi start`.

This will enable users to use `import/export` in their development environment code, and will also enable the use of hot reloading in that code (can be very useful when updating the mock services without restarting the app).

***Breaking Change*** - Since that code is now bundled and not run as-is, there will be subtle but important changes in the runtime (most importantly relative file references done outside the module system will break). We should probably add an experimental opt-in flag?

### 🗺️ Test Plan
Will add a test in app-flow tests.

### DIFF
--- a/packages/yoshi/config/webpack.config.js
+++ b/packages/yoshi/config/webpack.config.js
@@ -813,6 +813,7 @@ function createServerWebpackConfig({
   isDebug = true,
   isHmr = false,
   hmrPort,
+  devEntry,
 } = {}) {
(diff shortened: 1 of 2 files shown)


netanelgilad:
bump

PR #1496, opened 2019-08-13
Add ability to use process.env.ARTIFACT_VERSION

bodia-uz:
### 🔦 Summary
Currently to get `ARTIFACT_VERSION` in our code we need to use `window.__CI_APP_VERSION__`:
https://github.com/wix/yoshi/blob/b8feb9ac104e46eb7c54c6564d666d43058b8f60/packages/yoshi/config/webpack.config.js#L139-L141

But there is a case when your code is bundled not for browser (nodejs/web worker/?).

This change will add ability to use `process.env.ARTIFACT_VERSION` in same way as `process.env.ARTIFACT_ID`.

### 🗺️ Test Plan

**TODO**


ranyitz:
Thanks @bodia-uz.
Is this stopping you in any way?
If not, please close this and let's continue from #1602

bodia-uz:
It's not a blocker. Used `window. __CI_APP_VERSION__` with a comment in my code.
Closing issue.
Thanks @ranyitz

PR #1497, opened 2019-08-14
fix tests with `chalk`ed snapshots in color terms

zemlanin:
### 🔦 Summary
Snapshots with colored stdout do not contain color escape sequences (CES), so:
- on CI, shell don't have color support(?) => `chalk` does not add CES => snapshots match stdout => tests pass
- on developer's machine, there is a high chance of "modern shell" (like `zsh` or `fish`) => `chalk`  adds CES => snapshots differ from stdout => tests fail

### 🗺️ Test Plan
either

```
> zsh
> autoload -U colors && colors
> npm run test:unit
```

or

```
> fish
> npm run test:unit
```


zemlanin:
@ranyitz Are we waiting for someone else's approve? Because

Screen Shot 2019-08-30 at 10 46 03

ranyitz:
@zemlanin That's correct, I've probably waited for tests to pass, merging.

PR #1498, opened 2019-08-14
fix for node_module with svg in `*.st.css`

zemlanin:
failing test case for node_module with svg in `*.st.css`

TDD FTW

see also: https://github.com/wix-private/site-search/pull/249


ranyitz:
@zemlanin Thanks for sending this issue.

While adding a failing test is great, it would be also nice if you could communicate a bit what do you think is the **expected behavior**, **context** maybe add a **possible solution**.

> misconfigured svg-url-loader in yoshi@4

Do you think we should configure it differently?

Is it specific to `st.css`? is it specific to `node_modules`?

zemlanin:
> it would be also nice if you could communicate a bit what do you think is the expected behavior, context maybe add a possible solution

Sorry, was exhausted when wrote the original PR text. Added more details

> Do you think we should configure it differently?

I guess?..

> Is it specific to `st.css`? is it specific to `node_modules`?

Played with test cases a little more… It [*is*](https://github.com/wix/yoshi/pull/1498/commits/6da7102c162654a70516f680fb385754c3d8b0b2) specific to `st.css` and [*isn't*](https://github.com/wix/yoshi/pull/1498/commits/48521be993f476f588b118020f2b602653f5f7de) to `node_modules`

ranyitz:
@zemlanin Thanks for the great write-up! I think that the issue is well explained. I'm not sure how to solve it right now, but hopefully, the yoshi-team or someone from the community can solve it and contribute the solution.

PRs and discussion welcome!

PR #1500, opened 2019-08-15
support eslint for typescript

### LINKED ISSUE #1266: deprecate tslint and move linting TypeScript projects with eslint
# 🙋 Feature Request

### 😯 Current Behavior

Yoshi is using `tslint` in order to lint TypeScript projects.

### 🤔 Expected Behavior

Yoshi will use `eslint` in order to lint TypeScript projects.

### 🔦 Context

The [TSlint team announced on deprecating TSlint](https://medium.com/palantir/tslint-in-2019-1a144c2317a9) and move to eslint instead, not long ago, [create-react-app also added TypeScript eslint support](https://github.com/facebook/create-react-app/pull/6513) and you can look at it in order to take some inspiration.

I'm not sure if all the rules are supported, but I'm sure that it's the direction the community chose, and that one tool (instead of 2) will be better for the yoshi team and the users.

This change will need to be part of yoshi's next major version because It's hard to believe that we'll get to feature parity between tslint and eslint configs. There is currently a need for a static [`tslint.json`](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/client/typescript/tslint.json) file. We'll need to instruct devs to change it into an `eslint` equivalent.

eranshabi:
this fixes #1266

### DIFF
--- a/packages/yoshi-helpers/src/queries.ts
+++ b/packages/yoshi-helpers/src/queries.ts
@@ -47,6 +47,8 @@ export const shouldRunSass = () => {
 export const isTypescriptProject = () =>
   fs.existsSync(path.resolve('tsconfig.json'));
 
+export const isUsingTSLint = () => exists('tslint.*');
+
 export const shouldExportModule = () => {
   return !!config.pkgJson.module;
 };
--- a/packages/yoshi/src/commands/lint.js
+++ b/packages/yoshi/src/commands/lint.js
@@ -64,8 +64,8 @@ module.exports = runner.command(async () => {
     }
   } else if (isTypescriptProject()) {
     const tsFilesToLint =
-      shouldRunOnSpecificFiles && jsFiles.length
-        ? jsFiles
+      shouldRunOnSpecificFiles && (tsFiles.length || jsFiles.length)
+        ? [...tsFiles, ...jsFiles]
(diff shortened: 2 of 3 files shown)


ranyitz:
Hey @eranshabi, thanks for sending this PR. A few concerns,

The current changes will run `eslint` instead of `tslint`, but our `eslint` preset does not support linting TypeScript files yet.

I'm afraid that we are going to supply a feature that encourages our users to use a custom `eslint` preset. I prefer supporting `eslint` when we have a proper `eslint` preset that works on TypeScript files as well.

______________

If you can enrich `eslint-config-yoshi` with TypeScript support, we'll review it and merge. Hopefully, you'll be using it and we could test it on your project/s and iterate. On the next major version, we'll introduce it as official, we'll change it in the TypeScript templates and add it to the migration guide + deprecation warning on using `tslint`.

It's important to understand that after we add TypeScript rules to this config, it's going to be hard to change stuff (every new rule that we want to add will be a breaking change that we need to add in a major version) and that's why it makes sense to create a TypeScript eslint config and use it for some time before we declare it as official.

We are using `eslint` for our own codebase, see how we've added TypeScript support - https://github.com/wix/yoshi/blob/master/.eslintrc#L15-L39

PR #1506, opened 2019-08-21
Jest `testURL` config option override

### LINKED ISSUE #1499: Add an option to set testURL manually in Jest config
# 🙋 Feature Request

### 😯 Current Behavior

I tried to set the `testURL` parameter in the `jest` field in our project's package.json and the url didn't change.

### 🤔 Expected Behavior

The URL should change.

### 💁 Possible Solution 

Add a yoshi config option to set that URL, either as a string or even better as a result of a function.

### 🔦 Context 

I'm in the process of moving one of `chat`'s front end apps from mocha to Jest. Our tests use a given URL on a specific port as the base for all of the http mocking, which we used to set ourselves to our custom jsdom instance (since mocha doesn't provide one for us) and there were no problems with it.
Currently, though, there's no official way to change the global `jsdom` instance's URL (the instance that Jest uses) - see facebook/jest#5124. There are a bunch of solutions to this, the easiest to implement IMO would just be to provide a `testURL` option in the yoshi config, or a better one would be to use the custom [jest-environment-jsdom-global](https://www.npmjs.com/package/jest-environment-jsdom-global) which exposes the Jest jsdom instance globally for all of the tests.

jonohayon:
### 🔦 Summary
Fixes #1499. Added the option to override the Jest `testURL` config option in `jest-yoshi.config`.

### 🗺️ Test Plan
The `testURL` value is set in the JS kitchensink `jest-yoshi.config` file, and `window.location.href` is expected to match it.


jonohayon:
> I think that in the future we would like to have specific whitelist per environment. I'll open a dedicated issue for that.

I thought about that too, since the `testURL` option isn't relevant for e2e tests.

ranyitz:
@jonohayon https://github.com/wix/yoshi/issues/1507  issue opened. PRs welcome.

Issue #860, opened 2018-11-27
Unable to use `jest.mock` with Wallaby


ariel7342:
# 🐛 Bug Report

### 😯 Current Behavior

i'm getting failing tests when I use wallaby (there are no failing test in the ci or in the local terminal run)
Wallaby instrumentation is probably breaking the hoisting behaviour.
apparently the jest.mock is not hoisted in wallaby, therefore there are error which connect to the data

### 🤔 Expected Behavior

the wallaby and yoshi run suppose to console the same error, if there are no error in yoshi run there not suppose to be in the wallaby

### 🔦 Context 

when I develop I'm getting all the time errors that are not true, it distract me.  

### 💁 Possible Solution 

I looked up and found a solutions. the same issue is represents here : [link](https://github.com/wallabyjs/public/issues/872)
TL;DR
need to add to compilers :
```
...
compilers: {
    '**/*.js*': wallaby.compilers.babel({babelrc: true, presets: ['babel-preset-jest']})
},
```

### 💻 Code Sample

[git link to me repo](https://github.com/wix-private/ecom/tree/master/client/wixstores-client/wixstores-client-product-page)

### 🌍 Your Environment
```
System:
    OS: macOS 10.14.1
    CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  Binaries:
    Node: 8.9.1 - ~/.nvm/versions/node/v8.9.1/bin/node
    Yarn: Not Found
    npm: 5.5.1 - ~/.nvm/versions/node/v8.9.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 69.0.3497.100
    Firefox: 62.0.2
    Safari: 12.0.1
  npmPackages:
    @storybook/core: Not Found
    @storybook/react: Not Found
    babel-preset-yoshi:  3.17.0 
    eslint-config-yoshi:  3.21.0 
    eslint-config-yoshi-base:  3.21.0 
    jest-environment-yoshi-bootstrap:  3.20.0 
    jest-environment-yoshi-puppeteer:  3.20.0 
    jest-yoshi-preset:  3.20.0 
    storybook: Not Found
    tslint-config-yoshi:  3.21.0 
    tslint-config-yoshi-base:  3.21.0 
    typescript: ^2.9.2 => 2.9.2 
    webpack:  4.26.1 
    yoshi-config:  3.20.0 
    yoshi-helpers:  3.20.0 
[...]

PR #1509, opened 2019-08-26
chore(yoshi): support for jest.mock calls in TS

liorcode:
Added support for jest.mock calls in typescript, following the explanation in https://wallabyjs.com/docs/integration/jest.html#jest-with-typescript-and-jestmock-calls

This should solve #860


liorcode:
@ranyitz ?

PR #1510, opened 2019-08-26
Add stylable hook to protractor configuration

tomrav:
### 🔦 Summary
Whenever Stylable stylesheets are consumed in a node environment, the stylable require hook needs be inserted in order to provide handling for `.st.css` files.

Our hook will catch all `.css` files, handle Stylable files itself, and pass any other `.css` files to a previous hook (in this case, css modules).

**Warning:** It is possible that there are existing cases that are currently getting false results in protractor tests with Stylable (due to css modules somehow handling `.st.css` files), and that this change will brings those breakings to the surface.


tomrav:
Still relevant.

PR #1512, opened 2019-08-27
don't publish from monorepo release command

netanelgilad:
### 🔦 Summary
The current way that `release-monorepo` command works is that it published all the packages using `npm-ci`.
Since we want to remove `npm-ci` it's no longer going to be available in the CI, so `yoshi` shouldn't build on it.

After talks with @ronami we have decided that publishing will remain in the territory of the ci scripts, and they will just run `npm run release` on the root of the monorepo, where `yoshi` will bump the versions of the packages according to its own logic. 

So I have removed the publish part from `release-monorepo`.

### 🗺️ Test Plan
N/A


netanelgilad:
@yavorsky There is currently one repo that uses that feature, so I plan to release it in a minor/patch version, and fix that repo manually. So probably no one will see the warning :D

netanelgilad:
@ronami @ranyitz Let's please merge this. 🙏 🙏 🙏

PR #1513, opened 2019-08-27
transpile development bundle to last chrome version

netanelgilad:
### 🔦 Summary
Currently in development mode, we transpile the browser code to very old browsers even though everyone develops on the latest chrome (or other evergreen browser). This makes development harder. First it adds more transpilation than needed in development which is time wasted, and the more important thing it makes debugging much harder, cause the transpilation makes the breakpoints much harder to work with and it changes some of the variables names so it's hard to inspect them.
Transpiling to latest chrome should be our default for development builds.

@ranyitz - I showed these problems to @ronami and we saw that this solution works great (for my use case). But he also mentioned that this was tried in the past and reverted. Do you know something about it?

### 🗺️ Test Plan
Not sure, @ronami got a suggestion?

### DIFF
--- a/packages/babel-preset-yoshi/index.js
+++ b/packages/babel-preset-yoshi/index.js
@@ -41,7 +41,10 @@ module.exports = function(api, opts = {}) {
           // We don't need to be fully spec compatible, bundle size is more important.
           loose: true,
           // Allow users to provide its own targets and supply target node for test environment by default.
-          targets: options.targets || (isTest && 'current node'),
+          targets:
+            options.targets ||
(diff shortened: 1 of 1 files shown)


netanelgilad:
> Maybe we can give last 2 or 3 chrome versions? Not sure if it's ever going to affect someone, but it might😕

Agreed. let's go with 2?

yavorsky:
l would have something like `last 2 chrome versions, last 2 firefox versions`.
Anyway, evergreen browsers is updating much faster than caniuse-lite dependency :)

PR #1514, opened 2019-08-28
update tpa-style-webpack-plugin version

ronnyrin:
### 🔦 Summary
I've added new functionality to the plugin to support injecting static css to the bundle.

Question: Is it possible to update it also in yoshi 3?


ranyitz:
> I've added new functionality to the plugin to support injecting static css to the bundle.

Why this functionality is needed? 

> Question: Is it possible to update it also in yoshi 3?

We have a branch called `version_3.x`, when you create a PR to it we can release a new version with those changes.

ronnyrin:
We are going to put all the css (including the static css) inside the js bundle because we are having issues with loading the files asynchronously. It's a fix for https://jira.wixpress.com/browse/BOLT-1629

asaf-yonay:
@ranyitz - Any reason not to merge this considering Ronny's business justification?

Issue #1517, opened 2019-08-29
Preserve dynamic imports in EcmaScript Babel build


ronami:
### Summary

Currently, Yoshi can build a Webpack optimized build for libraries when the `module` prop is set (https://wix.github.io/yoshi/docs/guides/export-es-modules#docsNav). It preserves EcmaScript `import`/`export` syntax, instead of transpiling them to CommonJs, so Webpack can perform optimizations like tree-shaking and scope-hoisting.

To support using code-spitting via dynamic imports in libraries, we should also preserve dynamic imports in the Webpack specific (EcmaScript target) build too.

To create an EcmaScript build with Babel, Yoshi configures `babel-preset-yoshi` with `{ modules: false }` in https://github.com/wix/yoshi/blob/master/packages/yoshi/src/commands/build.js#L277. I think that `babel-preset-yoshi` shouldn't include `babel-plugin-dynamic-import-node` if `modules` are set to `false`, so dynamic imports won't be transpiled down (https://github.com/wix/yoshi/blob/master/packages/babel-preset-yoshi/index.js#L87).

PR #1519, opened 2019-08-29
keep dynamic `import()` if `modules: false` (e.g. ECMAScript build)

zemlanin:
### 🔦 Summary
Currently, Yoshi can build a Webpack optimized build for libraries when the `module` prop is set (https://wix.github.io/yoshi/docs/guides/export-es-modules#docsNav). It preserves EcmaScript `import`/`export` syntax, instead of transpiling them to CommonJs, so Webpack can perform optimizations like tree-shaking and scope-hoisting.

To support using code-spitting via dynamic imports in libraries, we should also preserve dynamic imports in the Webpack specific (EcmaScript target) build too.

closes https://github.com/wix/yoshi/issues/1517

### Context
https://wix.slack.com/archives/CAL591CDV/p1566990008004000

> I’m trying to dynamically import locale data in `@wix/wix-i18n-config` library, letting application’s `build` process create async chunks (instead of bundling them in when building the library) https://github.com/wix-private/fed-infra/pull/1300

### 🗺️ Test Plan
To test this changes, I've build a branch of [`wix-i18n-config`](https://github.com/wix-private/fed-infra/tree/bodia-uz-patch-1/wix-i18n-config) and `master` of [`search-results-widget`](https://github.com/wix-private/site-search)

- `import()`s were compiled away in library's `main` (`fed-infra/wix-i18n-config/dist/src/i18n.js`) target
  Screen Shot 2019-08-29 at 18 44 20

- kept in library's `module` target (`fed-infra/wix-i18n-config/dist/es/src/i18n.js`)
  Screen Shot 2019-08-29 at 18 44 26

- split into async chunks in application's target
  Screen Shot 2019-08-29 at 18 43 54


amitdahan:
Thanks!

PR #1522, opened 2019-09-02
added babel preset typescript

amiryonatan:
### 🔦 Summary

Added babel preset typescript to babel config, so babel can transpile TS.
In our project we use it like that:
```
transform: {
    '\\.tsx?$': 'babel-jest',
  },
```
in our jest config, because we use ts compiler and babel-jest instead of ts-jest


amiryonatan:
@ronami as discussed

ranyitz:
@amiryonatan If I understand your problem correctly, you would like to reduce testing time by removing TypeScript's type check.

It's can be done via disabling `diagnostics` option.

https://huafu.github.io/ts-jest/user/config/diagnostics#disablingenabling

_____________________________

We want to consider adding TypeScript support for babel, but we want to do it carefully and verify that it is part of a yoshi flow, we also not sure how it will affect others so if the above solution works for you, it would be great.

amiryonatan:
@ranyitz thanks for the suggestion!
I ran all 3 possibilities and wanted to share the results with you, just so you'll see the difference in performance:

1. babel-jest: 98.5 seconds
2. ts-jest, without diagnostics: 158 seconds
3. ts-jest, with diagnostics: 278 seconds

850 tests, all ran in CI.

PR #1020, opened 2019-02-25
Change app-flow tests to support testing Yoshi's testing infra

ronami:
### 🔦 Summary

This PR helps adjust the new integration tests to also test our test setup (`yoshi test` and `jest-yoshi-preset`).

Instead of running arbitrary Jest configs, every kitchensink project will be tested like the following:

#### Test for production (simulating CI environment):
- Build the project for production (`yoshi build`).
- Run the project's own tests in a CI environment against the build results (`yoshi test`).
- Run production integration tests against the build result.

#### Test for local development (simulating local dev environment):
- Start the app for development (`yoshi start`).
- Run the projects own tests against the running local dev setup (`yoshi test`).
- Run development integration tests against the running development setup.

#### Test additional scenarios
- Run additional integration tests (errors, analyze).

Each project should cover every way a developer can use Yoshi: By importing and using different types of assets, using the capabilities of its testing preset or others.

PR #1523, opened 2019-09-03
Update version-4.md

justame:
### 🔦 Summary
...

### 🗺️ Test Plan
...

### DIFF
--- a/docs/migration/version-4.md
+++ b/docs/migration/version-4.md
@@ -103,6 +103,13 @@ const logo = require("./logo.svg");
 const App = () => <img src={logo} />;
 ```
 
+### Assets import in scss files
+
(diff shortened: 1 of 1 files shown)


ronami:
Thanks for the PR @justame, but we found that using `~` is to resolve things from `node_modules` (https://github.com/webpack-contrib/css-loader#url).

The bug we were debugging happened since #1020 because we don't have unique asset names (https://github.com/wix/yoshi/blob/master/packages/yoshi/config/webpack.config.js#L77). If two assets in two different directories have the same name, they will override each other. I'll open a separate issue/PR for it so we can fix it quickly.

ronami:
See #1524.

Issue #1427, opened 2019-07-01
Add `@types/puppeteer` to the dependencies, so users will get them.


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior

Users do not get `@types/puppeteer` when using `jest-yoshi-preset` and therefore have type `any` on things like `Page` and `Browser`.

### 🤔 Expected Behavior

The users will get the correct types for `Page` and `Browser`.

### 💁 Possible Solution

Add them as a dependency to `jest-yoshi-preset`

* Understanding that this is a breaking change we will release it in the next major version.

PR #1554, opened 2019-09-16
CYA - Client stricter config

amitdahan:
### 🔦 Summary
Minor `tsconfig.json` tweaks for the client (TS) template.
**Note:** Making this template stricter uncovered the lack of `@types/puppeteer` in generated client projects.
This also adds `@types/puppeteer` to `jest-yoshi-preset` since it plants `page: Puppeteer.Page` in the global. 
The addition of type declarations for `yoshi-template-intro` is temporary until we can get https://github.com/wix/yoshi/pull/1487 merged.

### 🗺️ Test Plan
Template passed ✅


amitdahan:
I also wasn't so sure about `noEmit`, wdyt?

ronami:
> I also wasn't so sure about `noEmit`, wdyt?

I like it, if a developer checks for type errors by running `npx tsc` directly they don't need to remember passing `--noEmit`

packages/create-yoshi-app/templates/client/typescript/src/external-types.d.ts
```diff
@@ -2,3 +2,5 @@ interface Window {
   __LOCALE__: string;
   __BASEURL__: string;
 }
+
+declare module 'yoshi-template-intro';
```
ranyitz: `yoshi-template-intro` is something that our users should remove right after they are generating the project. Now they also need to remove it from here.

Maybe we can think of another solution that wouldn't add another file to the templates?

amitdahan: Yes, I forgot to mention, this is temporary (`noImplicitAny` forced this) until we can get https://github.com/wix/yoshi/pull/1487 merged and types will be shipped correctly so it's not needed. Updating the PR accordingly.

amitdahan:
Thanks! 🙏
Glad these tests were here to catch this!

amitdahan:
Thanks! Will definitely be on alert 😀

amitdahan:
Closes https://github.com/wix/yoshi/issues/1427 :)

PR #1385, opened 2019-06-17
Top commands / single Yoshi dependency for monorepos

ronami:
### 🔦 Summary

This PR changes how the `build` and `start` commands work for monorepos. Instead of spawning multiple `yoshi` commands, it builds the needed apps and libraries in a single flow.

I refactored `yoshi-config/paths`, `yoshi-config/loadConfig` and similar functions to accept `cwd`s other than `process.cwd()`.

I introduced the concept of `app`/`lib` to describe the information about packages in a monorepo. I used Lerna's APIs to get those packages.

PR #1555, opened 2019-09-16
Top commandד

ranyitz:
### 🔦 Summary
This PR is a continuation of #1385.

> This PR changes how the build and start commands work for monorepos. Instead of spawning multiple yoshi commands, it builds the needed apps and libraries in a single flow.
> 
> I refactored yoshi-config/paths, yoshi-config/loadConfig and similar functions to accept cwds other than process.cwd().
> 
> I introduced the concept of app/lib to describe the information about packages in a monorepo. I used Lerna's APIs to get those packages.

The reasoning of this PR is to make Yoshi understand what `monorepo` is and let it reuse Yoshi's flows in packages of the monorepo.

Essentially it will let you run `yoshi` commands from the root of the monorepo, and yoshi will build all of the libraries/apps for you. (instead of spawning multiple yoshis in each project). Exactly like @ronami explained in his original PR.

_____________________

### Some thoughts

Yoshi used to have a `config` and `paths` modules. Before the monorepo era both would be relative to `process.cwd()`.

When we talk about monorepo, we talk about multiple `apps`/`libraries` that are being placed in the same git repo, the reason for that, is mainly because those packages are dependent on one another and should be developed/tested/published together.

### Why not to spawn multiple yoshis?

One option would be to have multiple `yoshi` instances and just orchestrate them to run together on different processes.

__This solution has some problems:__

1. A package in a monorepo would not necessarily have the same characteristics as a standalone package. (There are going to be things you want to configure from the root of the monorepo but not from each package)
2. It won't let us make some important optimization. For example, we would like to have a single webpack watcher/compiler instead of let's say 6 of them, we would like to serve many applications on the same dev server using a simple namespacing convention. Those things would be very hard/impossible to do with multiple dev server instances.
3. We could leverage features like [typescript-project-refernces](https://www.typescriptlang.org/docs/handbook/project-references.html) which created due to similar reasons (instead of running multiple `tsc`).
4. We can also see the same thing on `jest`'s [projects](https://jestjs.io/docs/en/configuration#projects-array-string-projectconfig) configuration options which gave jest the ability to create a coherent testing process and sane output for multiple projects running at the same time. (This wouldn't be possible if you wanted to spawn multiple `jest`s)

### So what do you suggest?

**Let's make sure that we have a common language:**

* **Repo** - A git repository.
* **Package/project** - A directory that contains a module that is being published somewhere.
* **polyrepo** - A git repository that contains a single package. (polyrepo generally refers to the concept of having multiple git repositories for multiple packages but we'll use it as the opposite of monorepo, [read more](https://github.com/joelparkerhenderson/monorepo_vs_polyrepo#what-is-polyrepo))
* **Monorepo** - A git repository that contains multiple packages/projects.

Yoshi should know whether it's running a monorepo or a polyrepo. We already have development flows for a polyrepo (conceptually anyway):
* **App** - Deployable user-facing application.
* **Library** - Publishable npm module that is being installed in other libraries/applications.
* **Server** - Deployable node server that exposes API for others server/fullstack apps.

And more platform specific types, like `out-of-iframe` & `business-manager` flows.

We believe that packages in a monorepo should be from the same project types.
It will help us with code-reuse, but most importantly, our users would not need to learn a new concept. It's just the same old flows just tailored together to maximize performance and development velocity.

While this PR scope deals with a single `app` and multiple libraries (which happens to be a pretty common case), hopefully, it will place the infrastructure for a multi-app monorepo and one with servers + platform-specific flows.

The sky is the limit.

### 🗺️ Test Plan
All tests are passing


docs/getting-started/project-types.md
```diff
@@ -29,8 +29,8 @@ A client project that integrate with business manager APIs.
 ## Out Of Iframe
 
 An application that has it's view components written in React and run in the viewer.
-The application business logic runs inside the platform worker. 
-For more information head to the [official docs](https://bo.wix.com/wix-docs/client/client-frameworks#viewer-platform-(ooi)).
+The application business logic runs inside the platform worker.
+For more information head to the [official docs](<https://bo.wix.com/wix-docs/client/client-frameworks#viewer-platform-(ooi)>).
```
ronami: Can undo

docs/guides/stylelint-setup.md
```diff
@@ -41,4 +41,4 @@ Add the following to your `package.json`:
 
 Finally, configure your `package.json` to run Stylelint after `yoshi lint`:
 
-```yoshi lint && stylelint```
+`yoshi lint && stylelint`
```
ronami: Not relevant to PR

docs/migration/version-3.md
```diff
@@ -5,4 +5,3 @@ sidebar_label: Version 3
 ---
 
 The guide was originally posted on the [Yoshi site](https://wix.github.io/yoshi/), please read it [there](https://wix.github.io/yoshi/blog/2018/06/3/yoshi-3#migration-guide).
-
```
ronami: Not relevant to PR

packages/yoshi-config/src/config.ts
```diff
@@ -110,9 +110,23 @@ export type Config = {
...
   experimentalMinimalPRBuild: boolean;
   experimentalRtlCss: boolean;
-  experimentalMonorepoSubProcess: boolean;
   yoshiServer: boolean;
   projectType: ProjectType | null;
   webWorkerEntry?: WebpackEntry;
   webWorkerExternals?: WebpackExternals;
 };
+
+// export type ProjectInAMonorepo = {
```
ronami: Remove and add in a later PR?

packages/yoshi/config/webpack.config.js
```diff
@@ -74,40 +65,40 @@ const isTypescriptProject = checkIsTypescriptProject();
...
-    ? inTeamCity || isProduction
-    : project.separateCss;
+const computedSeparateCss = app =>
+  app.separateCss === 'prod' ? inTeamCity || isProduction : app.separateCss;
 
 const artifactVersion = process.env.ARTIFACT_VERSION;
 
 const staticAssetName = addHashToAssetName('media/[name].[ext]', 'hash:8');
 
+const { defaultEntry } = require('yoshi-helpers/constants');
```
ronami: Can be hoisted to top

packages/yoshi/src/webpack-manager.js
```diff
@@ -0,0 +1,87 @@
...
+    }, []);
+
+    const webpackStats = await this.runWebpack(webpackConfigs);
+
+    return {
+      getAppData: app => {
+        let appConfigStartIndex = 0;
+        let appConfigEndIndex = 0;
+
+        for (let i = 0; i < this.apps.length; i++) {
```
ronami: A small comment can help explain what's going on

PR #1556, opened 2019-09-17
Align BM template react versions with BM

amitdahan:
### 🔦 Summary
Bumps BM templates to BM's currently used version.

### 🗺️ Test Plan
Template tests


amitdahan:
On it

amitdahan:
@ronami Green

Issue #1324, opened 2019-05-23
Confusing error message when using app flow without a server.js


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

During the [migration to the App Flow](https://wix.github.io/yoshi/docs/guides/app-flow) I had no server located in `server.js` or `dev/server.js`. I was migrating a client project and inserted a wrong file name for the server. (`sever.js`)

That was the error that I've seen:
 
![image](https://user-images.githubusercontent.com/11733036/58263648-13ee0880-7d85-11e9-9ad8-1beb7afebbd6.png)

### 🤔 Expected Behavior + 💁 Possible Solution

If there is no server in one of the expected location yoshi should present a good error message suggesting to create a `server.js` or a `dev/server.js` file in order for it to work.

### 🔦 Context

During migration or just when moving files, it's important that error messages will be indicative and will help others to waste less time on trying to debug why things happen.

### 👣 Steps to Reproduce

1. Create a fullstack / client project
2. change the name of the `server.js` file to something else.
3. run `npm start`

### 🌍 Your Environment

`yoshi 4.6.1`

ranyitz:
Fixed on #1561

PR #1561, opened 2019-09-18
explicit error on missing server entry

deanshub:
### 🔦 Summary
Continuing the great work of @pycm in regards of #1372 and #1324 
This will give an explicit error on missing server entry

### 🗺️ Test Plan
app flow tests changed to check if the error received when no server entry found


packages/yoshi/config/webpack.config.js
```diff
@@ -848,17 +848,30 @@ function createServerWebpackConfig({
...
-        ...{
-          // If `yoshi-server` isn't being used we should show an error if no server entry exists
-          ...(serverEntry || !app.yoshiServer
-            ? { server: serverEntry || possibleServerEntries[0] }
-            : {}),
-        },
-      };
+      if (!serverEntry && !app.yoshiServer) {
+        throw new Error(
+          `Can't find server entry.
```
yanivefraim: I would try to add a more descriptive message here (as a user, I'm not sure what to do with this error). Something like:

```
We couldn't find your server entry. You should either:
1. Supply it explicitly using the "--server" option
2. Use one of the defaults:
    - "src/server": for a fullstack project
    - "dev/server": for a client only project

Please refer to 
https://wix.github.io/yoshi/docs/api/cli#server 
and
https://wix.github.io/yoshi/docs/guides/app-flow#migration-for-client-apps for more details
```

packages/yoshi/config/webpack.config.js
```diff
@@ -848,17 +848,30 @@ function createServerWebpackConfig({
...
-          // If `yoshi-server` isn't being used we should show an error if no server entry exists
-          ...(serverEntry || !app.yoshiServer
-            ? { server: serverEntry || possibleServerEntries[0] }
-            : {}),
-        },
-      };
+      if (!serverEntry && !app.yoshiServer) {
+        throw new Error(
+          `We couldn't find your server entry. You should either:
+          1. Supply it explicitly using the "--server" option
```
yanivefraim: So, my mistake, option 1 does not actually exist...
Let's change it to be something like:

`We couldn't find your server entry. Please use one of the defaults....`

packages/yoshi/config/webpack.config.js
```diff
@@ -848,17 +848,30 @@ function createServerWebpackConfig({
...
-        ...(app.yoshiServer && createServerEntries(config.context, app)),
-
-        ...{
-          // If `yoshi-server` isn't being used we should show an error if no server entry exists
-          ...(serverEntry || !app.yoshiServer
-            ? { server: serverEntry || possibleServerEntries[0] }
-            : {}),
-        },
-      };
+      if (!serverEntry && !app.yoshiServer) {
```
yanivefraim: `webpack.config.js` file became too large, and this thing makes it larger...

Later on, we can move the validation to a validate phase, but for now, let's just move the validation to a `validateEntry` method, in `webpack-utils`:

```
const serverEntry = validateEntry(...
```

if there is a problem, `validateEntry` will throw the error. Otherwise, it will return the entry.

WDYT?

packages/yoshi/config/webpack.config.js
```diff
@@ -847,18 +838,17 @@ function createServerWebpackConfig({
...
 
     entry: async () => {
-      const serverEntry = possibleServerEntries.find(exists(app));
+      const serverEntry = validateServerEntry(app, extensions);
 
-      return {
-        ...(app.yoshiServer && createServerEntries(config.context, app)),
+      let entryConfig = {};
+      if (app.yoshiServer) {
+        entryConfig = { ...createServerEntries(config.context, app) };
```
ronami: the `{ ... }` isn't needed if we changed it to `if()` checks instead of a ternary.

packages/yoshi/config/webpack.config.js
```diff
@@ -847,18 +838,17 @@ function createServerWebpackConfig({
...
+      }
+      if (serverEntry) {
```
ronami: we don't need to check it it exists

packages/yoshi/config/webpack.config.js
yavorsky: this call will be redundant if `serverEntry` is not falsy. [Next condition](https://github.com/wix/yoshi/pull/1561/files#diff-b91ad4512169022152dc9db9e5f4abcaR847) will just override it.
Can we have smth like:

```js
if (serverEntry) {
  entryConfig = { ...entryConfig, server: serverEntry };
} else if (app.yoshiServer) {
  entryConfig = createServerEntries(config.context, app);
}
```

PR #942, opened 2019-01-28
Support importing unknown assets

matveyok:
### 🔦 Summary
Fallback to url-upload for unknown files
Related Issue - #821

Issue #821, opened 2018-11-19
Support importing unknonwn file extensions


ronami:
### Summary

Currently, we explicitly support a limited set of extensions with `url-loader`. (https://github.com/wix/yoshi/blob/master/packages/yoshi/config/webpack.config.js#L421). We also test each of those extensions separately with the same test (https://github.com/wix/yoshi/blob/master/packages/yoshi/test/loaders.spec.js#L265). And every now and then when someone needs to import an extension we don't support they have to add a pull request with the extension and duplicating another one of those tests.

Instead of all that, we can have `url-loader` by the default loader and catch all unknown file extensions. See https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L498 and https://github.com/kriasoft/react-starter-kit/blob/master/tools/webpack.config.js#L245 as examples.

PR #1564, opened 2019-09-19
Support importing unknown assets

deanshub:
### 🔦 Summary
in reference to #821 and #942
this feature will allow users to require any file,
the files with the unknown extensions will be caught by the [url loader](https://github.com/wix/yoshi/compare/support-unknown-extensions?expand=1#diff-b91ad4512169022152dc9db9e5f4abcaR562)

### 🗺️ Test Plan
importing unknown file extensions and testing if they got into the app using kitchensink tests


packages/yoshi/config/webpack.config.js
```diff
@@ -546,17 +546,29 @@ function createCommonWebpackConfig({
...
+                reStyle,
+              ],
               options: {
                 name: staticAssetName,
                 limit: 10000,
               },
             },
           ],
         },
+        // ** STOP ** Are you adding a new loader?
```
ranyitz: This is great!

test/projects/kitchensink-javascript/src/components/features/unknown-inclusion.spec.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import UnknownInclusion from './unknown-inclusion';
+
+it('should pass', () => {
+  const div = document.createElement('div');
+  ReactDOM.render(<UnknownInclusion />, div);
```
ranyitz: Maybe I'm missing something, but in this case, this is not an `unknown` but rather a specific extension added to our jest preset's list of known extensions (`webp`)

deanshub: I needed a file extension that we have inside our preset so I won't have to eject from jest-yoshi-preset so tests will pass, but still isn't in our webpack config extensions for url-loader
any thoughts around that?

deanshub:
> The support for unknown extensions is currently added only to the webpack loader, but not to the test environment.
My thoughts around that is that we should still get issues\PRs for common extensions but for the rest of the extensions, the specific project can and should add a mapper\transformer of their own

test/projects/kitchensink-javascript/integration/loaders.test.js
```diff
@@ -324,4 +324,16 @@ describe('webpack', () => {
       expect(result).toBe('This is an abstract.');
     });
   });
+
+  describe('unknown', () => {
```
ranyitz: This is the same test as the image inclusion test, here we specifically testing `webp` but it is the same as any other extension. I would prefer removing this test.

Issue #1219, opened 2019-04-05
[create-yoshi-app] [templates dependency update] i18next


sidoruk-sv:
Please update templates with new i18next and i18next-rect and best way how to deal with it in tests (need React.Suspense wrapper, and some changes to i18next mock)

mastertheblaster:
We should use `wix-i18n-config` cause it brings all required dependencies and also enables `i18next-icu`. @MSiliunas wanna make a PR for this? Or at least reuse the existing one?

sidoruk-sv:
Full-stack templates changed here: #1929

sidoruk-sv:
Client templates changed here: #1961

ranyitz:
So we want to update the version of `18next` and `react-18next` and we thought it would be best to use https://github.com/wix-private/fed-infra/tree/master/wix-i18n-config which already includes the latest versions. This may be a bit more complex for `out-of-iframe` case when the react version is still below `16.8.0`

@vybu @amitdahan @yavorsky

sidoruk-sv:
That's why I stopped on client && fullstack, as more stable and controlled things)

sidoruk-sv:
`business-manager-module` case will be covered by #2363

PR #1570, opened 2019-09-22
CYA - Fullstack stricter tsconfig

amitdahan:
### 🔦 Summary
Minor `tsconfig.json` tweaks for the Fullstack (TS) template.

### 🗺️ Test Plan
Template tests


packages/create-yoshi-app/templates/fullstack/typescript/src/i18n.ts
```diff
@@ -1,10 +1,17 @@
...
       type: 'backend',
-      read: (language, namespace, callback) => {
+      read: (
+        language: string,
+        namespace: string,
+        callback: (
+          err: any | null,
+          translations?: Record<string, string>,
+        ) => void,
+      ) => {
```
ronami: The newer versions of `i18next` and `i18next-react` already include types that are a lot more specific. I think we should try to bump them when we can. Referencing the related issue: https://github.com/wix/yoshi/issues/1219.

packages/create-yoshi-app/templates/fullstack/typescript/src/external-types.d.ts
```diff
@@ -2,3 +2,14 @@ interface Window {
...
+// tslint:disable-next-line:no-namespace
+declare namespace Express {
+  interface Request {
+    aspects: any;
+  }
+}
+
+declare module 'yoshi-template-intro';
+declare module '@wix/wix-express-require-https';
+declare module '@wix/wix-express-csrf';
```
ronami: I would really like to avoid this but I understand it will take a bit of time until we can fully use node-platform types. Also referencing https://github.com/wix/yoshi/pull/1487 for ``yoshi-template-intro.

ronami: Also, I think a small comment that explains why those are here would be less confusing for developers that get this code generated. Wdyt?

PR #1571, opened 2019-09-22
adding watch mode to app flow tests

deanshub:
### 🔦 Summary
app flows should be the main way we test yoshi since its e2e,
but the way we test it is using a single run of copy, build, test.
This PR introduces a watch mode that copies files from the kitchensink templates when they change then uses jest watch mode to test them.


CONTRIBUTING.md
```diff
@@ -70,6 +70,30 @@ VERBOSE_TESTS=true npm test
...
+which the npm script uses
+
+```shell
+npm run test:app-flow:js
+```
+
+Or using cli arguments (watch mode only works with cli args)
+
+```shell
+npm run test:app-flow -- --FILTER_PROJECT=kitchensink-javascript --watch
```
yanivefraim: nit: I guess we should stick with either env vars or command args... WDYT?

scripts/utils/testProject.js
```diff
@@ -123,10 +134,17 @@ module.exports = async ({
     }
   }
 
-  await testProductionBuild();
-  await testLocalDevelopment();
-  await runAdditionalTests();
-
+  if (watchMode) {
```
yanivefraim: nit: why not just pass `watchMode` as the arg? (:

CONTRIBUTING.md
deanshub: thought the CI scripts added this env var, but it seems that we're the only ones adding it, I refactored to use only args

scripts/utils/testProject.js
```diff
@@ -123,12 +135,27 @@ module.exports = async ({
...
+
+  if (watchMode) {
+    // watch over tests and re run them
+    chokidar
+      .watch(`**/*${watchMode}*`, {
+        persistent: true,
+        ignoreInitial: true,
+        cwd: templateDirectory,
+      })
+      .on('all', (event, filePath) => runTests(filePath));
```
yanivefraim: - `runTests` expects an array, so in this case after first change you will have a watch for something like: `/s|r|c|c|o|m....`
- `filePath` is the changed file full path. This means that if you just watched for `svg-inclusion.spec.js`, which result in 3 relevant files,  now you are watching only `src/components/features/svg-inclusion.spec.js`. 

I guess we should change this line to:
`.on('all', (event, filePath) => runTests(filePath));`

scripts/utils/testProject.js
```diff
@@ -123,12 +135,27 @@ module.exports = async ({
...
+  }
 
-  // Clean eventually
-  await fs.remove(rootDirectory);
+  await runTests(typeof watchMode === 'string' ? [watchMode] : undefined);
+
+  if (watchMode) {
+    // watch over tests and re run them
+    chokidar
+      .watch(`**/*${watchMode}*`, {
```
yanivefraim: I think that we want to run only selected pattern, but watch everything. It is not very useful to watch only your test file / pattern.

CONTRIBUTING.md
```diff
@@ -70,6 +70,34 @@ VERBOSE_TESTS=true npm test
...
+Example of npm script usage
+
+```shell
+npm run test:app-flow:js
+```
+
+Example of flag filtering
+
+```shell
+npm run test:app-flow -- --filterProject=kitchensink-javascript --watch=svg-inclusion.spec
```
yanivefraim: The api is confusing (IMO). It is not clear that `watch` is the pattern to run tests for. I would try somehow to separate it to:
`--watch`
and 
`--pattern=XXX`

@ronami - WDYT?

test/scripts.js
```diff
@@ -74,8 +74,8 @@ module.exports = class Scripts {
     });
   }
 
-  async test(env = {}) {
-    return execa('npx', ['yoshi', 'test'], {
+  async test(env = {}, args = []) {
+    return execa('npx', ['yoshi', 'test', ...args], {
```
yanivefraim: This one will also fail if the pattern does not match. We should use some `--passWithNoTests` as we had in the other places

deanshub:
I agree, let's re-think the strategy of this feature

deanshub:
After our last discussion I moved back to jest watch mode and added a build mode argument which is required when using watch mode.

This will allow us to keep the terminal clear and still use jest's watch mode dependency watch features.

PR #1573, opened 2019-09-24
Jest preset asset transforms

### LINKED ISSUE #1568: Jest not transforming assets in node_modules
# 🐛 Bug Report

### 😯 Current Behavior
3rd party (`node_modules`) code that imports assets such as `.mp3` files (or images and so on) aren't transformed using our "fake" transformer.
Right now a user is experiencing an error (Some unexpected token, because the `.mp3` file is being parsed as a script) when he's importing (indirectly) an `.mp3` file from `node_modules`.

### 🤔 Expected Behavior
All webpack-handled assets should also work in (jest) tests.

### 💁 Possible Solution
Exclude supported extensions from `transformIgnorePatterns` in `jest-yoshi-preset`.

### 💻 Code Sample
https://github.com/wix-private/inbox/blob/687497927a70f806592cc14b8e3f646de272d136/inbox-keyboards/src/request-payment/components/RequestPaymentItemChatApp/RequestPaymentItemChatApp.driver.tsx#L11

### 🌍 Your Environment
Yoshi `4.16.2`

amitdahan:
### 🔦 Summary
Closes #1568.

### 🗺️ Test Plan
A test was added to each of the app-flow kitchensink tests.
The test imports a component from a fake node module that exports a component which imports an image.
This used to fail because of the following line (before the change):
https://github.com/wix/yoshi/pull/1573/files#diff-bc2cbaacdf4441b18314f3cec72419a4L136


amitdahan:
I added all types supported by jest (webpack will soon technically support "all" types, but using the common set sounds right) to our types exposed by yoshi, since those would fail TS projects importing some (even local) less-common imports like `.mp3`.

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -128,12 +128,29 @@ const config = {
           setupTestsFile,
         ].filter(Boolean);
 
+        const assetExtensions = [
```
ronami: I think the variable name should explain its content, how about having a `staticAssetsExtensions` which would be an array of strings and another variable that would pipe them before adding it into a regex?

packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -128,12 +128,29 @@ const config = {
...
+          'md',
+        ].join('|');
+
         return {
           ...project,
           modulePathIgnorePatterns,
 
           transformIgnorePatterns: [
-            '/node_modules/(?!(.*?\\.(st\\.css|svg)$))',
+            `/node_modules/(?!(.*?\\.(st\\.css|${assetExtensions})$))`,
```
ronami: I think we should (not necessarily as part of this PR) add a comment explaining what this (confusing) regex does.

packages/jest-yoshi-preset/jest-preset.js
amitdahan: Agreed, this was a small struggle point, couldn't figure out what I liked the most here.
Will make the changes.

Issue #1569, opened 2019-09-22
create-yoshi-app: A case when user's node version is different than the template's node version


amitdahan:
**EDIT: Issue specified here** - https://github.com/wix/yoshi/issues/1569#issuecomment-534976909

# ~~🙋 Feature Request~~
Node 12 is nearing LTS, I think CYA templates should start using node 12 as soon as possible.
A small problem might be that we currently don't switch node versions before testing the templates, so the fact that our templates use the same version of node as `yoshi` does - makes sure we _properly_ test our templates, but if we diverge from `yoshi`'s node version - this'll no longer be the case.
I thought about adding `nvm install && nvm use` before `npm install`ing, and since we only run template e2e's (properly) on CI - we can make the assumption that `nvm` exists.

### 😯 Current Behavior
They use `10`.

### 🤔 Expected Behavior
They'd use `12`.

amitdahan:
After rethinking this, we decided it still makes sense to purposefully run CI tests on the same version as yoshi, since we're making sure the templates' node version can run yoshi.

While we're at it, we did discover an actual bug:
1. User uses (or defaults to) a different version of node than the one used in the templates (e.g. `12`).
2. User generates any template (bug is more "felt" in client templates which use `node-sass`, which is built for the specific node version it's installed with).
3. The next time the user `nvm use`s inside his newly created project - he'll have to `npm rebuild` some dependencies, or reinstall everything after `nvm use`ing.

The way we'll solve this is by adding a check that'll make sure the node version used right before `npm install`ing, is semver-compatible with the one inside the template's `.nvmrc`.
If it's not - we'll stop before installing, and let the user know he doesn't have to start over, just use a compatible node version and `npm install` again.
We'll continue to install dependencies if it's compatible.

PR #1574, opened 2019-09-24
Run nvm use on create-yoshi-app template generation

amitdahan:
### 🔦 Summary
Run `nvm install && nvm use` before `npm install`ing after a new `create-yoshi-app` template was used.

### 🗺️ Test Plan
Template tests should still be green since they were using `yoshi`'s node version which happens to be the same as the templates (right now), but if we try to upgrade ungracefully this will now fail our build.

#1569


amitdahan:
We decided this isn't the way to go.
We'll do this instead - https://github.com/wix/yoshi/pull/1577.

amitdahan:
We decided this isn't the way to go.
We'll do this instead - https://github.com/wix/yoshi/pull/1577.

PR #1577, opened 2019-09-25
CYA - Skip npm install if current node isn't compatible with .nvmrc file

### LINKED ISSUE #1569: create-yoshi-app: A case when user's node version is different than the template's node version

amitdahan:
### 🔦 Summary
Closes #1569.

### 🗺️ Test Plan
Added some unit tests, plus existing tests keep passing *because* we're using the same original version that is used for testing yoshi which so happens to match the `.nvmrc` files in the templates.

### 🔮 Preview
image


amitdahan:
I'm closing this PR since we eventually decided on a different approach.

PR #1583, opened 2019-10-03
Fix BM template by emitting TS src files again

### LINKED ISSUE #1582: `npm start` fails to run local server in BM template
# 🐛 Bug Report

### 😯 Current Behavior

```
npx create-yoshi-app foobar
cd foobar
npm start
```

fails to connect to localhost

changing `tsconfig.json` from `noEmit: true` to `noEmit: false` solves the problem (`__tests__/dev/server.js` is not copied to `dist` folder)

...

### 🤔 Expected Behavior

the above scenario should just work
...

### 💁 Possible Solution 

This change was introduced [here](https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/business-manager-module/typescript/tsconfig.json#L13). I don't know if this should be reverted or maybe the `start` script in `package.json` should be updated...
...

amitdahan:
### 🔦 Summary
Closes #1582 

### 🗺️ Test Plan
Existing template tests

### DIFF
--- a/packages/create-yoshi-app/templates/business-manager-module/typescript/tsconfig.json
+++ b/packages/create-yoshi-app/templates/business-manager-module/typescript/tsconfig.json
@@ -9,8 +9,7 @@
     "allowJs": true,
     "jsx": "react",
     "lib": ["dom", "dom.iterable", "esnext"],
-    "resolveJsonModule": true,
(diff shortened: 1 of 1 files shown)


amitdahan:
Ping

PR #1591, opened 2019-10-06
Add support for Svelte as a Webpack loader

ronami:
### 🔦 Summary

This PR adds support for working with [Svelte](https://github.com/sveltejs/svelte). Yoshi won't install `svelte` as a dependency so npm install doesn't get any longer than it currently is (at least until it is more widely used). To use [Svelte](https://github.com/sveltejs/svelte), a project will first have to install it and its Webpack loader (`npm i svelte svelte-loader`).

### 🗺️ Test Plan

Since this is an opt-in feature that isn't widely used I think we should wait before adding e2e tests to test it. Also, I want to be sure we don't break projects that don't have the required opt-in dependencies installed until we fully support this feature.


packages/yoshi/config/webpack.config.js
```diff
@@ -54,7 +55,7 @@ const reScript = /\.js?$/;
 const reStyle = /\.(css|less|scss|sass)$/;
 const reAssets = /\.(png|jpg|jpeg|gif|woff|woff2|ttf|otf|eot|wav|mp3)$/;
 
-const extensions = ['.js', '.jsx', '.ts', '.tsx', '.json'];
+const extensions = ['.mjs', '.js', '.jsx', '.ts', '.tsx', '.svelte', '.json'];
```
ronami: This adds `mjs` to the possible extensions as needed by `svelte-loader`. This is an already popular extension that's also available in [CRA](https://github.com/facebook/create-react-app/blob/e7a2d6168a3a93ea58bafdc851eb2bbda20d5ce6/packages/react-scripts/config/paths.js#L49) and next.js.

PR #1593, opened 2019-10-07
Migrate BM templates to App Flow

amitdahan:
### 🔦 Summary
...


packages/create-yoshi-app/templates/business-manager-module/typescript/tsconfig.json
```diff
@@ -13,13 +13,13 @@
   },
   "include": [
     "./__tests__/**/*.ts",
-    "./jest-yoshi.config.js",
```
amitdahan: @ronami 
Was this included on purpose? This has to be a regular javascript file right? We don't transform it, why TS it then?
(This caused a type error for some weird unknown reason)

ranyitz: This was added for a reason, maybe check why it's failing?
https://github.com/wix/yoshi/issues/1091

amitdahan: Brought it back, it seems like a weird `ts-loader` behavior, if we don't set `noEmit` explicitly, `ts-loader` surprisingly does not do it, which in turn causes TS to throw because it's trying to emit `jest-yoshi.config.js` to the same path of the source file, which fails `yoshi build`.

amitdahan: It seems `ts-loader` won't make the change, reasons: https://github.com/TypeStrong/ts-loader/issues/719#issuecomment-363524508

amitdahan: This error can be "fixed" by making a similar change to our `webpack.config.js`:
https://github.com/TypeStrong/ts-loader/issues/161#issuecomment-188906356

Issue #1462, opened 2019-07-17
TypeScript migration tracking issue


ronami:
### Summary

This goal of this issue is to help us track which packages have migrated to TypeScript. We would love to get help from the community with the migration.

This PR is a continuation of #1401 which describes why we decided to migrate.

#### List of packages

|Package|Owner|PR|Status|
|-------|-----|--|------|
|~`babel-plugin-transform-hmr-runtime`~|||~Not relevant~|
|~`babel-preset-yoshi`~|||~Not relevant~|
|~`bootstrap-hot-loader`~|@ronami|#1447|**Merged**|
|~`create-yoshi-app`~|@amitdahan|#1598|**Merged**|
|~`eslint-config-yoshi`~|||~Not relevant~|
|~`eslint-config-yoshi-base`~|||~Not relevant~|
|`jest-yoshi-preset`|@yanivefraim|#1473|Up for grabs|
|~`stylelint-config-yoshi`~|||~Not relevant~|
|~`tslint-config-yoshi`~|||~Not relevant~|
|~`tslint-config-yoshi-base`~|||~Not relevant~|
|`yoshi`|||Up for grabs|
|~`yoshi-angular-dependencies`~|||~Not relevant~|
|~`yoshi-config`~|@ronami|#1401|**Merged**|
|~`yoshi-helpers`~|@ronami|#1463|**Merged**|
|~`yoshi-style-dependencies`~|||~Not relevant~|
|`yoshi-template-intro`|||Up for grabs|

#### How to migrate

Because Yoshi is a monorepo, we have dependencies between packages. To avoid having to define types for JavaScript packages, we will start by migrating leaf packages (without dependencies) and then walk up the dependency tree until everything is migrated.

1. Start by claiming a package in this issue.
2. Copy `tsconfig.json` from an already migrated package.
3. If the package you're migrating have dependencies on another package in this repo, use `references` (See [project references](https://www.typescriptlang.org/docs/handbook/project-references.html)).
4. Rename all files with `js` extension to `ts` (or `tsx` if they have jsx in them).
5. Change all `ts` files to use EcmaScript modules instead of CommonJs (Use [Lebab](https://github.com/lebab/lebab) to make it easier).
6. Make sure tests, type-checking and lint pass.
7. Open up a PR and comment on this issue.

Hope this goes smoothly.

amitdahan:
Hey guys! What's the plan regarding `scripts` folders?
Some are outside any of the sub-packages, some are in.
[...]

PR #1598, opened 2019-10-10
Migrate create-yoshi-app to TypeScript

amitdahan:
### 🔦 Summary
Part of #1462.

### 🗺️ Test Plan
...


packages/create-yoshi-app/package.json
```diff
@@ -9,11 +9,11 @@
...
   ],
   "author": "Ran Yitzhaki & Ronen Amiel",
-  "main": "dist/src/index.js",
-  "bin": "dist/src/bin/create-yoshi-app.js",
+  "main": "build/src/index.js",
+  "bin": "build/src/bin/create-yoshi-app.js",
   "scripts": {
     "test": "jest",
-    "templates": "node scripts/runE2E.js",
+    "templates": "ts-node scripts/runE2E.js",
```
amitdahan: I preferred not migrating scripts since this wasn't done anywhere else yet, so this was potentially unneeded, but the `runE2E.js` script eventually imports one of the source files from `src` which fails since that is now a `ts` file, we use `ts-node` merely to allow imports from `ts` files during this script.

packages/create-yoshi-app/__tests__/create-yoshi-app.test.ts
```diff
@@ -1,9 +1,10 @@
...
-const fs = require('fs');
-const path = require('path');
+import fs from 'fs';
+import path from 'path';
+import execa from 'execa';
+import tempy from 'tempy';
+import { bin } from '../package.json';
 
-const createYoshiAppBin = path.join(__dirname, '../bin/create-yoshi-app.js');
+const createYoshiAppBin = path.resolve(path.resolve(__dirname, '..'), bin);
```
ronami: Can be done with a single `path.resolve(__dirname, '..', bin)`

packages/create-yoshi-app/package.json
```diff
@@ -58,6 +61,8 @@
     "minimist": "^1.2.0",
     "mocha": "^5.2.0",
     "mocha-env-reporter": "^3.0.0",
-    "tempy": "^0.2.1"
+    "tempy": "^0.3.0",
+    "ts-node": "^8.4.1",
+    "typescript": "^3.5.3"
```
ronami: Not needed as we have a single TypeScript dependency for the entire monorepo (at the root `package.json`)

packages/create-yoshi-app/src/getQuestions.ts
```diff
@@ -69,5 +70,5 @@ module.exports = () => {
         { title: 'Babel', value: 'babel' },
       ],
     },
-  ];
+  ] as Array<PromptObject<string>>;
```
ronami: I think it would be better to define the function's return type as `Array<PromptObject<string>>` instead of casting this value to that. Wdyt?

packages/create-yoshi-app/src/templates.ts
```diff
@@ -1,9 +1,10 @@
-const { join } = require('path');
+import { join, resolve } from 'path';
+import { TemplateDefinition } from './TemplateModel';
 
-const toTemplatePath = templateName =>
-  join(__dirname, 'templates', templateName);
+const toTemplatePath = (templateName: string) =>
+  join(resolve(__dirname, '../templates'), templateName);
```
ronami: Can be done with a single `path.resolve()`

packages/create-yoshi-app/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build"
+  },
+  "include": [
+    "src/**/*",
+    "src/**/*.json"
+  ],
+  "files": ["package.json"],
+  "references": [
```
ronami: Can be removed instead of an empty array

packages/create-yoshi-app/src/getQuestions.ts
amitdahan: Ok this brought up something interesting, apparently when I did it like that it indeed as you say "casted" the type, and apparently this even skipped some type checking for this array.
When I moved this type annotation to the function's return type, we now see a type error, saying `value` must be a string, and we pass in a `TemplateModel` object (`{ name, path }`).
This actually works, and I couldn't find docs for whether this is supported or not, so I checked the source and it makes sense that `value` can be anything and not just a string, so I pushed a [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39269).
I can push the commit that moves the type to the function's return type but I'll have to silence the type error by casting the value to `any`. Wdyt?

packages/create-yoshi-app/__tests__/create-yoshi-app.test.ts
amitdahan: Awesome, pushed!

packages/create-yoshi-app/package.json
amitdahan: Oh whoops, fixed!

PR #1613, opened 2019-10-17
Yoshi BM Flow

amitdahan:
### 🔦 Summary
This introduces a new template to `create-yoshi-app` and adds an npm script to test it the way we test the rest of the templates.


packages/create-yoshi-app/scripts/runE2E.js
```diff
@@ -21,6 +21,8 @@ if (focusProjects) {
   process.env.FOCUS_PROJECTS = focusProjects;
 }
 
+process.env.EXPERIMENTAL_FLOW_BM = true;
```
ranyitz:
```suggestion
process.env.EXPERIMENTAL_FLOW_BM = 'true';
```

packages/create-yoshi-app/src/templates.ts
```diff
@@ -19,4 +19,11 @@ const templates: Array<TemplateDefinition> = [
   },
 ];
 
+if (process.env.EXPERIMENTAL_FLOW_BM) {
```
ranyitz:
```suggestion
if (process.env.EXPERIMENTAL_FLOW_BM === 'true') {
```

PR #1614, opened 2019-10-18
Fail template generation if node version does not match ours

### LINKED ISSUE #1569: create-yoshi-app: A case when user's node version is different than the template's node version

amitdahan:
### 🔦 Summary
Closes #1569.

Instead of validating for `>= 8.7.0` as we currently do, which allows template generation for node `12` for example, whereas the template to be generated would eventually require `10`, we now fail generation if current node version does not MATCH required version.

This means up until now `create-yoshi-app` supported `>= 8.7.0` and now it (arguably?) correctly checks for `10`.

### 🗺️ Test Plan
- Unit tests for valid and invalid node version.
- Unit tests to make sure all the following match:
`yoshi/.nvmrc`
`yoshi/packages/create-yoshi-app/.nvmrc`
`yoshi/packages/create-yoshi-app/templates/*/javascript/.nvmrc`
`yoshi/packages/create-yoshi-app/templates/*/typescript/.nvmrc`


amitdahan:
Red until https://github.com/wix/yoshi/pull/1617.

packages/create-yoshi-app/src/verifyNodeVersion.ts
```diff
@@ -0,0 +1,27 @@
+import path from 'path';
+import fs from 'fs';
+import semver from 'semver';
+import chalk from 'chalk';
+
+export default () => {
+  const requiredVersion = fs
+    .readFileSync(
+      path.join(
+        path.dirname(require.resolve('create-yoshi-app/package.json')),
```
ranyitz:
```js
const nvmrcPath = path.join(__dirname, '../.nvmrc');
```

packages/create-yoshi-app/src/verifyNodeVersion.ts
```diff
@@ -0,0 +1,27 @@
...
+        '.nvmrc',
+      ),
+    )
+    .toString()
+    .trim();
+
+  if (!semver.satisfies(process.version, requiredVersion)) {
+    console.log(
+      `Node version ${chalk.cyan(
+        process.version,
```
ranyitz: put into a variable

packages/create-yoshi-app/__tests__/verifyNodeVersion.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+};
+
+const readFile = (filePath: string) =>
+  fs
+    .readFileSync(filePath)
+    .toString()
+    .trim();
+
+const nvmrc = readFile(path.resolve(__dirname, '../.nvmrc'));
```
ranyitz: readNvmrc 😆

PR #1623, opened 2019-10-24
Extend `--source-map` argument to support `false` value

yavorsky:
### 🔦 Summary
In two words this PR adds an option to get rid of source-maps on CI :)
For ci we have `'devtool: souce-map'`, which is basically the largest and slowest in terms of generation type of source map. For big projects it takes a lot of time and sometime causing build to be fail with memory error:
https://github.com/webpack/webpack-sources/issues/66
https://pullrequest-tc.dev.wixpress.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=4998301&_focus=1289

It's not reproducible locally since we have 'cheap-module-eval-source-map' to be generated.
With current configuration approach it's not possible to disable source-map generation on CI.
Proposed solution is t extend `--source-map` argument to support boolean values (ex. `--source-map-false`). Old syntax (`--source-map`) is still supported.

### 🗺️ Test Plan
Updated a `build.spec.js` tests to check that

1. `--source-map` still working
2. `--source-map=true` is the same as an 1 option.
3. `--source-map=false` is preventing source maps to be created even if environment is a TC.

### DIFF
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -126,9 +126,9 @@ run webpack-bundle-analyzer plugin.
 
 output webpack stats file to `dist/webpack-stats.json` (see also [bundle analysis](../guides/bundle-analyze.md))|
 
-#### `--source-map`
+#### `--source-map` or `--source-map=[true|false]`
 
-Explicitly emit bundle source maps.
+Explicitly emit or bypass source maps generation.
 
 This task will perform the following:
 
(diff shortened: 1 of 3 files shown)


ranyitz:
Hey, let me try and answer some of the questions raised here:

> It's not reproducible locally since we have 'cheap-module-eval-source-map' to be generated.

You can add the CLI argument `--source-map` and it will use the `source-map` option

https://github.com/wix/yoshi/blob/cb5874070ba7fc57bd4426cf2f2687d934cab917/packages/yoshi/src/commands/build.js#L104

> With current configuration approach it's not possible to disable source-map generation on CI.

This is intentional, we want developers to have the best debugging experience on production. We don't want them to disable sourcemaps and then not to be able to debug their application in production.

________

With that said, any solution that optimizes the current behavior and **doesn't compromise production debugging experience** is acceptable.

PR #1627, opened 2019-10-28
run install & lint on first time project generation

ranyitz:
### 🔦 Summary
Run installation before linking on the first time after project generation.


yavorsky:
Hey @ranyitz . Looks good! I just want to ask - why do we need to lint something that was user didn't touch yet?

ranyitz:
Hey @yavorsky. We use `prettier` through `eslint`. After inserting dynamic values, the line length is affected and as a result after generation we may have lint warning (prettier warnings). By running `eslint --fix` we get to a pristine state (which is similar to what we do in production on initial app creation)

ranyitz:
Actually, this PR inserted a degradation for cases that we create a new module and use it in one of our templates, in this case `npm install` will fail.

In order to solve it, i'll try to use the "lerna" strategy to install only the "unlinked" dependencies from `npm` and link the relevant ones from `Yoshi`.

I'll open a PR soon

PR #1628, opened 2019-10-28
Remove thread-loader

ronami:
### 🔦 Summary

See https://github.com/facebook/create-react-app/pull/5170.


yavorsky:
I have few questions :)

1. Do our users experience warnings/errors from: https://github.com/facebook/create-react-app/issues/5138#issue-364555898
2. Maybe we should take some big projects we have and make a comparison to understand how cold build is in DEV or PROD slower?
3. Per comment https://github.com/facebook/create-react-app/issues/5138#issuecomment-425631999 CRA going to write own worker that handles caching and parallelization. Do we?

PR #1629, opened 2019-10-29
`manifest.json` for local dev

yanivefraim:
`manifest.json` is a file which contains static asset urls. This is useful / important in the world of `HtmlWebpackPlugin`, where we have a content based hash for each asset. (Different consumers will be able to query this manifest file and know the urls for the static files).

We have 3 types of manifest files:
1. `manifest.dev.json` - created during `start` command, useful for local developement only.
2. `{artifactVersion}/manifest.debug.json` - created during build command and points on non optimised bundles.
3. `{artifactVersion}/manifest.prod.json` - created during build command and points on optimised bundles.

Important notes: 
- Assets (such as `.png` files) will not be listed in the manifest.
- Non initial chunks (such as dynamic import chunks) will not be listed in the manifest.


packages/yoshi/src/commands/utils/start-single-app.js
```diff
@@ -215,4 +219,8 @@ module.exports = async (app, options) => {
   }
 
   openBrowser(options.url || app.startUrl || `http://localhost:${PORT}`);
+
+  clientCompiler.hooks.done.tap('promise', async stats => {
```
ronami: How about moving this to a Webpack plugin? https://webpack.js.org/api/plugins.

yanivefraim: Done!

ranyitz:
> The whole purpose of the manifest is to locate assets without knowing GA version.

Maybe I'm missing something, but how would we support testing RC versions if we override the manifest on every GA?

I do think that there is value to the manifest even when it's versioned. What do we try to achieve here?

test/projects/kitchensink-javascript/integration/manifest.test.js
```diff
@@ -19,23 +19,60 @@ async function replaceOriginalEntry(relativeFilePath) {
...
+    it('generates manifest stat file', async () => {
+      startResult = await global.scripts.start(localEnv);
 
-    const statsFilePath = path.join(
-      process.env.TEST_DIRECTORY,
-      'dist/statics/manifest.1.0.0.json',
-    );
+      const statsFilePath = path.join(
+        process.env.TEST_DIRECTORY,
+        'dist/statics/manifest.debug.json',
```
ranyitz: This naming makes me think that the `dev` manifest is similar to the `debug` one.

I want to state here that we have `dev`/`debug`/`prod` which are different for one another.

Also, this one is unversioned (which already makes it different for heuristics that will require it during dev mode).

__________

My suggestion would be to go with `manifest.dev.json` or `manifest.json`.

test/projects/kitchensink-javascript/integration/manifest.test.js
```diff
@@ -19,23 +19,60 @@ async function replaceOriginalEntry(relativeFilePath) {
...
+
   afterEach(() => {
     // reset state back to normal after every test
     fs.writeFileSync(originalFilePath, originalContent);
   });
 
-  it('generates manifest stat file', async () => {
-    await replaceOriginalEntry('src/cases/manifest.js');
+  describe('build', () => {
+    it('generates manifest stat file for initial chunks', async () => {
```
ronami: better title: `for dev` / `for prod`

test/projects/kitchensink-javascript/integration/manifest.test.js
```diff
@@ -19,23 +19,60 @@ async function replaceOriginalEntry(relativeFilePath) {
...
+      await global.scripts.build(ciEnv);
```
ronami: single build for multiple expectations

test/projects/kitchensink-javascript/integration/manifest.test.js
```diff
@@ -19,23 +19,60 @@ async function replaceOriginalEntry(relativeFilePath) {
 }
 
 describe('manifest', () => {
+  beforeEach(() => replaceOriginalEntry('src/cases/manifest.js'));
```
ronami: can be `before()` / `after()`

test/utils.js
```diff
@@ -107,11 +107,23 @@ function waitForPortToFree(port) {
   });
 }
 
+function waitForStdout(process, stringToMatch) {
```
ronami: Variable `process` can be called `spawnedProcess`?

yanivefraim:
@ronami - Thanks, all fixed

yanivefraim:
Closing in favor of #1681

PR #1630, opened 2019-10-30
adding verbose stdio

deanshub:
### 🔦 Summary
Reverting #1604

After getting exceptions from 
https://github.com/wix/yoshi/pull/1604/files#diff-f7a84b342dbdcfc34f734f90253f49baR86
the log in https://github.com/wix/yoshi/pull/1604/files#diff-f7a84b342dbdcfc34f734f90253f49baR86
will never happen, the error will fail the CI and we won't get the message in the build log.

So reverting the code will allow the stdout\err to be logged using verbosity env var


ronami:
Instead of completely reverting the change, we can wrap the `lerna.exec()` call in a `try {} catch {}` clause and `console.log()` the error (which includes all logs from the process)?

I don't understand if the change to always being verbose had a bug or if it was incorrect.

deanshub:
you're right, referencing #1634 to answer your question

Issue #1468, opened 2019-07-22
Single executable


zemlanin:
# 🙋 Feature Request
Distribute `yoshi` (and related packages) as a single executable to make installs and updates faster

### 😯 Current Behavior
Because of more than two thousand dependencies (_excluding_ `devDependencies`), `yoshi` takes _a lot_ of time and storage to install. For example, having this `package.json`:

```json
{
  "name": "egg",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "yoshi": "^4.10.4"
  }
}
```

it will take a long time to install this package:
```
> npm i  // no package-lock.json, deleted node_modules
added 2096 packages from 1231 contributors in 233.738s

> npm i  // with package-lock.json, deleted node_modules
added 2096 packages from 1231 contributors in 27.096s

> yarn i  // no yarn.lock, deleted node_modules
✨  Done in 113.83s.

> yarn i  // with yarn.lock, deleted node_modules
✨  Done in 14.90s.
```

and package will be bigger than a quarter of a gigabyte
```
> du -h
...
4.0K	./node_modules/postcss-reduce-initial/dist
 12K	./node_modules/postcss-reduce-initial/data
 28K	./node_modules/postcss-reduce-initial
4.0K	./node_modules/setprototypeof/test
 24K	./node_modules/setprototypeof
288M	./node_modules
289M	.
```

### 🤔 Expected Behavior
<img width="322" alt="Smaller, Lighter, Faster" src="https://user-images.githubusercontent.com/711514/61642018-ddf7d280-aca8-11e9-9a67-eab62a4424d2.png">

### 💁 Possible Solutions
Moving most of `dependencies` to `devDependencies` and bundling `yoshi` before publishing it on npm

ranyitz:
@zemlanin Thanks for this research!
you've raised some interesting points.
__________________________________________

1. We should all use yarn 😄 (According to your benchmark)
2. It's possible that distributing Yoshi into a single executable will improve install time, but we need to have a POC and see if it worth the trade-offs.
 * Would it be as easy to debug other build related dependencies as today?
 * How would it affect our build process?
 * How would it be installed on each project?

_____________________________

[...]

PR #1635, opened 2019-11-01
Use yarn

ranyitz:
### 🔦 Summary
Yarn has proven itself as a faster package manager client. You can see benchmarks of @zemlanin in this comment https://github.com/wix/yoshi/issues/1468#issue-471140071

Also, we believe that using yarn workspaces will work faster for our own monorepo setup, so this PR moves us to use `yarn` & `yarn workspaces`

____________

I've added a `yarn.lock` as well.

One of the problems in doing that was the fact that we wanted to see our build breaks when a transitive dependency down the line breaks. 

I believe that we still have our great `e2e` testing infrastructure that should give us a perfect view of our user's state. We install everything just like a user, which means that those tests would most likely to break on the case presented above.

This means we can use the fast installs that come along with a `yarn.lock`.

Let me know if I'm missing something @ronami @deanshub @amitdahan @netanelgilad @yanivefraim


lerna.json
```diff
@@ -1,5 +1,7 @@
 {
   "lerna": "3.15.0",
+  "npmClient": "yarn",
+  "useWorkspaces": true,
   "packages": [
     "packages/*"
   ],
```
ronami: Replace with [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/#toc-tips-tricks)

PR #1641, opened 2019-11-05
Page close error level

### LINKED ISSUE #1642: "Cannot read property 'close' of undefined" when jest e2e teardown is happening
# 🐛 Bug Report

### 😯 Current Behavior

Puppeteer page is sometimes undefined in the case where it should be available - the teardown after an e2e test. The teardown then errors and fails the test run.

### 🤔 Expected Behavior

No flakyness

### 🔦 Context 

It randomly breaks our products builds although the tests pass (server tests, which run in an e2e environment)

### 💁 Possible Solution

Guess: make sure the page is always there before starting the suite. But this needs investigation.
Possible workaround: do not break the build and just warn if this happens. If the tests are green, we don't want to break the suite for such thing. If the tests are red or using the global.page, it will fail anyway, so I see no drawbacks for such logic. Created a PR: https://github.com/wix/yoshi/pull/1641 

### 👣 Steps to Reproduce

We have it in our CI in Wix, please contact me for exact URL, but it basically looks like this and occasionally occurs (flaky)
Screenshot 2019-11-05 at 15 13 17

Not sure how to reproduce locally, but this been happening for a long time - let us know if you have any questions about the issue

### 🌍 Your Environment

Yoshi 4.10.0
jest-yoshi-preset 4.10.0

Tried yoshi 4.17.2 also, problem did not go away

rokasmik:
### 🔦 Summary
Fixes #1642 
There is an issue for global.page not existing when it is supposed to, resulting in an error which breaks a test suite. In our project, this happens in server integration tests, which does not require the page to be available. So we don't think this error is worth "break build" level, so reducing it to a console.warn .
For situations like ours, when the tests suite passes successfully, this does not give a negative effect. For situations when the page is actually consumed, the tests will still fail - this does not fix the issue, but reduces the consequences of it.

### 🗺️ Test Plan

No need to test


rokasmik:
@ranyitz ,
thank you, this makes sense - converted this to a check where page is checked before closing. Added a comment for clarification - let me know if you think that the comment is not needed.
Also, this does not fix the issue itself, just reduces the impact, maybe we shouldn't close https://github.com/wix/yoshi/issues/1642 with this PR?

PR #1647, opened 2019-11-06
Split to multiple flows

ronami:
### 🔦 Summary

Massive effort, still a work in progress.


packages/yoshi-common/src/verify-dependencies.ts
```diff
@@ -11,11 +11,15 @@ const relatedPackages = [
   'yoshi-angular-dependencies',
 ];
 
-module.exports = async () => {
+function truthy<T>(x: T | undefined | null): x is T {
```
amitdahan: 😍

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -283,6 +285,8 @@ export function createBaseWebpackConfig({
   const config: webpack.Configuration = {
     context: join(SRC_DIR),
 
+    name: configName,
```
ranyitz: We don't really have a specific test that verifies that, but it affects webpack's error messages and thus makes some output app flow tests to fail :)

ronami: Nice catch. It's nice to have a name per configuration when running multiple Webpack configurations. This way, if one fails, you know exactly which one. 👍

ranyitz: Yeah, you're right! Maybe we should have an assertion for it so in case it fails we'll know right away where it should be fixed. (not super important though, just a thought)

packages/yoshi/package.json
```diff
@@ -17,7 +17,8 @@
     "test:others": "npm run mocha -- 'test/{webpack-config,config,lint}.spec.js' 'src/**/*.spec.js' 'test/tasks/**/*.spec.js'"
   },
   "files": [
-    "bin"
+    "bin",
+    "types.d.ts"
```
ranyitz: indeed 😄

packages/yoshi-flow-app/tsconfig.json
```diff
@@ -8,6 +8,9 @@
     "src/**/*",
     "src/**/*.json"
   ],
+  "files": [
+    "src/external-types.d.ts"
+  ],
```
ronami: Not needed, all files in `src` are already included.

packages/yoshi-common/src/hot-reloader.ts
```diff
@@ -166,10 +166,18 @@ export default class {
   }
 
   async start() {
+    const serverCompilerDonePromise = new Promise(resolve => {
+      const [, serverCompiler] = this.webpackDevServer.compiler.compilers;
+
+      serverCompiler.hooks.done.tap('server-done', () => {
+        resolve();
+      });
```
ronami: can be a single line but super small

packages/yoshi-flow-app/tsconfig.json
ranyitz: Yes you're right!

packages/yoshi-common/src/hot-reloader.ts
ranyitz: Fixed

packages/yoshi-flow-legacy/bin/yoshi-legacy.js
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
+
+const importLocal = require('import-local');
+const verifyMinimumNodeVersion = require('yoshi-helpers/verifyMinimumNodeVersion');
+const { minimumNodeVersion } = require('yoshi-common/build/utils/constants');
+
+verifyMinimumNodeVersion(minimumNodeVersion);
```
ranyitz: We need to add both `verifyMinimumNodeVersion` and `importlocal` to app-flow. Actually I don't see a good reason not to use them on every flow, I'll try to think of something.

packages/yoshi-flow-legacy/config/webpack.config.js
```diff
@@ -0,0 +1,166 @@
...
+      entryConfig = { ...entryConfig, server: serverEntry };
+    }
+
+    return entryConfig;
+  };
+
+  return webpackConfig;
+}
+
+function createWebWorkerWebpackConfig({ isDebug = true, isHmr = false }) {
```
ranyitz: I don't think that we have a usecase of the webWorker on legacy flow. Maybe we should start without it and add it just in case.

packages/yoshi-flow-legacy/package.json
```diff
@@ -0,0 +1,168 @@
...
+    "webpack": "4.41.2",
+    "webpack-dev-middleware": "3.7.2",
+    "webpack-hot-client": "4.1.2",
+    "wnpm-ci": "8.0.87",
+    "word-wrap": "1.2.3",
+    "xmldoc": "1.1.2",
+    "yoshi-common": "4.18.0",
+    "yoshi-config": "4.18.0",
+    "yoshi-helpers": "4.18.0",
+    "yoshi-runtime": "1.0.314",
```
ranyitz: We used to bring `yoshi-runtime` to everyone?

packages/yoshi/bin/yoshi.js
```diff
@@ -1,10 +1,23 @@
...
-verifyMinimumNodeVersion(minimumNodeVersion);
+const execa = require('execa');
+const config = require('yoshi-config');
 
-if (!importLocal(__filename)) {
-  require('./yoshi-cli');
+const passedArgv = process.argv.slice(2);
+const command = passedArgv[0];
+
+function runCommand(binPath) {
```
ranyitz: Reminder to ourselves that we may be able to change this into a regular `require` instead of spawning a new process.

scripts/build.js
```diff
@@ -33,6 +33,9 @@ const args = [
 console.log(chalk.inverse('Building TypeScript files'));
 process.stdout.write(adjustToTerminalWidth('Building\n'));
 
+// TODO - add watch capability
+copyAssets();
```
ranyitz: ohhh (adding a reminder)

packages/yoshi-config/src/root-app.ts
```diff
@@ -1,11 +1,11 @@
-import { getPaths } from './paths';
-import config from './index';
+// import { getPaths } from './paths';
```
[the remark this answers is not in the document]
ronami: I thought so but I'll just see the original code from `master`

.npmrc
```diff
@@ -1 +1 @@
-package-lock = false
+package-lock = true
```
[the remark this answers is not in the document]
ronami: No, nice catch

packages/yoshi-flow-legacy/package.json
ronami: Unfortunately :( https://github.com/wix/yoshi/blob/master/packages/yoshi/package.json#L157

packages/yoshi-flow-app/src/scripts/build.ts
```diff
@@ -0,0 +1,84 @@
...
+      '--stats': Boolean,
+      '--source-map': Boolean,
+    },
+    { argv },
+  );
+
+  const {
+    '--analyze': isAnalyze,
+    '--stats': shouldEmitWebpackStats,
+    // '--source-map': shouldCreateFullSourceMaps,
```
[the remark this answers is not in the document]
ranyitz: done

packages/yoshi-flow-legacy/bin/yoshi-cli.js
```diff
@@ -0,0 +1,114 @@
...
+
+const prog = require('commander');
+const runCLI = require('../src/cli');
+const { version } = require('../package');
+const infoCommand = require('../src/commands/info');
+const chalk = require('chalk');
+const { inTeamCity } = require('yoshi-helpers/queries');
+
+// IDEs start debugging with '--inspect' or '--inspect-brk' option. We are setting --debug instead
+require('yoshi-common/normalize-debugging-args')();
```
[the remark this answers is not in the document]
ranyitz: Added specifically to start command (which is currently the only one needs it in `yoshi-flow-app`)

The other command is `test` which uses the legacy flow (which already normalizing the args)

packages/yoshi/bin/yoshi.js
ranyitz: done

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -310,6 +312,14 @@ export function createBaseWebpackConfig({
       hotUpdateMainFilename: 'updates/[hash].hot-update.json',
       hotUpdateChunkFilename: 'updates/[id].[hash].hot-update.js',
 
+      ...(exportAsLibraryName
+        ? {
+            library: exportAsLibraryName,
+            libraryTarget: 'umd',
```
ranyitz: Wow, that's critical. Also things like `externals` won't work (Which also caused `ooi` tests to fail constantly)

Nice catch

packages/yoshi-flow-app/src/bin/yoshi-app.ts
```diff
@@ -23,6 +23,7 @@ const commands: {
   test: () => import('yoshi-flow-legacy/bin/yoshi-legacy'),
   lint: () => import('yoshi-flow-legacy/bin/yoshi-legacy'),
   info: () => import('yoshi-flow-legacy/bin/yoshi-legacy'),
+  release: () => import('yoshi-flow-legacy/bin/yoshi-legacy'),
```
ranyitz: 🥂

Issue #1639, opened 2019-11-04
Always emit Webpack stats as a JSON file


ronami:
### Summary

Currently, we're emitting Webpack's stats (JSON) file only when `npx yoshi build --stats` is run. 

Webpack's stats file is mostly used by analyze tools (other than the built-in `webpack-analyzer-plugin`) but it's also useful for servers that want to load the minimal number of chunks in a client project (See https://github.com/faceyspacey/react-universal-component).

Instead of having a CLI argument that only works for `build` command, we can always write this file, both, when running `build` and `start` (with watch). Writing this file doesn't cost a lot and it helps remove somewhat redundant configurations. If a process that runs locally (like a fullstack server) expects it, it will also be available locally.

See  [webpack-stats-plugin](https://github.com/FormidableLabs/webpack-stats-plugin) which can help extract some duplicated code that will be shared between `start` and `build`.

yanivefraim:
@ronami - can we close this one?

PR #1648, opened 2019-11-06
Write Webpack stats file on start command

yanivefraim:
We want to generate Webpack's `stats` file in start / build commands. See https://github.com/wix/yoshi/issues/1639 for more details


packages/yoshi/src/webpack-plugins/stats-writer-plugin.js
```diff
@@ -0,0 +1,52 @@
+module.exports = class StatsWriterPlugin {
```
yanivefraim: This Plugin is based on https://github.com/FormidableLabs/webpack-stats-plugin. I couldn't use it because it [hard-codes](https://github.com/FormidableLabs/webpack-stats-plugin/blob/master/lib/stats-writer-plugin.js#L83) the `'forToString'` in `toJson(options, forToString)`. Our current version does not use this second argument (adding it changes stats output).

I can add a PR for changing this arg in https://github.com/FormidableLabs/webpack-stats-plugin, or consider using it with  `'forToString'` argument

test/projects/kitchensink-javascript/integration/stats.test.js
```diff
@@ -1,17 +1,39 @@
...
+        'target/webpack-stats.json',
+      );
 
-    expect(fs.existsSync(statsFilePath)).toBeTruthy();
+      expect(fs.existsSync(statsFilePath)).toBeTruthy();
+    });
+  });
+  describe('start', () => {
+    let startResult;
+    afterEach(async () => {
```
ranyitz: We have only one test here, maybe we can go with a `try` and `finally` blocks instead of `afterEach`?

See https://github.com/wix/yoshi/blob/43293887f1d5bc91e6060f98c16a5e9b6b1ac997/test/projects/kitchensink-javascript/integration/analyze.test.js#L9-L16

packages/yoshi/config/webpack.config.js
```diff
@@ -780,6 +782,21 @@ function createClientWebpackConfig({
             }),
           ]
         : []),
+
+      ...(withStats
+        ? [
+            new StatsWriterPlugin({
+              filename: path.relative(
```
ranyitz: The fact the filename has to be a relative path to the output dir wasn't clear to me at first, maybe it worth adding a comment above.

```js
// https://github.com/FormidableLabs/webpack-stats-plugin/blob/master/README.md#L170-L171
```

Issue #1638, opened 2019-11-04
Support Web Workers in Webpack


ronami:
### Summary

Currently, spawning your own web-workers with `new Worker()` is very hard. You can create a web-worker bundle with [config.webWorker](https://github.com/wix/yoshi/blob/0dd8bc71b8c5429302db804507b5afc377f6c497/packages/yoshi-config/src/config.ts#L69) and target `new Worker('...')` to its output but it's very cumbersome and not immediately obvious.

A better way to add seamless support for web-workers is with [worker-plugin](https://github.com/GoogleChromeLabs/worker-plugin) or [worker-loader](https://github.com/webpack-contrib/worker-loader). Both statically analyze the app code and create child compilations that target `webworker`.

See create-react-app's in progress [issue](https://github.com/facebook/create-react-app/issues/3660) and [PR](https://github.com/facebook/create-react-app/pull/5886) for more information.

ranyitz:
Closed in #1650

PR #1650, opened 2019-11-07
support importing web workers

roymiloh:
Support importing web workers using [`worker-loader`](https://github.com/webpack-contrib/worker-loader).

Closing https://github.com/wix/yoshi/issues/1638.


packages/yoshi/types.d.ts
```diff
@@ -107,3 +107,8 @@ declare module '*.graphql' {
   const value: DocumentNode;
   export = value;
 }
+
+declare module '*.worker.ts' {
```
yanivefraim: We should better have `*.inline.worker.ts`, right?

test/projects/kitchensink-javascript/src/components/features/web-worker-chunk-container.js
```diff
@@ -0,0 +1,21 @@
...
+export default class WebWorker extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { text: '' };
+  }
+
+  componentDidMount() {
+    this.worker = new Worker();
+    this.worker.addEventListener('message', () => {
```
yanivefraim: nit: I would test for the actual message ('halo', in this case) to arrive, just to be 100% sure (:

test/projects/kitchensink-javascript/integration/loaders.test.js
```diff
@@ -324,4 +324,21 @@ describe('webpack', () => {
       expect(result).toBe('This is an abstract.');
     });
   });
+
+  describe('web worker', () => {
+    it('web worker inclusion', async () => {
+      await initTest('web-worker-chunk-container');
```
yanivefraim: nit: the word `chunk` here might be confusing, I would remove it (also from the inline entry name)

PR #1656, opened 2019-11-10
add watch capability

ranyitz:
### 🔦 Summary
This PR moves our build script to TypeScript and add a watch capability to `copy-assets` task.

__Few things to note:__
* I'm using `execa.sync` only on `build` (not `watch`) because it will block the main thread and `copy-assets`'s watcher won't work.


scripts/copy-assets.ts
```diff
@@ -0,0 +1,77 @@
...
+import chokidar from 'chokidar';
+import multimatch from 'multimatch';
+
+export default function copyAssets(
+  packagesDir: string,
+  packagesList: Array<string>,
+  watch: boolean,
+) {
+  const patterns = packagesList.reduce((acc: Array<string>, pkg: string) => {
+    return [...acc, `${pkg}/src/**/*`, `!${pkg}/src/**/*.(ts|tsx)`];
```
ronami: exclude `json`

PR #1657, opened 2019-11-11
support browser env var

roymiloh:


test/projects/kitchensink-typescript/src/components/features/env-vars.tsx
```diff
@@ -4,5 +4,6 @@ export default () => (
   <div id="env-vars">
     <p id="node-env">{process.env.NODE_ENV}</p>
     <p id="ci-app-version">{window.__CI_APP_VERSION__}</p>
+    <p id="browser">{JSON.stringify(process.env.browser)}</p>
```
ronami: Why do we need to stringify it here and not for `process.env.NODE_ENV`?

PR #1678, opened 2019-11-15
Fix way to resolve babel-preset-yoshi from yoshi-helpers

yavorsky:
### 🔦 Summary
After latest refactor we've forgot to update the way `createBabelConfig` actually trying to find the root of the directory where `babel-preset-yoshi` is present.
So before we had small workaround which overrides paths to resolve it and adds `yoshi`. For now, we don't have any `babel-preset-yoshi` in `yoshi`'s dependencies, moreover w/o `main` field, node.js won't even resolve `yoshi` package [here](https://github.com/wix/yoshi/compare/fix-babel-preset-yoshi?expand=1#diff-11cc9eea6089afbf667c21ee89086cd2L49).

So since yoshi-helpers is using it, let keep it as a dependency.

Screen Shot 2019-11-15 at 1 28 49 PM

If you want to reproduce it locally, just clone for example https://github.com/wix-private/analytics
and remove babel-preset-yoshi from root dependency (workaround for them until it will be fixed).

### DIFF
--- a/packages/yoshi-helpers/package.json
+++ b/packages/yoshi-helpers/package.json
@@ -27,6 +27,7 @@
   "peerDependencies": {
     "@babel/core": "^7.1.2",
     "@babel/plugin-transform-modules-commonjs": "7.6.0",
+    "babel-preset-yoshi": "4.14.0",
     "typescript": "^2.9.0 || ^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


ranyitz:
@yavorsky Actually today it's part of the [legacy flow](https://github.com/wix/yoshi/blob/a34f9ba00ac84396148f0f01fceda1007286945c/packages/yoshi-flow-legacy/package.json#L38) which is required from all other flows, at least right now. In the future, we would like to gradually remove the `yoshi-helpers` package in favor of other flows / the `yoshi-commons` package.

ranyitz:
Solved on #1847

PR #1685, opened 2019-11-17
Correctly link Yoshi's CLIs

amitdahan:
### 🔦 Summary
It appears the existing symlink logic for Yoshi's CLIs didn't really do its job.

What happened up until recently, is that we first installed the project's dependencies (including production yoshi and friends), which in turn added the symlink `node_modules/.bin/yoshi -> node_modules/yoshi/bin/yoshi-cli.js`).
We then continued to symlink `yoshi` itself, by adding the symlink `node_modules/yoshi -> yoshi-monorepo/packages/yoshi`, which means the CLI's symlink pointed at the correct code.
We then placed a symlink at the project's *parent `node_modules/.bin`*, which did nothing because `npx` nor regular npm scripts resolve CLIs in the parent `node_modules/.bin`.
A recent change that stopped installing local packages before linking them uncovered this bug.

This should fix that.

### 🗺️ Test Plan
This is intended for local development, so no testing needed.
Was tested locally though.


scripts/utils/symlinkModules.js
```diff
@@ -33,22 +29,29 @@ module.exports.symlinkModules = repoDirectory => {
     );
   };
 
-  const packageJsonPath = path.join(repoDirectory, 'package.json');
-  const pkg = fs.readJsonSync(packageJsonPath);
+  const symlinkPackageBins = packageDir => {
+    const { bin: bins = {} } = fs.readJSONSync(
```
amitdahan: Fix `create-yoshi-app`'s `bin`.

PR #1688, opened 2019-11-18
create-yoshi-app: update Server template with Node Platform types

alexandervain:
Following introduction of types in some modules of Node Platform, we want to update the _Server_ (typescript) template to start using it.


packages/create-yoshi-app/templates/server/typescript/src/server.ts
```diff
@@ -1,9 +1,9 @@
 import { Router } from 'express';
-import * as wixExpressCsrf from '@wix/wix-express-csrf';
-import * as wixExpressRequireHttps from '@wix/wix-express-require-https';
-import { AppContext } from './config';
+import wixExpressCsrf from '@wix/wix-express-csrf';
+import wixExpressRequireHttps from '@wix/wix-express-require-https';
+import { AppContext } from './types';
 
-module.exports = (app: Router, context: AppContext) => {
+export = (app: Router, context: AppContext) => {
```
ronami: Do we have a dedicated type for the express function?

For example:

```ts
type ExpressFunction = (app: Router, context: AppContext): Router
```

This way all arguments can be annotated and the return value instead of annotating each one (currently the return value isn't declared). Same suggestion for the `config` function.

PR #1693, opened 2019-11-18
Temporary webpack.config.ts work around for thunderbolt-elements 

hadarge:
### 🔦 Summary
Temporary added `thunderbolt-elements` to the whitelisted packages to be transpiled by Yoshi into the server bundle


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -832,6 +832,8 @@ export function createBaseWebpackConfig({
                 reAssets,
                 /bootstrap-hot-loader/,
                 /yoshi-server/,
+                // Temporary walk around for thunderbolt
```
ronami:
```suggestion
                // Temporary work-around for thunderbolt
```

PR #1696, opened 2019-11-19
Flow editor template

deanshub:
### 🔦 Summary
creating a new template for flow-editor instead of working on the ooi template


packages/create-yoshi-app/src/templates.ts
```diff
@@ -19,4 +19,11 @@ const templates: Array<TemplateDefinition> = [
   },
 ];
 
+if (process.env.EXPERIMENTAL_FLOW_EDITOR) {
+  templates.push({
+    name: 'editor-flow',
```
ranyitz: Our convention is `flow-editor`

PR #1707, opened 2019-11-21
entry points with dynamic naming

deanshub:
### 🔦 Summary
entry points with dynamic naming, will allows multiple ooi widgets


packages/create-yoshi-app/templates/flow-editor/javascript/__tests__/viewerApp.e2e.js
```diff
@@ -1,6 +1,6 @@
 const { viewerUrl } = require('../dev/sites');
 
-describe('Viewer App', () => {
+describe.skip('Viewer App', () => {
```
ranyitz: Why is this skipped?

packages/create-yoshi-app/templates/flow-editor/javascript/{%packagejson%}
```diff
@@ -11,10 +11,10 @@
     "start": "yoshi-flow-editor start --entry-point=dist/server.js",
     "precommit": "lint-staged",
     "pretest": "yoshi-flow-editor build",
-    "test": "yoshi test --jest",
+    "test": "yoshi-flow-editor test --jest",
```
ranyitz: `--jest` should be the default (so no need to add it)

packages/yoshi-flow-editor/config/wallaby-common.js
```diff
@@ -0,0 +1,89 @@
+const { spawnSync } = require('child_process');
+const { isAbsolute } = require('path');
```
ranyitz: Does this flow need a specific wallaby config? I think we can start without adding this concern. If we really want we can bring it from `yoshi-flow-legacy` (But I don't think it's important in this stage)

packages/yoshi-flow-editor/config/wallaby-jest.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function(wallaby) {
+  const wallabyCommon = require('./wallaby-common')(wallaby);
+  // We need to add a wallaby compiler, because apparently wallaby does not
```
ranyitz: Same as above (we can remove this file)

packages/yoshi-flow-editor/package.json
```diff
@@ -26,6 +26,7 @@
     "yoshi-helpers": "4.22.0"
   },
   "devDependencies": {
+    "@babel/core": "^7.7.4",
```
ranyitz: Why is this needed?

packages/create-yoshi-app/templates/flow-editor/javascript/__tests__/viewerApp.e2e.js
deanshub: fixed in the next PR

packages/yoshi-flow-editor/package.json
deanshub: don't need it no more 👍

PR #1708, opened 2019-11-21
Correctly forward template e2e tests' output

amitdahan:
### 🔦 Summary
Currently we only know that `tsc --noEmit --strict` failed, we do not know why.
[**Example Build**](http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=5504664&buildTypeId=Yoshi_TestTemplatesBm&tab=buildLog#_state=3240,3588,1217)
This *should* fix that.


packages/create-yoshi-app/scripts/e2e.js
```diff
@@ -70,15 +70,23 @@ const testTemplate = mockedAnswers => {
...
+        try {
+          const { all: tscOutput } = execa.sync(
+            './node_modules/.bin/tsc --noEmit --strict',
+            {
+              shell: true,
+              cwd: testDirectory,
+            },
+          );
+
+          console.log(tscOutput);
```
ronami: not needed

amitdahan:
#### Update:
We discovered the [incapabilities of `execa.sync`](https://github.com/sindresorhus/execa/issues/391) of returning the `all` output property.
I refactored both `execa` usages to do the same thing, which is return the `execa` result (dropped the `sync`) regardless of the process failing or not, and add a simple assertion for what we actually meant.
The new logic defaults to outputting everything *if* the process actually failed, regardless all outputs are returned in the result object.

amitdahan:
I'll revert the typechecking breaking commit after everything else passes correctly and I see the error in CI.

amitdahan:
`npm test` failures are swallowed. Fixing.

packages/create-yoshi-app/scripts/e2e.js
```diff
@@ -68,43 +82,21 @@ const testTemplate = mockedAnswers => {
...
-            this.message = `\n  ${error.message}\n\n  stdout: ${error.stdout}\n  stderr: ${error.stderr}`;
-            this.stack = '';
-          }
-        }
-
-        throw new NpmTestFailureError();
-      }
+      const result = await exec('npm test');
 
+      expect(result.exitCode).toBe(0);
```
ranyitz: Add a failure reason

amitdahan: Seems it's only possible in jest using a plugin. Wdyt?
https://github.com/facebook/jest/issues/1965

PR #1711, opened 2019-11-22
Use Node 12

amitdahan:
### 🔦 Summary
Use Node 12 since it's LTS now and node-platform will switch to it Dec 2nd.
We need to decide whether we want to merge this beforehand. 
I know the node-platform uses `.nvmrc` to test but they asked users to start testing and building their artifacts with 12 beforehand.
Regardless, this can be merged Dec 2nd.


amitdahan:
Migration email mentioned updating the `Dockerfile`s as well, from `stable` to `12`. Wdyt? Should we keep it at `stable` and wait for it to point at `12`? Or update to `12` and revert to `stable` once it's shifted?

amitdahan:
Should this be a breaking change since we'll technically stop supporting yoshi on node 10?

hugebdu:
@ranyitz since this applies only to new artifacts IMHO you can merge and not wait

amitdahan:
@ranyitz so should this be marked a breaking change? I don't think we need to up the major, but `yoshi` will essentially ditch "support" for node 10 with this one.

ranyitz:
@amitdahan It's not a breaking change as we haven't actually changed anything in our user facing code (and `create-yoshi-app` is only for new users) we're just stop testing our code on node `10`.

Currently there are two things in our tests which are broken:
1. A test in the old integration tests for mocha and for jasmine: `should not transpile tests if 'transpileTests' is 'false'`
2. `npm install` of ooi template as well as the new editor flow, @deanshub It would be nice if you could check it out.

deanshub:
issue seems related to #1722 can't find module `yoshi-common/build/telemetry` , checking

deanshub:
should be resolved when #1726 is merged

PR #1718, opened 2019-11-24
Use transpiled JavaScript in the yoshi-flow-editor's templates (instead of jsx)

deanshub:
### 🔦 Summary
fixing failing build (not transpiling yoshi flow editor)


ranyitz:
BTW, I know that this was the situation before, I just think that even if we choose to stick with the hack it worth changing the explanation comment above.

PR #1720, opened 2019-11-25
resolve bootstrap env vars according to cwd for monorepo support

roymiloh:
We have to make sure that bootstrap env vars are relative to the pkg we're running the `start` command on.


ranyitz:
image

You've got me 😉

PR #1727, opened 2019-11-27
resolving telemetry dependency

deanshub:
### 🔦 Summary
small fix for resolving telemetry dependency related to #1722


packages/yoshi-common/telemetry.d.js
```diff
@@ -0,0 +1 @@
+export { default } from './build/telemetry';
```
ronami: Should be `*` instead of `{ default }` since we have a named export.

PR #1730, opened 2019-11-27
Stop telemetry until we have our own custom server

ronami:
### 🔦 Summary

Since Google Analytics doesn't provide support for filtering/aggregating the data we should look into having our own custom server to collect the data. Stopping telemetry until we have an idea of how it should look like.


deanshub:
😢 
So are we going to use mongo instead?

PR #1748, opened 2019-12-08
generating ci config

deanshub:
### 🔦 Summary
generating .ci_config file


packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,28 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+export function generateCiConfig(
+  appDefId: string,
+  widgetName: string,
```
ranyitz: appName

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,28 @@
...
+  baseUrl: string,
+  subWidgets: Array<{ id: string; name: string }>,
```
ranyitz: Widgets

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,28 @@
...
+) {
+  const ciConfig = {
+    app_def_id: appDefId,
+    ignore_dependencies: 'clear',
+    tpa_url_templates: [
+      `appFields.platform.baseUrls.${widgetName}BaseUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/`,
+      "widgets[?(@.widgetId=='13afb094-84f9-739f-44fd-78d036adb028')].componentFields.controllerUrl=>https://static.parastorage.com/services/wixstores-client-gallery/{version}/galleryController.bundle.min.js",
```
ranyitz: Hard coded

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,28 @@
...
+      ...subWidgets.reduce((res: Array<string>, { id, name }) => {
```
ranyitz: widgets

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,28 @@
...
+        return [
+          ...res,
+          `widgets[?(@.widgetId=='${id}')].componentFields.controllerUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/${name}.bundle.min.js`,
+          `widgets[?(@.widgetId=='${id}')].componentFields.controllerUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/${name}Controller.bundle.min.js`,
+        ];
+      }, []),
+    ],
+  };
+  console.log(path.join(process.cwd(), '.ci_config'));
```
ranyitz: non relevant

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  fs.outputFileSync('.ci_config', JSON.stringify(ciConfig, null, 2));
```
ranyitz: fs.outputJson

ranyitz: `.ci_config` -> put in constants

packages/yoshi-flow-editor/src/commands/build.ts
```diff
@@ -74,6 +75,8 @@ const build: cliCommand = async function(argv, config) {
   await copyTemplates();
 
   if (inTeamCity()) {
+    // TODO: make the parameters dynamic
+    generateCiConfig('', 'todo', 'wixstores-client-todo', []);
```
ranyitz: Let's make those parameters dynamic before merging this PR

packages/yoshi-flow-editor/src/constants.ts
```diff
@@ -0,0 +1,3 @@
+export enum FileNames {
```
ranyitz: Is there a benefit of using `enum` over exporting named properties?

for example:
```ts
export const CI_CONFIG = '.ci_config';
```

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,51 @@
...
+    .concat(
+      widgets.map(widget => generateWidgetControllerUrl(widget, baseUrl)),
+    );
+}
+
+function generateWidgetControllerUrl(
+  { id, name }: Widget,
+  baseUrl: string,
+): string {
+  return `widgets[?(@.widgetId=='${id}')].componentFields.controllerUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/${name}Controller.bundle.min.js`;
```
ranyitz: I'm not sure that that's the case for all apps. Not everybody uses `controllerURL` yet. So I think that it wouldn't change anything for apps that has a single `viewerScript`.

Also, for those kinds of apps, you need to also upgrade their `viewerScript` version.

packages/yoshi-flow-editor/src/constants.ts
deanshub: nope, just makes it easier to understand that these are file names

ranyitz: We use this convention to describe file names / directory names - https://github.com/wix/yoshi/blob/master/packages/yoshi-config/src/paths.ts#L16-L19

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
deanshub: added both methods just in case

deanshub:
1. create type for the controller config by moving it to native components infra
2. remove empty ctrl (not needed because we throw an error on build)
3. remove platform config file (move them to named exports for appdefid, and support default separate controller pattern)
4. minimize glue code in viewer script wrapping (without controller factory)

Issue #1742, opened 2019-12-04
Separate entries generate identical classnames when using shared components and css modules


jjaskulis:
# 🐛 Bug Report

### 😯 Current Behavior

We have two OOI TPAs with defined separate entries in Yoshi config. These two TPA is our main app forum and it's widget(where recent posts are displayed).
It's possible to display widget and forum on the same site page or use them independently (have forum on one page and widget on the other one). Our problem appears only when using widget and forum on the same page. I have to mention that widget imports and uses components from the main forum app. Even though these two have separate entries and css bundles, classes generated for the components used in both TPA are identical. This presents unpredictable styling overrides. Let's say forum app uses component which has base styling class and some extended styling class, which overrides some of the base class properties.
Widget uses component, but only with base styles. Forum css bundle is loaded first, then the widget css. As css specificity is the same, the properties from class which was loaded last for the component are applied(in this case, the base styling from widget):

COMPONENT baseClass (from TPA1) extendedClass (from TPA1)
COMPONENT baseClass (from TPA2)

Here any properties from extendedClass which have been overridden properties from baseClass are reverted back to base styling.

### 🤔 Expected Behavior

Each entry should generated unique css classes.

### 🔦 Context 

...

### 💁 Possible Solution 
Generate unique classes per entry, adding unique hashes or unique random seed.

### 👣 Steps to Reproduce 

Please check the repository for example where this issue is reproduced.
About the repository setup:
Two entry points are specified in yoshi config, app1 and app2.
app1 has and ExampleButton component with some default styles described.
app2 imports ExampleButton and adds additional styles(different border color and padding), which override the default styles.
Styling works as expected, if app2 is loaded in isolation.
If both apps are loaded, app1 overrides the css and button styling reverts to base. (to simulate the situation where both apps are loaded on the same page, i’ve modified the index.ejs css from app1 and app2 are loaded)

### 💻 Code Sample

https://github.com/wix-a/example-classnames

### 🌍 Your Environment

System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  Binaries:
    Node: 12.7.0 - ~/.nvm/versions/node/v12.7.0/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.10.0 - ~/.nvm/versions/node/v12.7.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
[...]

PR #1750, opened 2019-12-10
updating css-loader

deanshub:
### 🔦 Summary
updating [css-loader](https://github.com/webpack-contrib/css-loader) to version 3

(also helps with #1742)


deanshub:
yup

PR #1751, opened 2019-12-10
Construct CDN URL according to the package's `experimentalBuildHtml`

ronami:
### Summary

This PR addresses a problem for monorepo flow where the root config for `experimentalBuildHtml` affects the CDN URL construction instead of the package's config.


packages/yoshi-helpers/src/utils.ts
```diff
@@ -246,12 +246,14 @@ export const getProjectArtifactVersion = () => {
 };
 
 // Gets the CDN base path for the project at the current working dir
-export const getProjectCDNBasePath = () => {
+export const getProjectCDNBasePath = (
+  useUnversionedBaseUrl = config.experimentalBuildHtml,
```
ronami: I had to keep this default since this function gets called by https://github.com/wix/yoshi/blob/construct-correct-cdn-url/packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/cdnProxy.js#L9.

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -285,6 +285,7 @@ export function createBaseWebpackConfig({
 
   const publicPath = calculatePublicPath({
     devServerUrl,
+    useUnversionedBaseUrl: contentHash,
```
ranyitz: useContentHash

PR #1753, opened 2019-12-11
distinguish between apps and libs by projectType config

roymiloh:
Up until now, we choose the build flow for apps and libs by their `private` prop in `package.json` file (apps should be marked as private; libs not). We're now having a situation where we want to publish apps as well, but still continue with the app flow. Hence, this PR changes the way we decide if a module is an app or lib by moving from `private` property to explicit property inside yoshi config.


packages/yoshi-config/src/config.ts
```diff
@@ -2,7 +2,7 @@
 import { Entry, EntryFunc, ExternalsElement, Options } from 'webpack';
 import { PackageJson } from 'read-pkg';
 
-type ProjectType = 'app';
+type ProjectType = 'app' | 'lib';
```
ranyitz: (nitpick...)

```suggestion
type ProjectType = 'app' | 'library';
```

packages/yoshi-flow-monorepo/src/load-package-graph.ts
```diff
@@ -59,7 +59,10 @@ export default async function loadPackageGraph(): Promise<LoadGraphResult> {
 
   const pkgsArray = Array.from(graph.values());
 
-  const [apps, libs] = partition(pkgsArray, node => node.pkg.private);
+  const [apps, libs] = partition(
+    pkgsArray,
+    node => node.config.projectType === 'app',
```
ranyitz: It would be nice to throw an error for unknown project type.

@roymiloh What do you think?

PR #1755, opened 2019-12-11
[create-yoshi-app] upgrade `lint-staged` for templates to `9.5.0`

sidoruk-sv:
### 🔦 Summary
Upgrade `lint-staged` to v`9.5.0`

Also replaced advance configuration with simple configuration (minimatch string), because advanced was deprecated in v9: https://github.com/okonet/lint-staged/releases/tag/v9.0.0


sidoruk-sv:
Any objections to this PR?

sidoruk-sv:
@ranyitz sure) also seems tricky to me

As an option to test micromatch pattern `/{!(*.d).ts,*.{js,tsx,scss,less}}` I used:
https://globster.xyz/

With file structure:
```
/code.ts
/code.tsx
/code.js
/code.d.ts
/styles.scss
/styles.less
```

As a proposal, for now, maybe it is better not to ignore `*.d.ts` files at all.
Can something go wrong if `*.d.ts` would be also formatted?

ranyitz:
@sidoruk-sv I thought about the same thing, I'm not sure why we exclude `*.d.ts` files. @amitdahan @deanshub @ronami @yanivefraim maybe one of you remember why we're excluding `d.ts` files from lint-staged or have an opinion about it?

sidoruk-sv:
@ranyitz found such an example of running prettier over d.ts files:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/.prettierrc.json (seems like all `DefinitelyTyped` packages are formatted with `prettier`)

(from this message: https://github.com/prettier/prettier/issues/6215#issuecomment-506559776)

ranyitz:
If we don't have any objections, i'm ok with running prettier over `d.ts` files. @sidoruk-sv Let's change this PR to not ignore `d.ts` files and let's merge.

____________

**A note on the side...**
I think that the biggest problem here is that we don't have any abstraction over git hooks, so every time we want to make changes/upgrades we only do it to new users. Users that already generated projects in the past won't get our upgrades and If we've made a mistake it would be hard to resolve. Also, we don't have any immediate feedback after making the change.

sidoruk-sv:
@ranyitz tuned, excluding of `*.d.ts` eliminated

PR #1758, opened 2019-12-15
Editor Elements fix - Transpile node-modules only for TB App

NitayRabi:
### 🔦 Summary
When running build (Storybook/Yoshi) in `editor-elements` mono-repo, `babel-loader` transpiles `node_modules`, causing issues (e.g Storybook doesn't start with 
`Error: exports is not defined` in the browser) and slowing the build.

The issue is caused due to the specific solution made for TB-App.


ronami:
Please fix linting and let's merge.

PR #1759, opened 2019-12-15
disable chunks on server bundle on thunderbolt-elements only

roymiloh:
There's no reason to split server bundle into chunks, as it just complicates things. We first try it on `thunderbolt-elements` package, and we'll extend it to everyone in the future.


packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -122,6 +122,11 @@ export function createServerWebpackConfig(
 
   if (customThunderboltElements) {
     serverConfig.output!.path = path.join(pkg.location, STATICS_DIR);
+    serverConfig.plugins!.push(
+      new webpack.optimize.LimitChunkCountPlugin({
```
ranyitz: ❤️

PR #1632, opened 2019-10-31
Initial suricate support implementation

netanelgilad:
### 🔦 Summary
Initial implementation for supporting `suricate` in `yoshi`.

### 🗺️ Test Plan
...

PR #1764, opened 2019-12-17
Integrate Suricate (tunneling)

ranyitz:
This PR is the continuation of #1632 

------------

It adds integration with @netanelgilad's tunneling [suricate](https://github.com/wix-private/suricate) service.

The purpose of the PR is to support local development (`yoshi start`) against production leveraging tunneling, you can look at a similar [tool](https://ngrok.com/) for more information.

### How to use?

In order to opt in to the feature, you need to add `{ suricate: true }` to the yoshi config.

### Implementation details 

The integration spread in many places in the codebase, i'll explain the integration in details:

1. **App-server** - We open a socket before we initialize the user's app server. 
2. **Webpack-dev-server** - We open another socket with a different tunnnelID for the static assets.
3. **Public Path** - Public path should be changed from `localhost:3200` to the cdn suricate path from above.
4. **Log Tunneled Urls to the console**
5. **Open the browser on the tunneled app server url**

### Still needs to be done:

* Verify the way HMR works with tunneling
* See why/if we need to configure `cors` specifically for suricate.
* Consider changing the strategy of generating the tunnel Ids or moving it to suricate.
* Add tests - @netanelgilad said he will expose a testkit so we'll be able to run suricate server locally for `e2e` testing of the feature. Maybe we can also test it on production, what do you say?

### important notes

This feature is not ready for mass usage, in order to work, it needs to be combined with the `exerimentalBuildHTML`. This dependency is due to the fact that the `cdn` url decided there and we use the `publicPath` in order to inject the tunnled one. (otherwise the requests will happen to `localhost` instead)

* **Note:** The app server & cdn will still be availble on their original ports on localhost.

______________

In order to prevent code duplications this PR also refactors parts of the codebase like changing variable names and also move all the duplicated loggers into one logger configured in the dev environment, in case we would like to create custom logger it will always be possible.


packages/yoshi-common/src/dev-environment.ts
```diff
@@ -16,14 +23,18 @@ type WebpackStatus = {
...
+
+export type State =
   | {
       status: 'compiling';
     }
   | ({
       status: 'success';
       serverUrls: Urls;
       devServerUrls: Urls;
+      suricate: boolean;
```
ranyitz: Do not put on state, instead move to logger function

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -33,15 +44,24 @@ export default class DevEnvironment {
...
 
   constructor(
     webpackDevServer: WebpackDevServer,
     serverProcess: ServerProcess,
     multiCompiler: webpack.MultiCompiler,
+    appName: string,
+    suricate: boolean = false,
+    startUrl?: StartUrl,
   ) {
     this.webpackDevServer = webpackDevServer;
```
ranyitz: Move to be an options object

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -75,7 +95,9 @@ export default class DevEnvironment {
 
         this.store.setState({
           status: 'success',
+          appName: this.appName,
```
ranyitz: move to logger function call

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -217,11 +253,15 @@ export default class DevEnvironment {
         throw new Error('client webpack config was created without an entry');
       }
 
+      const socketForHmr = suricate
```
ranyitz: socketForHmr => hmrSocketPath

packages/yoshi-common/src/utils/suricate.ts
```diff
@@ -0,0 +1,53 @@
...
+export const getUrl = (namespace: string) =>
+  `${suricateURL}/tunnel/${getTunnelId(namespace)}/`;
+
+export const getDevServerUrl = (appName: string) =>
+  getUrl(`${appName}-dev-server`);
+
+export const createDevServerSocket = (appName: string, port: number) =>
+  createSocket(`${appName}-dev-server`, port);
+
+//https://bo.wix.com/suricate/tunnel/rany.@wix-generated/?sockPath=/suricate/tunnel/rany.@wix-generated/sockjs-node
```
ranyitz: remove

packages/yoshi-common/src/webpack-dev-server.ts
```diff
@@ -81,9 +88,16 @@ export class WebpackDevServer extends OriginalWebpackDevServer {
         '.ooidev.com',
         '.deviantart.lan',
       ],
+      disableHostCheck: true,
```
ranyitz: `true` only if suricate is `true`

packages/yoshi-common/src/webpack-dev-server.ts
```diff
@@ -81,9 +88,16 @@ export class WebpackDevServer extends OriginalWebpackDevServer {
...
       before(expressApp) {
         // Send cross origin headers
-        expressApp.use(cors());
+        // TODO - support only suricate URL
+        expressApp.use(
+          cors({
```
ranyitz: use only for suricate

packages/yoshi-config/src/normalize.ts
```diff
@@ -27,7 +27,7 @@ export default (initialConfig: InitialConfig, pkgJson: PackageJson): Config => {
     : singleDir;
 
   const config: Config = {
-    name,
+    name: name!,
```
ranyitz: not needed

packages/yoshi-config/src/normalize.ts
```diff
@@ -81,6 +81,7 @@ export default (initialConfig: InitialConfig, pkgJson: PackageJson): Config => {
 
     isAngularProject: !!dependencies.angular || !!peerDependencies.angular,
     isReactProject: !!dependencies.react || !!peerDependencies.react,
+    suricate: !!initialConfig.suricate,
```
ranyitz:
```suggestion
    suricate: initialConfig.suricate,
```

packages/yoshi-flow-bm/package.json
```diff
@@ -15,10 +15,10 @@
     "react-dev-utils": "9.1.0",
     "webpack": "4.41.2",
     "webpack-node-externals": "1.7.2",
-    "yoshi-common": "4.25.1",
     "yoshi-config": "4.23.3",
     "yoshi-flow-bm-runtime": "4.24.0",
     "yoshi-flow-legacy": "4.25.1",
+    "yoshi-common": "4.25.1",
```
ranyitz: revert

packages/yoshi-config/src/normalize.ts
ranyitz: It is needed as the type comes from `read-pkg` is optional

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -264,21 +308,30 @@ export default class DevEnvironment {
...
+      startUrl,
+    });
 
     devEnvironment.startServerHotUpdate(serverCompiler);
 
     if (webWorkerCompiler) {
       devEnvironment.startWebWorkerHotUpdate(webWorkerCompiler);
     }
 
+    devEnvironment.store.subscribe((state: State) =>
```
ronami: I don't think it's needed, should be deduced from the `devEnvironment.store.subscribe()` call.

PR #1765, opened 2019-12-17
HtmlWebpackPlugin: Inline CSS chunks / don't inject chunks

ronami:
### Summary

This PR makes changes to how [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin) is being used in Yoshi.

Since Thunderbolt is the only project using it, the changes being introduced are specific to Thunderbolt's use-case. While specific, I still think that many of these changes will be relevant to other projects once they start using this.

This PR includes the following changes:

#### Inline chunks support

Add support for inlining selected chunks into the initial HTML via a new Webpack plugin: `InlineChunkHtmlPlugin`.

I used [Create React App's plugin](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/InlineChunkHtmlPlugin.js) and enhanced it with support for `style` chunks.

This is especially important for server-rendered apps (See [1](https://kinsta.com/blog/eliminate-render-blocking-javascript-css/#optimize-delivery-of-css-resources), [2](https://developers.google.com/speed/docs/insights/OptimizeCSSDelivery) and [3](https://web.dev/extract-critical-css)). We can also use this as a way of inlining specific chunks like the BI/Beat chunk that is way too small to warrant an additional HTTP request.

#### Stop injecting JS/CSS tags

Configure [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin) to not inject JS/CSS assets and instead allow finer control (placeholders). This is especially important for Thunderbolt's use-case that is significantly more complex.

Since we transpile templates once in build time and then a second time by the developer's server in runtime, I added a custom loader with custom tags (`{%` instead of the default `<%`) so it's clear which tags are being used for build time and which are for runtime.

I think it's better to have Yoshi use custom (non-standard) tags so most users can use EJS without having to configure custom tags on their own. To do this I copied the 50 lines-of-code loader from [html-webpack-plugin's default loader](https://github.com/jantimon/html-webpack-plugin/blob/master/lib/loader.js).

#### Don't inject polyfills

Remove polyfill injection since Thunderbolt will be built into multiple targets (another bundle specifically for older browsers).

It also makes sense since platforms will include polyfills once.

#### Don't map template name to a matching entry name

Remove the convention of automatically injecting all initial chunks of an entry to a template with the same name (See #1090) and instead allow projects to use their own logic to filter chunks from/to EJS templates.

---

When we decide to roll this feature out for other projects we should probably consider better defaults for Thunderbolt's custom and more complex logic.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -356,6 +355,8 @@ export function createBaseWebpackConfig({
         // Search in `yoshi-common`'s node_modules first
         path.join(__dirname, '../node_modules'),
 
+        path.join(__dirname, '../../../node_modules'),
```
ronami: Not related to the PR but since #1635 has been merged it's hoisting all deps so symlinking Yoshi while working on a different project doesn't work.

PR #1768, opened 2019-12-18
locking dependencies for flow editor

deanshub:
### 🔦 Summary
locking all dependencies in yoshi flow editor


ranyitz:
The PR has too many conflicting files, we can use it as a reference in the future

PR #1774, opened 2019-12-19
fix| ignore browser mainField when target:node

tompere:
### 🔦 Summary
@ronami and myself run into an issue when bundled server code (thunderbolt-becky) in target `web` and using isomorphic-unfetch which has `browser` entry in its package.json

### 🗺️ Test Plan
...


ronami:
Thanks!

Can you please push a lint fix (`y lint --fix`)?

This is probably not a breaking change as almost all projects that use `target: node` have all of their `node_modules` external.

PR #1776, opened 2019-12-22
`jest-environment-yoshi-puppeteer`: detect network errors + add a message for missing cdn server

### LINKED ISSUE #1134: Do not throw an error telling to run `npm start` when it's not actually needed.

yanivefraim:
fixes #1134 

Problem:
When running server e2e tests without running `npm start`, we receive an error telling us to run `npm start` when it's not actually needed.

Solutions:
Instead of checking for `webpackDevServerProcess` and throwing an error, we will catch network errors, and:
- In case the error is from `//localhost:3200` - add a specific message about missing cdn server (with details about the failing assets), telling the user to run `npm start` in a different terminal
- else: print the error (details about the failing asset)

### DIFF
--- a/packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/index.js
+++ b/packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/index.js
@@ -38,6 +38,27 @@ module.exports = class PuppeteerEnvironment extends ParentEnvironment {
       console.warn(`Puppeteer page error: ${error.message}`);
       console.warn(error.stack);
     });
+
+    this.global.page.on('requestfailed', request => {
+      if (request.url().includes('//localhost:3200')) {
+        console.warn(
+          `We found that some of your static assets failed to load:
+
+          url: ${request.url()}, errText: ${
+            request.failure().errorText
+          }, method: ${request.method()}
+
+          Please try running 'npm start' in another terminal in order to start your CDN server.
+          `,
+        );
+      } else {
+        console.warn(
+          `url: ${request.url()}, errText: ${
+            request.failure().errorText
+          }, method: ${request.method()}`,
+        );
+      }
+    });
   }
 
   async teardown() {
--- a/packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/index.js
+++ b/packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/index.js
@@ -35,13 +35,13 @@ module.exports = class PuppeteerEnvironment extends ParentEnvironment {
     this.global.page.setDefaultNavigationTimeout(10 * 1000);
 
     this.global.page.on('pageerror', error => {
-      console.warn(`Puppeteer page error: ${error.message}`);
-      console.warn(error.stack);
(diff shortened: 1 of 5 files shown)


packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/index.js
```diff
@@ -35,8 +35,29 @@ module.exports = class PuppeteerEnvironment extends ParentEnvironment {
...
 
     this.global.page.on('pageerror', error => {
-      console.warn(`Puppeteer page error: ${error.message}`);
-      console.warn(error.stack);
+      this.global.console.warn(`Puppeteer page error: ${error.message}`);
+      this.global.console.warn(error.stack);
+    });
+
+    this.global.page.on('requestfailed', request => {
+      if (request.url().includes('//localhost:3200')) {
```
ranyitz: unfortunately, `localhost:3200` is a configurable value, which means we'll need to read it out of `yoshi-config` if we want to capture all use-cases.

yanivefraim: ooops, I have no idea how I missed this one.

This makes me think, we can also have an error from other static assets, under cdn address, which is not 100% related to missing a cdn server... Maybe we should look for the error itself, or, check if it is one of our entry points / css files?
WDYT?

ranyitz: I think that it worth checking.

the [`requestFailed`](https://pptr.dev/#?product=Puppeteer&version=v2.0.0&show=api-event-requestfailed) event would occur in case of a timeout request. while in case there is a wrong URL you should get a `404` response which is a `requestfinished` event. So the current heuristic sounds good to me.

I don't think that we'll have many false positives (tell the user to run `npm start` while in reality there is a different problem) but again, it worth checking that the cases that you describe are working as expected.

yanivefraim:
closing in favor of #1885

PR #1777, opened 2019-12-23
Generate CI config in editor flow

deanshub:
### 🔦 Summary
generating .ci_config file in ci for the dev-center


packages/yoshi-flow-editor/src/model.ts
```diff
@@ -7,85 +7,87 @@ export interface FlowEditorModel {
...
+      };
+
+      const widgetFileName = checkIfExists('Widget');
+      const pageFileName = checkIfExists('Page');
+      const controllerFileName = checkIfExists('controller');
+      const settingsFileName = checkIfExists('Settings');
+
+      if (!controllerFileName) {
+        throw new Error(`Missing controller file for the component in "${componentDirectory}".
+        Please create "controller.js/ts" file in "${path.dirname(
```
ranyitz: I'm not sure that `path.dirname` is necessary here.

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -7,85 +7,87 @@ export interface FlowEditorModel {
...
+          componentDirectory,
+        )}" directory`);
+      }
+      if (!widgetFileName && !pageFileName) {
+        throw new Error(`Missing widget or page file for the component in "${componentDirectory}".
+        Please create either Widget.js/ts or Page.js/ts file in "${path.dirname(
+          componentDirectory,
```
ranyitz: I think that `dirname` would result in upper level directory

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -7,85 +7,87 @@ export interface FlowEditorModel {
...
+        type: widgetFileName ? 'widget' : 'page',
+        controllerFileName,
+        settingsFileName,
+        // TODO: import from named export
+        id: '',
+      };
+    },
+  );
+
+  const [initApp] = await globby('./src/components/initApp.js', {
```
ranyitz: This is not a glob pattern, you can just `resolveFrom` the file.

btw, In the future, I guess that we would like a smarter resolver that checks for `ts` first and than `js` but it's totally ok to start simple.

packages/yoshi-flow-editor/src/commands/build.ts
```diff
@@ -74,10 +74,12 @@ const build: cliCommand = async function(argv, config) {
   ]);
 
   await copyTemplates();
-  const model = generateFlowEditorModel();
+
+  const model = await generateFlowEditorModel();
```
ronami: What do you think about moving the model generation to the bin file? I think that most commands (build/start) will need access to it and I think it's somewhat cheap to create in terms of performance.

I did something similar in monorepo flow since I need to know which packages exist in a monorepo for most commands to operate: https://github.com/wix/yoshi/blob/master/packages/yoshi-flow-monorepo/src/bin/yoshi-monorepo.ts#L13. Then, every command can get this model on top of the config and the CLI arguments (https://github.com/wix/yoshi/blob/master/packages/yoshi-flow-monorepo/src/bin/yoshi-monorepo.ts#L13).

packages/yoshi-flow-editor/src/editorAppWrapping.ts
```diff
@@ -10,10 +10,7 @@ const componentWrapper = (
   model: FlowEditorModel,
 ) => {
   return model.components.reduce(
-    (
-      acc: Record<string, string>,
-      component: FlowEditorModel['components'][0],
-    ) => {
+    (acc: Record<string, string>, component: ComponentModel) => {
```
ronami: I think we have some duplicated code that iterates over the components, writes an entry for each into a tmp directory and then returns the list of files.

I think we can adjust it to something simpler: Each wrapping file (`settingsWrapping`, `viewerScriptWrapping`, etc) will export a function that will get a component and return the path to the entry file along with its content.

Wdyt? Does it make sense?

packages/yoshi-flow-editor/src/commands/start.ts
```diff
@@ -83,7 +83,7 @@ const start: cliCommand = async function(argv, config) {
     fs.emptyDir(join(TARGET_DIR)),
   ]);
 
-  const model = generateFlowEditorModel();
+  const model = await generateFlowEditorModel();
 
   const clientConfig = createClientWebpackConfig(config, {
```
ronami: Not related to the PR but I think we should pass the model to our functions that create the webpack configs so they can call functions like `buildEditorPlatformEntries()` internally. I think it makes sense since it’s part of the Webpack config (will help with watch mode later) and it will also remove some code duplication between build/start. Wdyt?

packages/yoshi-flow-editor/src/model.ts
ronami: We can use [resolve()](https://github.com/browserify/resolve#resolveid-opts-cb) as it gets an `extensions` option.

packages/yoshi-flow-editor/src/editorAppWrapping.ts
deanshub: will do next PR

packages/yoshi-flow-editor/src/commands/start.ts
deanshub: will do next PR

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,49 @@
...
+    generateWidgetControllerUrl(component, baseUrl),
+  );
+
+  return controllerConfig.concat(
+    components.map(component => generateWidgetComponentUrl(component, baseUrl)),
+  );
+}
+
+function generateWidgetControllerUrl(
+  { id, name }: FlowEditorModel['components'][0],
```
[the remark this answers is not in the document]
deanshub: good catch

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+function generateWidgetComponentUrl(
+  { id, name }: FlowEditorModel['components'][0],
+  baseUrl: string,
+): string {
+  return `widgets[?(@.widgetId=='${id}')].componentFields.componentUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/${name}.bundle.min.js`;
+}
+
+export function writeCiConfig(model: FlowEditorModel): Promise<void> {
+  const appName = model.components[0].name;
```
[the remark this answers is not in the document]
deanshub: appName, replaced it with the package name instead of the first component

packages/yoshi-flow-editor/src/commands/ciConfigGenerator.ts
```diff
@@ -0,0 +1,49 @@
...
+  baseUrl: string,
+): string {
+  return `widgets[?(@.widgetId=='${id}')].componentFields.controllerUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/${name}Controller.bundle.min.js`;
+}
+
+function generateWidgetComponentUrl(
+  { id, name }: FlowEditorModel['components'][0],
+  baseUrl: string,
+): string {
+  return `widgets[?(@.widgetId=='${id}')].componentFields.componentUrl=>https://static.parastorage.com/services/${baseUrl}/{version}/${name}.bundle.min.js`;
```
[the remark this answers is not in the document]
deanshub: good catch, for now I changed the string, in the clean up PR I'll add this change as well

PR #1779, opened 2019-12-24
moving translations to the controller

### LINKED ISSUE #1778: using translations in ooi breaks SSR
# 🐛 Bug Report

### 😯 Current Behavior

SSR fails on ooi templates created by `create-yoshi-app`

### 🤔 Expected Behavior

SSR should work as expected

### 🔦 Context 

usage of the translation files is by async import from the component side of the project
https://github.com/wix/yoshi/blob/39077e900e8fc0e5a62feca5a8b0049ab6ddfea2/packages/create-yoshi-app/templates/out-of-iframe/typescript/src/config/i18n.ts#L17
platform SSR doesn't know how to handle these async imports on the server, so it breaks the SSR

### 💁 Possible Solution 

The best practice from the platform states that the worker should request the translations using `fetch` and if processing is needed than also do it on the worker side (viewerscript\ controller)

deanshub:
### 🔦 Summary

moving translation usage to the worker in ooi projects
closes #1778


packages/create-yoshi-app/templates/out-of-iframe/typescript/src/components/Widget/appController.ts
```diff
@@ -23,12 +30,60 @@ async function getExperimentsByScope(scope: string) {
...
 export async function createAppController(
   controllerConfig: IWidgetControllerConfig,
 ): Promise<IWidgetController> {
   const { appParams, setProps } = controllerConfig;
   const language = getSiteLanguage(controllerConfig);
   const experiments = await getExperimentsByScope(EXPERIMENTS_SCOPE);
+  const fields = getMultiLangFields(
+    controllerConfig.wixCodeApi.window.multilingual,
+  );
+  const translations = await getTranslations(
```
tuchk4: Do we have any benchmarks/tests on how such `await` usage may affect on TTI?

tuchk4: Is it possible to load experiments and translations in parallel?

deanshub: I'll search for it but now I'm not changing anything since translation fetching was already in component, I just moved it here

packages/create-yoshi-app/templates/out-of-iframe/typescript/src/config/translationHoc.tsx
```diff
@@ -0,0 +1,55 @@
...
+): TranslationFunction {
+  return (translationKey: string, values: {}) => {
+    return translations[translationKey] === undefined
+      ? translationKey
+      : translate(translations[translationKey], values);
+  };
+}
+
+export const withTranslations = (
+  Component: any,
```
tuchk4: Much better to use generic types instead of `any`

packages/create-yoshi-app/templates/out-of-iframe/typescript/src/components/Widget/appController.ts
tuchk4: yeah, my question is not so related to this PR, I just curious that we use `await` in such places. So we slow-down rendering start

deanshub: I didn't find any benchmarks/tests, since this is an integration with the viewer architecture, maybe the platform guys have it?

deanshub:
We decided not to merge this PR for complexity reasons, this PR will still be a how-to reference and the best practice will be implemented in `yoshi-flow-editor` behind a simple to use api.

Thanks @tuchk4 for the awesome feedback, really appreciate it. Keep up the good work!

PR #1781, opened 2019-12-25
fix eslint not finding any file

eranshabi:


ranyitz:
@eranshabi Thanks for opening a PR, and sorry that it took so long to get to it, can you explain what didn't work in the current syntax?

eranshabi:
it seems that with the current pattern this line:
https://github.com/wix/yoshi/blob/b183c1923ad2fbe3c18447e31a9e3e966567d1b3/packages/yoshi-flow-legacy/src/tasks/eslint.js#L5

doesn't find anything (return empty array)
I'm not sure why it works in yoshi tests, but when running `yoshi lint` it doesn't work

ranyitz:
@eranshabi I created a repo, where i've changed one file and ran `npx yoshi lint`, everything worked as expected. Can you do the same, but show where it doesn't work as you expect?

image

eranshabi:
@ranyitz you ran tslint please change to eslint to see the bug
The minimal steps is to remove the `tslint.json` file and add this to package.json
```
 "eslintConfig": {
    "extends": "yoshi-base"
  }
```

You can see that now `npx yoshi lint` will fail silently (will just print nothing)
If you will run `eslint` natively with on the same files you will see the error `No files matching`
If you apply my fix you can see that now `npx yoshi lint` works

ranyitz:
@eranshabi 

I've tried to do exactly what you wrote and that's my result

image

I still can't recreate the problem.

Let's solve this, [Please include a repro](https://gist.github.com/Rich-Harris/88c5fc2ac6dc941b22e7996af05d70ff)

eranshabi:
@ranyitz here is a repo:
https://github.com/wix-a/eslint-bug

eranshabi:
it seems that having only `ts` files (`no `tsx`) is what causing the problem

ranyitz:
@eranshabi Thanks for creating the repro. I've cloned it and I can confirm there is a problem. Trying to understand the root cause, I've found out the `globby` does not identify the current and the suggested syntax.

image

A possible solution would be to create both `*.ts` and `*.tsx`:

```ts
globs.baseDirs.flatMap(dir => [`${dir}/**/*.ts`, `${dir}/**/*.tsx`])
```
__________________________

I think that the test is passing because it runs `tslint` (it has no `eslintconfig`)

@eranshabi what do you say?

eranshabi:
my test runs `eslint`:
image

Also when I apply the change to yoshi locally it works in the example repo

I checked and it's because yoshi uses a newer `globby` so it works. When you did `require('globby')` it used and older version required by some other module. I used `require('./node_modules/yoshi-flow-legacy/node_modules/globby')` and it works.

eranshabi:
@ranyitz please
This is the only thing between us and eslint typescript ⭐️

PR #1782, opened 2019-12-25
Support split chunks config in monorepos

ronami:
### 🔦 Summary

This is currently supported in other flows like [app-flow](https://github.com/wix/yoshi/blob/b6ce7cd0f1c8058c307b6bff7026bd5e41617012/packages/yoshi-flow-app/src/webpack.config.ts#L87) and [legacy-flow](https://github.com/wix/yoshi/blob/6de308885f901257796c1c749a42d067d8487c3f/packages/yoshi-flow-legacy/config/webpack.config.js#L92) and wasn't supported in monorepos because of a mistake.

Going forward, I think we should configure it with good defaults for all projects that opt-into `html-webpack-plugin`/`manifest` along with content hashing.


packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -99,6 +106,16 @@ export function createClientWebpackConfig(
   clientConfig.resolve!.alias = pkg.config.resolveAlias;
   clientConfig.externals = pkg.config.externals;
 
+  const useSplitChunks = pkg.config.splitChunks;
+
+  if (useSplitChunks) {
```
yanivefraim: I guess that we would have to think if we want to have some common code in one place, not sure about it.

PR #1784, opened 2019-12-25
Change when a chunk is inlined to be based on its name (.inline)

ronami:
### 🔦 Summary

This PR fixes a few mistakes in #1765 and changes when a chunk is inlined.

Instead of checking an asset's size, which doesn't work for most cases, it lets the app decide which chunks should be inlined by checking if a chunk's name ends with `.inline.js`.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -461,6 +461,8 @@ export function createBaseWebpackConfig({
                                     },
                                   };
                                 }
+
+                                return tag;
```
ronami: I accidentally filtered all inline tags 😢

PR #1788, opened 2019-12-26
Separate jest overrides supported type by project type.

### LINKED ISSUE #1507: Enable specific jest config overrides per project
# 🙋 Feature Request

Currently, there is a way (in yoshi's codebase) to configure the possible overrides for jest's project config.

See https://github.com/wix/yoshi/pull/1467 and https://github.com/wix/yoshi/pull/1506 for past work that has been done on the subject.

Those two overrides are meant to be configured only for the `component` project/environment. 

### 😯 Current Behavior

Today you can configure them also for the `e2e` project/environment.

### 🤔 Expected Behavior
It should be possible to configure an override for a specific project (`e2e`/`component`), the `testUrl` and `globals` options should be modified for the `component` project only.

### 🔦 Context

We believe that leaving a configuration option for an unverified use-case can lead to misuse of the product. We want to throw an error if something is misconfigured and by that, minimize mistakes.

NitayRabi:
### 🔦 Summary
Closes #1507 
- [x] Separate configuration by jest project override key
- [x] Will break users who use the now unsupported e2e keys. @ranyitz - WDYT?

For reference - 
1. https://github.com/search?q=org%3Awix-private+e2eOptions&type=Code
2. https://github.com/search?q=org%3Awix+e2eOptions&type=Code

(Seems like not used at all in e2e) 

### 🗺️ Test Plan
Tested via kitchensink tests.


packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -86,9 +91,29 @@ const config = {
...
 
         const projectValidOverrides = projectOverrides
-          ? pick(projectOverrides, supportedProjectOverrideKeys)
+          ? pick(
+              projectOverrides,
+              supportedProjectOverrideKeys[projectOverrideKey],
+            )
           : {};
 
+        if (
```
yanivefraim: No need to manually validate user's config. We are using [`jest-validate`](https://www.npmjs.com/package/jest-validate#api) which does the validation. See here https://github.com/wix/yoshi/blob/master/packages/yoshi-config/src/jest/index.ts#L28.

Please notice, we have code duplication here. The allowed keys are in [`validConfig.ts`](https://github.com/wix/yoshi/blob/master/packages/yoshi-config/src/jest/validConfig.ts#L17), and also in `validate`'s [`recursiveBlacklist`](https://github.com/wix/yoshi/blob/master/packages/yoshi-config/src/jest/index.ts#L28) property and in [`jest-preset`](https://github.com/wix/yoshi/blob/master/packages/jest-yoshi-preset/jest-preset.js#L30). For now, you should, at least, update those places as well.

docs/guides/jest-setup.md
```diff
@@ -103,7 +103,7 @@ Has a default test timeout of 10 seconds.
 - `server` - server to be bootstraped
   - `command` - runs a command which bootstrap the server
   - `port` - wait for a server to start listening on this port before running the tests. This port will be available in your server script as an environment variable (PORT)
-- `e2eOptions` - overrides for `e2e` environment. For now, only `globals` and `testURL` are available
+- `e2eOptions` - overrides for `e2e` environment. For now, only `moduleNameMapper` is available
 - `specOptions` - overrides for `spec` environment. For now, only `globals` and `testURL` are available
```
yanivefraim: Do we want `moduleNameMapper` to be for specs as well?

PR #1798, opened 2019-12-31
Fix PR website publishing

NitayRabi:
### 🔦 Summary
Change baseUrl if in PR. 

### 🗺️ Test Plan
...


website/siteConfig.js
```diff
@@ -1,12 +1,17 @@
 // See https://docusaurus.io/docs/site-config for all the possible
 // site configuration options.
 
+// TODO - this is a temp hack until teamcity-autorelease-surge would support Base Url.
+const isPr =
+  process.env.VCS_BRANCH_NAME &&
+  !!process.env.VCS_BRANCH_NAME.replace(/\D+/g, '');
```
ranyitz: ```!!process.env.VCS_BRANCH_NAME.replace(/\D+/g, '');```
This is probably not critical, but we'll remove it when the fix arrives

PR #1799, opened 2019-12-31
Create initial commit for `create-yoshi-app`

### LINKED ISSUE #1790: Create an initial commit after creating a new project [create-yoshi-app]
# 🙋 Feature Request

### 😯 Current Behavior

When creating a new app using CYA, a git repository is created but an initial commit needs to be performed by hand.

### 🤔 Expected Behavior

After creating an app, yoshi will create an initial commit.

### 🔦 Context

Every time that I generate an app I perform an initial commit, if i forget to do that, than my initial commit is not pristine, which means that I may accidentally made changes that caused my app to stop working, and now I'm not sure what are they because I don't have a diff between the template and my changes.

Also, looking at other toolkits, it looks like this is a standard. See https://github.com/facebook/create-react-app/pull/1288 for example. and also `create-next-app` and `angular-cli` are doing the same.

NitayRabi:
### 🔦 Summary
Closes #1790 

### 🗺️ Test Plan
Added test to validate first commit.

### DIFF
--- a/packages/create-yoshi-app/src/utils.ts
+++ b/packages/create-yoshi-app/src/utils.ts
@@ -49,6 +49,23 @@ export const lintFix = (dir: string) => {
   });
 };
 
+export const gitCommit = (dir: string) => {
+  console.log(`\nRunning ${chalk.magenta('initial commit')}`);
+  execa.sync('git add -A', {
+    shell: true,
+    cwd: dir,
+    stdio: 'ignore',
+  });
+  execa.sync(
+    'git commit -m "Initial commit from Create Yoshi App" --no-verify',
+    {
+      shell: true,
+      cwd: dir,
+      stdio: 'ignore',
+    },
+  );
+};
+
 export const gitInit = (dir: string) => {
   console.log(`\nRunning ${chalk.magenta('git init')}\n`);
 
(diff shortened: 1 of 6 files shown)


packages/create-yoshi-app/__tests__/createApp.test.ts
```diff
@@ -55,6 +55,18 @@ test('it should not create a git repo if the target directory is contained in a
...
 
+test('should create a git repo and commit an initial commit by CYA', async () => {
+  (runPrompt as jest.Mock).mockReturnValue(minimalTemplateModel());
+  (verifyRegistry as jest.Mock).mockReturnValue(undefined);
+
+  const tempDir = tempy.directory();
+  const projectDir = path.join(tempDir, 'project');
+  fs.ensureDirSync(projectDir);
+  await createApp({ workingDir: projectDir, install: false, lint: false });
+  const gitLog = execa.sync('git', ['log']).stdout;
```
ranyitz: Looking at the last commit message would be a bit cleaner, and you can match the whole message.

```suggestion
  const gitLog = execa.sync('git', ['show', '-s', '--format=%s']).stdout;
```

ranyitz: Also, I guess you'd like to run the git command on the [projectDir](https://github.com/sindresorhus/execa#cwd)

NitayRabi:
Replaced by #1822

PR #1805, opened 2020-01-01
jest-yoshi-preset - jest debug watch mode plugin.

### LINKED ISSUE #1178: Debugging our app's code in the browser takes too many steps.
# 🙋 Feature Request

### 😯 Current Behavior

In order to debug our app's in the browser when working with `jest-yoshi-preset` there are some steps that you need to do:

Looking at [puppeteer debugging tips](https://github.com/GoogleChrome/puppeteer#debugging-tips) and [jest-yoshi.config documentation](https://wix.github.io/yoshi/docs/guides/jest-setup#configuration)

1. Change `jest-yoshi.config.js`'s `puppeteer.devtools` option from `false` to  `true`.
2. Add a debugger statement in your code or using evaluate :`await page.evaluate(() => {debugger;});`
3. Extend jest's timeout: `jest.setTimeout(100000);` so it won't close the jest process.
4. Restart the jest process

### 🤔 Expected Behavior

1. No need to change a configuration for a specific run, the user experience should be better, for example, using jest's watch mode.
2. `jest-yoshi-preset` can assume that when `devtools: true` is used we want to extend jest's timeout, so there is no need to do that manually.
3. No need to restart the jest process (Should be solved on #1177)

### 💁 Possible Solution 

1. Write a new [jest watch plugin](https://jestjs.io/docs/en/watch-plugins#watch-menu-integration) that adds the `'d' (debug/devtools)` option to the jest watch menu. When you press on it, it will change the `devtools` option and will run the tests again, so essentially there is no need for a configuration change.

2. We can extend [jest's timeout](https://jestjs.io/docs/en/jest-object#jestsettimeouttimeout) when the `devtools` option is truthy.

### 🔦 Context

Debugging is important, and when so many steps are required, it makes debugging a hard task. Debugging should be easy, and we can use those simple heuristics in order to improve the workflow for everyone.

NitayRabi:
### 🔦 Summary
closes #1178 

- [x] Create a jest watch plugin to keep internal state for isDebuggingMode. 
- [x] Use plugin to dictate `devtools` behaviour in e2e.
- [x] Use devtools/debugMode to set jest timeout in e2e (passed it as global similar to `browserWSEndpoint`)
- [x] Add `debugBrowser` function to add breakpoint to puppeteer. 
- [x] Docs


packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -39,10 +39,14 @@ const supportedGlobalOverrideKeys = [
...
 const globalValidOverrides = pick(jestYoshiConfig, supportedGlobalOverrideKeys);
 
+global.__isDebugMode__ =
+  jestYoshiConfig.puppeteer && jestYoshiConfig.puppeteer.devtools;
+
 const config = {
   watchPlugins: [
     require.resolve('jest-watch-typeahead/filename'),
     require.resolve('jest-watch-typeahead/testname'),
+    __dirname + '/plugins/jest-watch-debug.js',
```
ranyitz: Can you use `path.join` instead?

PR #1810, opened 2020-01-02
Move suricate private dependency to be loaded by the consumer

ranyitz:
# 🔦 Summary

`@wix/suricate-client` is a private dependency, which added in #1764.

Currently there are projects that uses Yoshi as their build tool and use public CI machines like Travis/Circle to build their app. Those CIs can't access the private registry of Wix, and therefore the installation fails when there is a private dependency (such as `suricate-client`).

The temporary solution is to let the consumer to install the `@wix/suricate-client` dependency if it needs it. It's not the first time that we're using this workaround. The problem with it is that users needs to install various dependencies in specific versions which adds mental load and can lead to a waste of time.

___________

This situation isn't easy to handle and we would like to find a way to support only private users or have a good solution for OSS.


ronami:
Failing lint/tests and let's merge ❤️

ranyitz:
@ronami Thanks! I solved the lint problem, but the app flow tests seems to be unrelated - https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=5992800&buildTypeId=Yoshi_TestAppFlow&tab=buildLog&branch_Yoshi=1764%2Fhead&_focus=2827#_state=299

Something about the manifest

PR #1811, opened 2020-01-02
Generate `stats.json` during build and start into `dist/statics`

ronami:
### 🔦 Summary

Currently, [Webpack's stats](https://webpack.js.org/configuration/stats) are created only during `yoshi build` and written into `target/webpack-stats.json`.

Going forward, we want to use Webpack's stats to optimize SSR'd responses and this requires writing it during both (`yoshi build`), CI build and local development (`yoshi start`). On top of it, emitting it into `dist/statics` instead of `target` makes sense as it would make it available to see and debug projects that have been deployed. It also makes sense as we write all of Webpack's output to `dist/statics` and having this one exclusion is only confusing.

To not break any existing projects that count on the stats being emitted to the current folder, and to see that always writing stats, without an explicit `--stats` opt-in, this PR writes stats to the suggested directory on top of the existing one specifically for Thunderbolt. If this works well we can break the existing location for everyone and consider deprecating the `--stats` CLI flag.

### DIFF
--- a/packages/yoshi-flow-monorepo/src/webpack.config.ts
+++ b/packages/yoshi-flow-monorepo/src/webpack.config.ts
@@ -5,6 +5,8 @@ import {
   validateServerEntry,
   createServerEntries,
 } from 'yoshi-common/webpack-utils';
+// @ts-ignore
+import { StatsWriterPlugin } from 'webpack-stats-plugin';
 import { createBaseWebpackConfig } from 'yoshi-common/webpack.config';
 import { defaultEntry } from 'yoshi-helpers/constants';
 import { Config } from 'yoshi-config/build/config';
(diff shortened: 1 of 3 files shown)


ronami:
Adding for future reference: #1649, #1648.

yanivefraim:
One small suggestion: we are now using two different methods for writing the stats file:
1. For `--stats` flag: we write the stats file 'manually', using `bfj`, to `target` folder.
2. For `thunderbolt`: we use the webpack plugin, to `dist/statics`

I think that we can avoid this duplication by always using the plugin, and check for the write condition regarding where to write it to (and whether to write it)

WDYT?

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -113,6 +115,22 @@ export function createClientWebpackConfig(
 
   const useSplitChunks = pkg.config.splitChunks;
 
+  // Write stats file to `dist/statics` during start and build instead of only during build and into
+  // `target`
+  //
+  // In a following major version we will only generate stats into `dist/statics`
+  if (customThunderboltApp) {
```
yanivefraim: One small suggestion: we are now using two different methods for writing the stats file:
1. For `--stats` flag: we write the stats file 'manually', using `bfj`, to `target` folder.
2. For `thunderbolt`: we use the webpack plugin, to `dist/statics`

I think that we can avoid this duplication by always using the plugin, and check for the write condition regarding where to write it to (and whether to write it)

WDYT?

PR #1817, opened 2020-01-05
Add Kitchen Sink tests for moduleNameMapper configuration

### LINKED ISSUE #1814: Technical debt - add a test for `e2eOptions` jest override
Following this [PR](https://github.com/wix/yoshi/pull/1788), we actually removed the test for `e2eOptions` (we had one e2e test and now we practically removed it).

Solution: 
Add a new test for `e2eOptions`, using the available options that we have today

NitayRabi:
### 🔦 Summary
Closes #1814 

- [x] Remove bad config from the project (`globals` in `e2eOptions`)
- [x] Add example moduleNameMapper

### Points to consider
- [ ] `moduleNameMapper` in `specOptions` overrides the default value provided by Yoshi (for styling) - using it without specifying something specific for css will break specs. maybe we should merge the configurations. 

### 🗺️ Test Plan
Added test for configuration

### DIFF
--- a/test/projects/kitchensink-javascript/src/components/features/module-name.mapper.spec.js
+++ b/test/projects/kitchensink-javascript/src/components/features/module-name.mapper.spec.js
@@ -1,0 +1,8 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import ModuleNameMapper from './module-name-mapper';
+
+it('should pass', () => {
+  const div = document.createElement('div');
+  ReactDOM.render(<ModuleNameMapper />, div);
+});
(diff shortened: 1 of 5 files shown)


test/projects/kitchensink-javascript/src/components/features/module-name.mapper.spec.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import ModuleNameMapper from './module-name-mapper';
+
+it('should pass', () => {
+  const div = document.createElement('div');
+  ReactDOM.render(<ModuleNameMapper />, div);
```
yanivefraim: I would add a test for the div's content to have 'Stub module content', so the test would be more bulletproof. Now it only checks that you can import a `.foo` file without an error. WDYT?

yanivefraim:
Closing in favor of #1873

PR #1821, opened 2020-01-06
mind entry points when structuring the manifest

roymiloh:
This PR changes the structure of the emitted manifest to the following one:

```ts
interface Manifest {
  entry: string[];
}
```

The idea is to exclusively select files of specific entry, if needed (for example - in order to render only one specific bundle files in html). Currently all chunks are on the top level (flatten), so it's impossible to know which chunk belongs to each entry point, in case there are multiple.


test/typescript/features/manifest/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "feature",
+  "private": true,
+  "yoshi": {
+    "projectType": "app",
+    "splitChunks": {
```
ronami: I think we can configure this with `chunks: "all"` so it has less clutter (https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkschunks)

test/typescript/features/manifest/package.json
```diff
@@ -0,0 +1,18 @@
...
+      "maxSize": 100,
+      "minSize": 100
+    },
+    "entry": {
+      "app1": "./client.tsx",
```
ronami: Having the same name both for entries will be less confusing

test/typescript/features/manifest/pom.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ronami: Is this needed? Can this be moved to common template?

test/typescript/features/manifest/src/some-module.ts
```diff
@@ -0,0 +1,5 @@
+import styles from './styles.css';
+
+export const a = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
```
ronami: Can be `const a = "a"` if we use the suggestion above about `chunks: "all"`

test/typescript/features/manifest/manifest.test.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path');
```
yanivefraim: we have all tests in JS and TS. We should have this one in JS as well...

test/typescript/features/manifest/pom.xml
roymiloh: In order to generate url in prod manifest which contains parastorage thingy, it requires artifact id.

PR #1822, opened 2020-01-06
`create-yoshi-app` - perform initial commit

### LINKED ISSUE #1790: Create an initial commit after creating a new project [create-yoshi-app]

NitayRabi:
### 🔦 Summary
Closes #1790
Commit "Initial commit from Create Yoshi App" similar to CRA. 

### 🗺️ Test Plan
Test under createApp.


ranyitz:
Let's merge when the templates tests are green 🍏

PR #1825, opened 2020-01-07
Install with yarn in e2e tests

ranyitz:
### 🔦 Summary
See if installing with yarn is faster

> Edit

Trying to understand if this change improves the build time or not i've seen that it's hard to know from the CI log how much time each step took (for example the install stage).

So i've added some logs that shows when we start a new stage (`publish`, `install`)

### 🗺️ Test Plan
Compare with last builds and see if yarn gives better results


ranyitz:
It's hard to be 100% sure due to the changing times in our CI, but it looks like the time has reduced in around 30s-60s which is good enough for this small change.

PR #1830, opened 2020-01-07
Test infra - move more tests to the new infra

yanivefraim:


ranyitz:
If it's a work in progress PR you can open a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) instead

test/typescript/fixtures/base-template/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "skipLibCheck": true
   },
   "files": [
-    "../../../../packages/yoshi/types.d.ts"
+    "{%typesPath%}"
```
ronami: Possibly `yoshi-root-path`?

test/typescript/features/externals/externals.test.js
```diff
@@ -0,0 +1,20 @@
...
+      await page.goto('http://localhost:3000');
+      const result = await page.$eval('#externals', elm => elm.textContent);
+
+      expect(result).toBe('Some external text.');
+    });
+  });
+
+  // it('component tests', async () => {
+  //   await scripts.test(mode);
+  // });
```
ronami: Normally users will need to install an external dependency for component tests (as a dev dependency), I think we should consider a way to mock `node_modules` for each test, possibly like the `__node_modules__` in kitchensink tests. Wdyt?

test/scripts.js
```diff
@@ -48,6 +49,19 @@ module.exports = class Scripts {
       filter: file => !file.includes('.test.js'),
     });
 
+    //Process tsconfig template for dynamic content
+    const tsConfigPath = path.join(featureDir, 'tsconfig.json');
+    if (fs.pathExistsSync(tsConfigPath)) {
+      const templateData = {
+        typesPath: path.join(__dirname, '../packages/yoshi/types.d.ts'),
+      };
+      const fileContents = fs.readFileSync(tsConfigPath, 'utf-8');
```
ronami: Might be YAGNI but what about enabling the `templateData` in all files?

test/utils.js
```diff
@@ -142,6 +142,27 @@ async function terminateAsyncSafe(pid) {
...
+const templateRegex = /{%\w+%}/g;
+
+const replaceTemplates = (content, map, { graceful } = {}) =>
+  content.replace(templateRegex, match => {
+    const key = match.slice(2, -2);
+
+    if (!map.hasOwnProperty(key)) {
+      if (graceful) {
+        return match;
+      }
```
ronami: What's the use case for having this turned off?

test/typescript/fixtures/base-template/tsconfig.json
yanivefraim: Awesome!

test/typescript/features/externals/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "skipLibCheck": true
+  },
+  "files": [
+    "{%typesPath%}",
+    "./external-types.d.ts"
+  ],
+  "include": [
+    "src"
+  ]
+}
```
ronami: Can this possibly extend from the base template tsconfig.json with most defaults? We can use `yoshi-root-dir` template var along with `extends: ""`.

test/typescript/features/externals/src/external-types.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'externals';
+
+// interface Window {
+//   __LOCALE__: string;
+//   __BASEURL__: string;
+//   __CI_APP_VERSION__: string;
+// }
```
ronami: Can this be removed?

test/typescript/features/externals/externals.test.js
yanivefraim: Great idea. I'll do it in my next PR!

test/scripts.js
yanivefraim: I think that we should wait until it is a real use case. WDYT?

PR #1837, opened 2020-01-08
Test infra: add a component test for externals

yanivefraim:
Current kitchensink tests did not test for component tests with externals.


yanivefraim:
Thinking about this, this test is actually useless. It does not test anything, and we can remove it.

PR #1839, opened 2020-01-08
Allow Opt-in Storybook support

NitayRabi:
#### Without opting-in

- [x] without opting in - no storybook dependencies should be set. consumers have to install `yoshi-storybook-dependencies` for the feature.

#### With opting-in
- [x] app-flow should support opt-in `storybook` option. with that option set to `true`, storybook server with watch mode will start along side the dev-server.

image

- [x] In case feature is set to true, consumers will have to install `yoshi-storybook-dependencies`
	- [x] Storybook process will not prevent or interfere with Yoshi's dev-server

image

- [x] Once feature is on and all dependencies installed - notify the user and open the browser. 

image

- [x] Consumer should be able to create stories similar to Storybook's [CSF's format](https://storybook.js.org/docs/formats/component-story-format/)

image

#### Not in current scope - 

- Addons
- Custom head/body
- Svelte support

### 🗺️ Test Plan
Should be tests E2E under features for both TS & JS projects


NitayRabi:
Closed for better - https://github.com/wix/yoshi/pull/1861

PR #1844, opened 2020-01-12
adding test command

deanshub:
### 🔦 Summary
adding test command of it's own to yoshi-flow-editor


deanshub:
because currently it uses the legacy test command which loads yoshi config (something that we don't want in `yoshi-flow-editor`) plus if we want to give the best development experience we need a predefined test and preset

ranyitz:
I agree that we would want it in the future. I think it's safe to say that the first users will use the current testing infrastructure (the one we have on both legacy and app flows). I don't want to diverge from the main flow yet.

deanshub:
this will cause issues...
cdn ssl default is false so they need to create a yoshi.config file
I know that the perfect solution is to refactor the way test command works but since this is a complete divergence I decided to copy and omit the irrelevant parts, so when we have time we can get back to it

ranyitz:
You are right, we don't want them to create a yoshi config, or even be aware to the concept of Yoshi config. Currently the amount of projects we work with is quite small and we can verify that they don't have such config.

I'm ok with having the `ssl` option configurable via environment parameter so the editor flow will be able to configure it.

https://github.com/wix/yoshi/blob/master/packages/yoshi-flow-legacy/src/commands/test.js#L89

We can check for `CDN_SSL=true` and set the `ssl` option to be truthy accordingly.

PR #1855, opened 2020-01-14
Remove yarn.lock file (Fix renovate bot)

ranyitz:
### 🔦 Summary
**TL;DR**
Remove the lock file so renovate will continue to work.

Since we've merged the `yarn.lock` file we've got much faster installs, but renovate stopped working.

This PR will remove `yarn.lock` and renovate will continue to work, hopefully in the future we'll be able to work with both.

The reason for renovate to stop working is that we have private dependencies and renovate doesn't have access to the private registry when trying to recreate the lock file (needed for the version bumps)


amitdahan:
Ahh, so very unfortunate.
Can we still work with a `yarn.lock` locally?
I guess what I mean is, can we only instruct renovate to not use a lock file, but keep using it ourselves / locally?

ranyitz:
@amitdahan The problem is that you don't want to update `package.json` while not updating `yarn.lock`. The problem is that those changes to the lock file are getting to unrelated PRs. Therefore we run `yarn --frozen-lockfile` in CI. 

I also haven't found a solution for it in [renovates docs](https://docs.renovatebot.com).

ranyitz:
When trying to make this PR pass I've suffered from many transitive dependencies of clashing `@types`. The lock file helped to maintain those...

Also, going back to `150sec` install (after around `30sec`) i'm willing to invest time in finding another way.

As the problem is private modules, renovate [has a solution for it](https://docs.renovatebot.com/private-modules/)

I'm talking about this:

> **Renovate App with a private registry**: Add an unencrypted npmrc plus an encrypted npmToken in config

Now i'm checking if that's possible, and hopefully we'll be able to continue working with a lock file 💪

ranyitz:
Our private registry requires being connected to VPN, which would require us to run renovate-bot on our own.

Lock file is too important for now.

We'll **upgrade the dependencies manually** and we'll return to the renovate flow when it will be possible.

PR #1858, opened 2020-01-15
Use async attribute instead of defer for injected script tags

ronami:
### 🔦 Summary

Changed accidentally in #1796. Performance-wise, and according to Thunderbolt's performance needs, we're going to use `async` currently.


kfiroo:
lets not merge this yet please 🙏

PR #1859, opened 2020-01-16
Support private modules for renovate (Add encrypted `npmToken`)

ranyitz:
### 🔦 Summary

As suggested in https://github.com/wix/yoshi/pull/1855#issuecomment-574710369

We want to support private modules in lock file.

We can do it with the encrypted `npmToken` explained [here](https://docs.renovatebot.com/private-modules/)


ranyitz:
Closing until we'll get security allowance

PR #1861, opened 2020-01-16
Storybook managed under experimental flag

NitayRabi:
### 🔦 Summary
- Not opting in should result in same log (no breaking changes in product)

image

- Activating feature using `experimentalStorybook: true` in Yoshi configuration.

image

- Show errors in priority (if both Storybook and Yoshi compilation has error - Yoshi wins) 

### 🗺️ Test Plan
Tested TS/JS projects - 
- Happy flow. 
- Show informative errors on storybook compilation issues. 
- Show informative errors on yoshi compilation errors (Yoshi priority) 

#### Regarding [changes in test script](https://github.com/wix/yoshi/pull/1861/files#diff-4e2a83a598cdaf6ab71d5de299dcd1f2R112)
(@yanivefraim @ranyitz @ronami WDYT?)

- Throttling the results from `waitForStdout` allows using and checking output for "non-happy" flow testing. 
- Catching errors in start.dev should not output log unless verbose passed (again supports "non-happy" flow testing)


test/scripts.js
```diff
@@ -122,9 +129,13 @@ module.exports = class Scripts {
 
       await callback();
     } catch (e) {
-      console.log('--------------- Yoshi Start Output ---------------');
-      console.log(startProcessOutput);
-      console.log('--------------- End of Yoshi Start Output ---------------');
+      if (this.verbose) {
```
yanivefraim: We want to see the compilation output in case that the test failed, always (not just for verbose mode). We can even say that we want to use here `if (!this.verbose) {...`, because in case of verbose the user already has those logs, and now he will get them twice!

test/utils.js
```diff
@@ -117,16 +117,26 @@ function waitForPortToFree(port) {
...
+function waitForStdout(spawnedProcess, stringToMatch, options) {
   let data = '';
 
   return new Promise(resolve => {
     spawnedProcess.stdout.on('data', function listener(buffer) {
       data += buffer.toString();
       if (buffer.toString().includes(stringToMatch)) {
         spawnedProcess.stdout.off('data', listener);
-        resolve(data);
+        if (options && options.throttle) {
```
yanivefraim: nit: nicer to use default options here, IMO.

jest.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   verbose: true,
   testResultsProcessor: 'jest-teamcity-reporter',
   testMatch: ['**/*.test.js'],
-  testPathIgnorePatterns: ['/node_modules/', '/projects/'],
```
ranyitz: Is this related?

packages/yoshi-common/.storybook/config.js
```diff
@@ -0,0 +1,5 @@
+import { configure }  from 'yoshi-storybook-dependencies/node_modules/@storybook/react';
```
ranyitz: We can't be sure that `@storybook/react` would be installed there (for example when using hoisting in a monorepo).

A good way to do it is to re-export the module from the root of `yoshi-storybook-dependencies`

See an [example from create-react-app](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/globby.js)

packages/yoshi-common/package.json
```diff
@@ -27,7 +27,9 @@
     "@types/webpack": "4.41.0",
     "@types/webpack-bundle-analyzer": "2.13.3",
     "@types/webpack-manifest-plugin": "2.1.0",
-    "@types/xmldoc": "1.1.4"
+    "@types/webpack-node-externals": "1.7.0",
+    "@types/xmldoc": "1.1.4",
+    "execa": "^4.0.0"
```
ranyitz: I think that, this should be a dependency (instead of `devDependency`)

packages/yoshi-common/.storybook/config.js
```diff
@@ -0,0 +1,5 @@
...
+
+const loaders = require.context(PROJECT_ROOT, true, /\.stories\.(ts|tsx|js)$/);
+configure(loaders, module);
```
ranyitz: Is this conventional way to configure storybook? If not, it makes sense to add a comment here or send to the docs/exmaple

packages/yoshi-helpers/src/utils.ts
```diff
@@ -20,6 +20,55 @@ export function logIfAny(log: any) {
   }
 }
 
+export const unprocessedModules = (p: string) => {
```
ranyitz: This is unrelated to this PR (probably removed in a different PR)

packages/yoshi-common/.storybook/config.js
NitayRabi: So using PROJECT_ROOT is Yoshi related and it's uncommon to use in `require.context()` which expects a string relative path. But the general usage is indeed storybook convention (https://storybook.js.org/docs/basics/writing-stories/#loading-stories)

*Note: I now see they added main.js/preview.js which seems new

packages/yoshi-helpers/src/utils.ts
NitayRabi: Maybe got in here via rebase?

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -25,24 +26,34 @@ type WebpackStatus = {
...
-      devServerUrls: Urls;
-    } & WebpackStatus)
-  | ({ status: 'errors' } & WebpackStatus)
-  | ({ status: 'warnings' } & WebpackStatus);
+      urls?: Urls;
+    } & Partial<WebpackStatus>)
+  | ({ status: 'errors' } & Partial<WebpackStatus>)
+  | ({ status: 'warnings' } & Partial<WebpackStatus>);
+
+export const enum ProcessType {
```
ronami: Please don't create enums, use a type alias instead. It's simpler and doesn't require an additional import.  For example https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d0c3a91328a5fa57940ec3e0b64152011a81187a/types/webpack/index.d.ts#L69.

packages/yoshi-common/src/storybook/storybook-worker.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+  storybookProcess.on('listening', (port: number) => {
+    process.send && process.send({ type: 'listening', port });
+  });
+  try {
+    await storybookProcess.start();
+  } catch (e) {
+    process.send && process.send({ type: 'error', error: e.message });
+  }
+})();
```
ranyitz: I think that we don't need `IIFE`/`async/await` here. We can just use good old promises (as we don't await for anything on `storybookProcess.start()`)

packages/yoshi-common/src/storybook/storybook-process.ts
```diff
@@ -0,0 +1,69 @@
+import path from 'path';
+import { EventEmitter } from 'events';
+import waitPort from 'wait-port';
+import createWebpackConfig from './storybook.webpack.config';
+
+const storyBookConfigFolder = path.join(__dirname, '../..', '.storybook');
```
ranyitz: We're using `storybook` everywhere, let's don't use `storyBook` as well.

packages/yoshi-common/src/storybook/storybook-process.ts
```diff
@@ -0,0 +1,69 @@
...
+    super();
+    this.port = port;
+  }
+
+  static create({ port = 9009 }: StorybookProcessOptions) {
+    return new StorybookProcess({ port });
+  }
+
+  start = async () => {
+    let storyBookYoshiDepsFolder;
```
ranyitz: storyBook => storybook

packages/yoshi-common/src/storybook/storybook-process.ts
```diff
@@ -0,0 +1,69 @@
...
+    } catch (e) {
+      throw new Error(
+        'Please install yoshi-storybook-dependencies in order to run storybook',
+      );
+    }
+
+    this.emit('compiling');
+
+    // eslint-disable-next-line import/no-extraneous-dependencies
+    const storybook = require(`${storyBookYoshiDepsFolder}/node_modules/@storybook/react/standalone`);
```
ranyitz: We can't be sure that `@storybook/react` will be installed exactly there. (for example, when using `hoisting`) let's expose `@storybook/react` as a whole on `yoshi-storybook-dependencies` similarly to [this](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/globby.js)

packages/yoshi-common/src/storybook/storybook-process.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const webpackConfig = createWebpackConfig({
+      projectRoot: path.join(process.cwd(), 'src'),
```
ranyitz: This can be problematic in monorepos (e.g. `editor elements`). Project root should be passed from outside

packages/yoshi-common/src/storybook/storybook-process.ts
```diff
@@ -0,0 +1,69 @@
...
+      mode: 'dev',
+      webpackConfig,
+      port: this.port,
+      ci: true,
+      quiet: false,
+      configDir: storyBookConfigFolder,
+    });
+
+    await waitPort({
+      port: +this.port,
```
ranyitz: Why the `+`?

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -25,24 +26,30 @@ type WebpackStatus = {
...
-  | ({ status: 'warnings' } & WebpackStatus);
+      urls?: Urls;
+    } & Partial<WebpackStatus>)
+  | ({ status: 'errors' } & Partial<WebpackStatus>)
+  | ({ status: 'warnings' } & Partial<WebpackStatus>);
+
+export type ProcessType = 'DevServer' | 'Server' | 'Storybook';
+
+export type State = {
+  [type in ProcessType]?: ProcessState;
```
ranyitz: why the `?` is needed?

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -85,29 +98,103 @@ export default class DevEnvironment {
...
         this.store.setState({
-          status: 'success',
-          serverUrls,
-          devServerUrls,
-          ...messages,
-        } as State);
+          Server: {
+            status: 'success',
+            urls: serverUrls,
+            ...messages,
```
ranyitz: those are webpack stats messages, meaning they are related to "webpack-dev-server" so it makes sense to me that why will be passed on the "Dev-server" project.

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -85,29 +98,103 @@ export default class DevEnvironment {
...
         if (messages.errors.length > 1) {
           messages.errors.length = 1;
         }
 
         this.store.setState({
-          status: 'errors',
-          ...messages,
+          Server: {
+            status: 'errors',
+            ...messages,
```
ranyitz: same here

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -85,29 +98,103 @@ export default class DevEnvironment {
...
+        this.store.setState({
+          Storybook: { status: 'errors', errors: [message.error] },
+        });
+        break;
+      case 'compiling':
+        if (isInteractive) {
+          clearConsole();
+        }
+        this.store.setState({
+          Storybook: { status: 'compiling', errors: [], warnings: [] },
```
ranyitz: why do we need to set `errors: []` and `warnings: []`?

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -314,13 +403,25 @@ export default class DevEnvironment {
       cwd,
     });
 
+    let storybookProcess;
+
+    if (storybook) {
+      const pathToStorybook = path.join(
```
ranyitz: `pathToStorybook` => `pathToStorybookWorker`

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -314,13 +403,25 @@ export default class DevEnvironment {
...
+        __dirname,
+        'storybook',
+        'storybook-worker',
+      );
+      storybookProcess = execa.node(pathToStorybook, [], {});
```
ranyitz:
```suggestion
      storybookProcess = execa.node(pathToStorybook);
```

packages/yoshi-common/src/dev-environment-logger.ts
```diff
@@ -1,44 +1,122 @@
...
+      break;
+
+    case 'success':
+      logUrls({ processType, suricate, appName, urls: state.urls });
+      break;
+  }
+};
+
+const hasErrorsOrWarnings = (state: State): boolean => {
+  return Object.keys(state).some(stateName => {
```
ranyitz: You can use `Object.values` which also removes the line below.
```suggestion
return Object.values(state).some(processState => {
```

packages/yoshi-common/src/dev-environment-logger.ts
```diff
@@ -1,44 +1,122 @@
...
+    const processState = state[stateName as ProcessType];
+    return ['errors', 'warnings'].includes(processState?.status as string);
```
ranyitz: It should also work like that as `processState` always has a status
```suggestion
    return ['errors', 'warnings'].includes(processState.status);
```

packages/yoshi-common/src/dev-environment-logger.ts
```diff
@@ -1,44 +1,122 @@
...
+  });
+};
+
+const logStateErrorsOrWarnings = (state: State) => {
+  const { Server, Storybook } = state;
```
ranyitz: I think this should be the "DevServer" state ("failed to compile")

packages/yoshi-helpers/src/utils.ts
```diff
@@ -20,6 +20,25 @@ export function logIfAny(log: any) {
   }
 }
 
+export const createBabelConfig = (presetOptions = {}) => {
```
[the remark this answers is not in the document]
NitayRabi: It's not the first time this happens... I think the rebase get's this in for some reason

packages/yoshi-storybook-dependencies/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "description": "",
+  "main": "build/index.js",
+  "scripts": {
+    "start": "start-storybook"
+  },
+  "author": "Nitay Rabinovich",
+  "license": "ISC",
+  "devDependencies": {},
+  "dependencies": {
+    "@babel/core": "^7.7.7",
```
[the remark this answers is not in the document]
NitayRabi: Got in with Storybook init command - I see we bring them with Yoshi, so we'll remove them.

packages/yoshi-common/src/dev-environment-logger.ts
NitayRabi: It's due to TS inferring the array as array of string, so it doesn't know how to compare the `status` to string.

packages/yoshi-common/src/dev-environment-logger.ts
NitayRabi: Currently DevServer & Server are sharing they same compiler (and then the same logs) - I stored the messages from the compiler in the Server. (I Can move it to DevServer if you think it's more suitable)

PR #1864, opened 2020-01-19
Always use Svelte as an external dependency in server bundle

ronami:
### 🔦 Summary

This PR makes two changes:

- Removes the un-needed custom workaround for not having `thunderbolt-elements` as an external dependency in the server bundle.
- Regardless of `useNodeExternals `, always use `svelte` as an external dependency in the server bundle. This is needed so we don't have two different copies of Svelte (one for `thunderbolt` and one for `thunderbolt-elements`). Otherwise, context can't be shared between the two different builds.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -924,6 +922,16 @@ export function createBaseWebpackConfig({
                 return callback();
               }
 
+              // Svelte should always be external on the server side
+              if (res.match(/svelte\/node_modules/)) {
```
ranyitz: This is a specific code for `thunderbolt`/`thunderbolt-elements` case.

If we plan on leaving it like this, I would mention it with a comment (e.g. `this is a specific code for thunderbolt...`), the other option would be to expose a `server.externals` configuration. (Similarly to `externals` or `webWorker.externals`) but maybe we should wait for another use-case for that.

ronami: 👍

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -304,8 +304,6 @@ export function createBaseWebpackConfig({
     reAssets,
     /bootstrap-hot-loader/,
     /yoshi-server/,
-    // Temporary work-around for thunderbolt
```
ranyitz: Only in our codebase, Temporary actually means Temporary 😃

PR #1868, opened 2020-01-21
Add a rule to disallow enums

ranyitz:
### 🔦 Summary

Currently there are two ways of achieving the same thing: 

[string literal type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types)

```ts
type Direction = "up" | "down" | "left" | "right"
````

[enum](https://www.typescriptlang.org/docs/handbook/enums.html)

```ts
enum Direction {
  Up,
  Down,
  Left,
  Right,
}
```

This change prevent the usage of enums.

### Why

* [Some of the problems that enums brings](https://github.com/typescript-eslint/typescript-eslint/issues/561)
* We don't want two ways of achieving the same thing
* The same developer experience capabilities (type check & autocomplete) are given to you with both, see also - https://github.com/palantir/blueprint/issues/324#issuecomment-270002485 & https://stackoverflow.com/a/49762122/5678535

### How

Prior art - https://github.com/pmedianetwork/eslint-config/pull/43/files


ranyitz:
@yavorsky I'm not sure. This would make the migration much harder (as code changes will be required) and it is a very opinionated change. It is also not autofixable. So as much as I believe it's the right thing to do, we can discuss it, but it does not follows our linting guidelines:

> Autofixable - We try to use mostly autofixable rules, so lint won't be much of a hassle
> Strict & balanced - Add non-fixable rules only if they will help to find bugs

https://wix.github.io/yoshi/blog/2018/06/3/yoshi-3#-lint

1. We can add a `yoshi/strict` mode which will use it!
2. We still expose `tslint` preset, we should curate an `eslint` one.
3. We should update the guidelines so they would be discoverable: Maybe the contributing is a good place? maybe the documentation, maybe both.

yavorsky:
Or as an option we can follow some major updates migration flow.
1. Don't break it on the next major update, but add the rule with `warn` level. Collect some more feedback from people who has warnings.
2. Update level to `error` for the major update after.

I like this solution b/c it's a long-term way to implement some breaking changes and on the same time provide good practices and collect more feedback.. but yeah, we need to discuss it and understand how strictly we want to be :)

ranyitz:
> we need to discuss it and understand how strictly we want to be

That's the first thing we need to understand. Currently, as much as I want everyone to use it, I'm also understand that it's opinionated and doesn't necessarily prevents bugs. Therefore it doesn't pass our guidelines.

Regrading the how? I think that your idea is great.

> Or as an option we can follow some major updates migration flow.

1. Deprecation period
2. Removal on a major version

PR #1874, opened 2020-01-22
Lock file maintanance 👷‍♂️

ranyitz:
### 🔦 Summary
Perform `yarn upgrade` to update the lock file

### 🗺️ Test Plan
All tests are passing


amitdahan:
Why'd we have to pin `@types/node` after this?

ranyitz:
@amitdahan When updating the lock file, we get `@types/node@13` which clashes with  other types that we have.

We don't even use `node@13` so I don't see a reason we'll use those types right now

amitdahan:
Why did we suddenly get `@types/node@^13`? We specify `12` in our packages, no?

ranyitz:
@amitdahan This is the error.

image

We got `@types/node@13` because many packages in defintly typed which has a reference to `@types/node` get it as `*` in their `package.json`'s dependencies.

[Further reading](https://wix.slack.com/archives/CAL591CDV/p1577203493047000?thread_ts=1577177730.033800&cid=CAL591CDV)

ranyitz:
Pinned `@types/webpack`.

There were multiple versioned installed

image

We need to remember to remove this resolution when we update the `@types/webpack` version

ranyitz:
I needed to pin `@types/react` version as well

image

We'll need to remove/change the resolution in the future when we'll update the version

ranyitz:
image

I needed to update @wix/wix-experiments, it supplied types and there were mismatches between the different versions that got to `node_modules`

PR #1883, opened 2020-01-23
Fix svelte-server-side-preprocess to report svelte filename

nirnatan:
### 🔦 Summary
Fix svelte-server-side-preprocess to report svelte filename instead of module.id in order to have the name of the module both in production build and development build

### 🗺️ Test Plan
...


packages/yoshi-common/src/svelte-server-side-preprocess.ts
```diff
@@ -4,30 +4,35 @@ export default function({ packageName }: { packageName: string }) {
...
+  const injectedCode = (filename: string) => /* js */ `
 
   const modulesContext = require('svelte').getContext('__modules__')
 
   if (modulesContext) {
-    modulesContext(module, '${packageName}')
+    modulesContext('${filename}', '${packageName}')
   }
 
 `;
```
ronami: Instead of passing a variable I think we can move this into the scope of the `markup` function bellow. Then we can just use `filename` instead of passing it to a function.

packages/yoshi-common/src/svelte-server-side-preprocess.ts
```diff
@@ -4,30 +4,35 @@ export default function({ packageName }: { packageName: string }) {
...
   const scriptRegex = /(?<=<script>)((.|\n|\r)*)(?=<\/script>)/;
 
-  const injectedCode = `
+  const injectedCode = (filename: string) => /* js */ `
 
   const modulesContext = require('svelte').getContext('__modules__')
 
   if (modulesContext) {
-    modulesContext(module, '${packageName}')
+    modulesContext('${filename}', '${packageName}')
```
ronami: I think we should pass `filename` as a 3rd argument and still pass the `module` variable. It has a lot of information and I think that we might want to access it in the future.

packages/yoshi-common/src/svelte-server-side-preprocess.ts
```diff
@@ -4,30 +4,35 @@ export default function({ packageName }: { packageName: string }) {
...
   }
 
 `;
 
   return {
-    markup: ({ content }: { content: string }) => {
+    markup: ({ content, filename }: { content: string; filename: string }) => {
       const scriptContent = scriptRegex.exec(content);
+      const pathParts = filename.split('/');
+      const svelteFileName = pathParts.pop()!;
```
ronami: Can we use [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) instead?

PR #1888, opened 2020-01-26
Yoshi server refactor + tests 

yanivefraim:
This PR adds back (#1476) support for `yoshi-server`. It also added tests for it.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -272,9 +273,9 @@ export function createBaseWebpackConfig({
   exportAsLibraryName?: string;
   nodeExternalsWhitelist?: Array<RegExp>;
   useAssetRelocator?: boolean;
+  useYoshiServer?: boolean;
 }): webpack.Configuration {
   const join = (...dirs: Array<string>) => path.join(cwd, ...dirs);
-
```
ronami: Super tiny but I think an empty line here helps in terms of readability.

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -687,6 +687,20 @@ export function createBaseWebpackConfig({
           },
         },
 
+        ...(useYoshiServer
+          ? [
+              {
+                test: /\.api\.(js|ts)$/,
+                ...(target === 'node'
+                  ? {
+                      issuer: () => true,
```
ronami: I think we should have a comment that explains why this is needed. It's not immediately obvious.

test/javascript/features/yoshi-server/api-client-server-react/src/components/App/App.spec.js
```diff
@@ -0,0 +1,17 @@
...
+  it('renders a title correctly', () => {
+    const { getByTestId } = render(
+      <I18nextProvider i18n={i18n}>
+        <App />
+      </I18nextProvider>,
+    );
+
+    expect(getByTestId('app-title').textContent).toBe('app.title');
+  });
+});
```
ronami: Is this file needed? How does it relate to Yoshi Server? Can it be simplified (i18next...)?

test/javascript/features/yoshi-server/api-client-server-react/src/templates/app.ejs
```diff
@@ -1,13 +1,11 @@
...
-    <title><%= title %></title>
-  </head>
   <body>
-    <h1><%= title %></h1>
     <div id="root"></div>
 
     <script crossorigin src="https://unpkg.com/react@16/umd/react.development.js"></script>
     <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
+    <script src="http://localhost:3200/webpack-runtime.bundle.js"></script>
+    <script src="http://localhost:3200/app.chunk.js"></script>
```
ronami: Since we have `experimentalBuildHtml` turned on in all Yoshi Server tests (maybe we can enable it if Yoshi Server is enabled?),

test/javascript/features/yoshi-server/api-client-server-react/src/components/App/App.spec.js
yanivefraim: thanks

test/javascript/features/yoshi-server/api-client-server-react/src/templates/app.ejs
yanivefraim: I'll open an issue, let's discuss it

PR #1894, opened 2020-01-27
Transpile webworker importScripts into fetch and eval

giladsegal:
### 🔦 Summary
Dynamic import statements in web workers are transpiled by default into `importScripts` statements. Those statements are synchronous and yield bad performance. This PR transpiles dynamic imports into `fetch` and `eval` which can download chunks **asynchronously** by using a custom target function

### 🗺️ Test Plan
TBD


packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -197,7 +198,7 @@ export function createWebWorkerWebpackConfig(
   const workerConfig = createBaseWebpackConfig({
     cwd: pkg.location,
     configName: 'web-worker',
-    target: 'webworker',
+    target: asyncWebWorkerTarget as any,
```
giladsegal: TODO..

packages/yoshi-common/src/AsyncWebWorkerTarget/AsyncWebWorkerTarget.ts
```diff
@@ -0,0 +1,29 @@
...
+// @ts-ignore
+import NodeSourcePlugin from 'webpack/lib/node/NodeSourcePlugin';
+// @ts-ignore
+import FunctionModulePlugin from 'webpack/lib/FunctionModulePlugin';
+// @ts-ignore
+import LoaderTargetPlugin from 'webpack/lib/LoaderTargetPlugin';
+import { AsyncWebWorkerPlugin } from './AsyncWebWorkerPlugin';
+
+export function asyncWebWorkerTarget(compiler: any) {
+  // TODO: this is the original webworker target implementation
```
giladsegal: what do you think? @ronami

packages/yoshi-common/src/AsyncWebWorkerTarget/AsyncWebWorkerTarget.ts
```diff
@@ -0,0 +1,29 @@
...
+  // here I don't have access to the options so I couldn't config
+  // all plugins properly - WDYT?
```
giladsegal: what do you think?
@ronami

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -567,7 +567,7 @@ export function createBaseWebpackConfig({
           ]
         : []),
 
-      ...(target === 'webworker'
+      ...(target === 'webworker' || typeof target === 'function'
```
giladsegal: we need to solve this hack as well @ronami

NitayRabi:
@giladsegal + @ronami - We want the async webworker to be the default configuration for all webworkers targets in Yoshi?

giladsegal:
We are going to implement a new target in yoshi so you can pick between `webworker` and `webworker-async`

packages/yoshi-common/src/AsyncWebWorkerTarget/AsyncWebWorkerTarget.ts
ronami: Can this be a function in the form of `(options) => compiler => void`? Then we can use it with: `{ ..., target: asyncWebWorkerTarget(options) }`. Wdyt?

giladsegal: Just realized it by myself and then I saw the comment, lol

giladsegal: But in order to reference the options, I have first to initialize the object and then invoke the target function with a reference to the existing object

packages/yoshi-common/async-webworker-target.d.ts
```diff
@@ -0,0 +1 @@
+export * from './build/AsyncWebWorkerTarget/AsyncWebWorkerTarget';
```
ronami: With #1896 and #1895, we are moving away from this pattern for internal APIs (not user-facing). I think we can import this file directly from `/build/...`.

packages/yoshi-common/async-webworker-target.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./build/AsyncWebWorkerTarget/AsyncWebWorkerTarget');
```
ronami: Same as above.

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -197,7 +197,7 @@ export function createWebWorkerWebpackConfig(
   const workerConfig = createBaseWebpackConfig({
     cwd: pkg.location,
     configName: 'web-worker',
-    target: 'webworker',
+    target: 'async-webworker',
```
ronami: This is going to affect every project that uses the monorepo setup and a web worker (probably only thunderbolt). Because this is a pretty experimental feature, I think we should see that it works on one project before expanding it.

I think we should use this target specifically for `thunderbolt-elements` and expand it (possibly with an opt-in flag?) a bit later when we see that it works. Wdyt?

packages/yoshi-common/src/AsyncWebWorkerTarget/AsyncWebWorkerPlugin.ts
```diff
@@ -0,0 +1,20 @@
+// @ts-ignore
+import WebWorkerChunkTemplatePlugin from 'webpack/lib/webworker/WebWorkerChunkTemplatePlugin';
+// @ts-ignore
+import WebWorkerHotUpdateChunkTemplatePlugin from 'webpack/lib/webworker/WebWorkerHotUpdateChunkTemplatePlugin';
```
ronami: Since we have quite a lot of these, I think using `require()`s can be better. It returns `any` by default and doesn't require having `@ts-ignore` on top of each of them. I think these files can even by JavaScript files, and use the `allowJs` compiler option. This way we won't have to declare many params as `any`. Wdyt?

ronami:
> @giladsegal + @ronami - We want the async webworker to be the default configuration for all webworkers targets in Yoshi?

Good question, I don't know. I think we should test it specifically on one use-case (`thunderbolt-elements`) before we expand it (possibly) to other projects.

packages/yoshi-flow-monorepo/src/webpack.config.ts
giladsegal: NP - I will do it, but does the mono-repo flow has additional users? I thought only `thunderbolt` and `thunderbolt-elements` are using it

packages/yoshi-common/src/AsyncWebWorkerTarget/AsyncWebWorkerPlugin.ts
giladsegal: `allowJs` would change the entire project compilation
I think that we can simply use `// @ts-nocheck` which is like `// @ts-ignore` but for an entire file. WDYT?

PR #1895, opened 2020-01-27
Remove files from the root of yoshi-common

ranyitz:
and replace with import to the `build` folder

### 🔦 Summary
Using the re-exporting pattern from the root is really good for external API.

For our case, this is no external API, everything is checked using TypeScript and it's not critical how it looks.

In reality it made the searching for a file in the IDE (`cmd+p`) hard as we had many files that did nothing.

It also created a procedure when we needed to create additional `.js` & `.d.ts` in order to expose a function.

This PR verifies that `yoshi-common` is removing this pattern.


amitdahan:
I'd like to take a moment and appreciate the confidence we have in this refactor mostly thanks to TypeScript.

ranyitz:
@amitdahan yeah, and this change as well - https://github.com/wix/yoshi/pull/1895

amitdahan:
@ranyitz you mean some other PR? You linked this one 😄

ranyitz:
@amitdahan are you suggest there is a [recursion](https://github.com/wix/yoshi/pull/1895) here?

amitdahan:
Black magic!

PR #1907, opened 2020-01-29
add description to `shouldTranspileFile` function

ranyitz:
### 🔦 Summary

Add description to a not very clear function


packages/yoshi-common/src/utils/should-transpile-file.ts
```diff
@@ -1,6 +1,7 @@
 import path from 'path';
 import config from 'yoshi-config';
 
+// accepts a file path and return true if it needs to be transpiled
```
ronami: Nice! I think changing the argument `p` to `fileName` can also help with readability.

packages/yoshi-common/src/utils/should-transpile-file.ts
```diff
@@ -1,6 +1,7 @@
...
 export default (p: string): boolean => {
```
yanivefraim:
```suggestion
export default (fileName: string): boolean => {
```

yanivefraim:
```suggestion
export default (fileName: string): boolean => {
```

PR #1910, opened 2020-01-30
Refactor tests, merge yoshi configs

yanivefraim:
This is a possible solution to the `package.json` duplication across test projects with custom yoshi configs. 
The problem with this solution is that the file is running in a different location than the one it is written in. In this case, when requiring the baseConfig, we have to use:

```
const config = require('./baseConfig');
```

Which is confusing. Plus the use cannot click to go to definition.
@ranyitz - WDYT?


ranyitz:
@yanivefraim In this case I prefer longer import paths with clear API, jump to definition.

Something like:

```diff
- const config = require('./baseConfig');
+ const config = require('../../../fixtures/yoshi.config');
```

yanivefraim:
> @yanivefraim In this case I prefer longer import paths with clear API, jump to definition.
> 
> Something like:
> 
> ```diff
> - const config = require('./baseConfig');
> + const config = require('../../../fixtures/yoshi.config');
> ```

Jump to definition cannot work, because the location of the test and the place that it is running in are different. We will have to write the `require` according to the place that the test is running at (`.tmp/...`), which is different from test original location.

yanivefraim:
@ranyitz - so this solution is working exactly the way we wanted (I think):

https://github.com/wix/yoshi/pull/1910/commits/e94b5d4e466e340a521031619438c115ac37b6f6#diff-1622ba861798db21c7fb62f58d00687bR1

```
const config = require('../../../../test/fixtures/javascript/base-template/yoshi.config');
```

The only thing left is the duplication between JS and TS. I put this file in another place, and copy it separately, but, we have more duplications between JS and TS base templates, and we do not take special care for them as well...
WDYT?

PR #1911, opened 2020-01-30
Simplify moment exclusion test (Stabler CI)

NitayRabi:
### 🔦 Summary
- toMatchJS with complex (recursive) regex might cause call stack issues. 
- Use simple regex instead of "negative" regex search
- implement notToMatchJS to support simple regex


test/utils.js
```diff
@@ -72,6 +72,26 @@ const matchJS = async (chunkName, page, regexes) => {
   }
 };
 
+const notToMatchJS = async (chunkName, page, regexes) => {
```
yanivefraim: `matchJs` and `notToMatchJS` are duplicated. Can we try to reuse code between them?

PR #1915, opened 2020-02-02
Refactor test utils, remove code duplication

yanivefraim:
Remove code duplication: `notToMatchJS` and `toMatchJS` had common code. (this is an addition to #1911)


test/utils.js
```diff
@@ -53,44 +53,35 @@ const matchCSS = async (chunkName, page, regexes) => {
...
+  const content = await getChunkContentFromScriptTag(page, chunkName);
 
   for (const regex of regexes) {
-    expect(content).toMatch(regex);
+    expect(content).not.toMatch(regex);
   }
 };
 
-const notToMatchJS = async (chunkName, page, regexes) => {
+async function getChunkContentFromScriptTag(page, chunkName) {
```
ronami: Not part of the code review but I see now that we can probably remove the `page` variable from being passed and used the globally available one (same for other functions, we're just passing a global variable). I'll try to push this in a future change.

PR #1918, opened 2020-02-03
eslint-tslint-config-yoshi - remove no-full-wsr-lib

mykas:
### 🔦 Summary

no-full-wsr-lib rules has been created previously to change users imports automatically:

```diff
- import { TextButton } from 'wix-style-react';
+ import TextButton from 'wix-style-react/TextButton';
```

But now our library finally supports named-imports and tree-shaking so there is no need to enforce it.


mykas:
Sure, should be good now :)

PR #1921, opened 2020-02-04
Yoshi Server: add a test for route with params

yanivefraim:
An important test for routing with params.


packages/yoshi-server/src/server.ts
```diff
@@ -81,8 +81,8 @@ export default class Server {
     return serverChunks.map(absolutePath => {
       const chunk = importFresh(absolutePath) as RouteFunction<any>;
       const relativePath = `/${relativeFilePath(routesBuildDir, absolutePath)}`;
+      // Change `/users/[userid]` to `/users/:userid`
```
ronami: Nice 😅

PR #1925, opened 2020-02-04
Yoshi Server: add global middleware support

yanivefraim:
This is the first step in adding middleware support to Yoshi Server.
We want to be able to easily add a global middleware, which will be applied to all routes and functions.
This is a suggestion for the following syntax:

add `src/_middleware_.js` file with the following:

```
const helloMiddleware = (req, res, next) => {
  req.hello = 'hello from yoshi server middleware';
  next();
};
export default [helloMiddleware];

```

This middleware will be applied to all routes and server functions.

TODO:
1. Add a test for middleware which does `res.send(...)` inside
2. Add support for middleware specifically for a route/function
3. Add a test for middleware order (global and local middleware)
4. Add support for error middleware (?)

### DIFF
--- a/packages/yoshi-common/src/webpack-utils.ts
+++ b/packages/yoshi-common/src/webpack-utils.ts
@@ -1,5 +1,6 @@
 import path from 'path';
 import fs from 'fs-extra';
+import resolve from 'resolve';
 import chalk from 'chalk';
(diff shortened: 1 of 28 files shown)


yanivefraim:
closing for #1928 1928

PR #1927, opened 2020-02-05
Emit Webpack stats file for worker compilation

ronami:
### 🔦 Summary

This is needed specifically to support analyzing the webworker output with `webpack-bundle-analyzer` or similar tools. This PR adds this change specifically for monorepo-flow.

### DIFF
--- a/packages/yoshi-flow-monorepo/src/webpack.config.ts
+++ b/packages/yoshi-flow-monorepo/src/webpack.config.ts
@@ -207,6 +207,16 @@ export function createWebWorkerWebpackConfig(
     ...defaultOptions,
   });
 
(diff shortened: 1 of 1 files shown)


packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -207,6 +207,16 @@ export function createWebWorkerWebpackConfig(
     ...defaultOptions,
   });
 
+  workerConfig.plugins!.push(
```
yanivefraim: It would be nice to add a test for it, but since it is only for the monorepo flow, it is a little problematic. 
What about other flows?

PR #1929, opened 2020-02-06
upgrade 18next and change fullstack templates

sidoruk-sv:
# 🔦 Summary
Because of the upgrade to lates `i18next`, we need also upgrade to react > 16.9. 

Also, `jsx-a11y/anchor-has-content` rule will not work for default usage of `<Trans/>` component. So need to discuss how it would be better to handle such case (for now just disabling this rule for 1 line, but in real code, it could spam with a lot of errors and `eslint-disable-next-line` is not a good solution for production code)


sidoruk-sv:
@ranyitz @ronami
Guys, sorry for interruption, but who can help here.

Tried this changes in CI (new project generated with ZERO_TO_PRODUCTION)
And all tests are passed there, but in yoshi tests templates failes for some reason.

http://tc.dev.wixpress.com/viewLog.html?buildId=115070373&buildTypeId=HelloSerhii2_HelloSerhiis2_O&tab=testsInfo

packages/create-yoshi-app/templates/fullstack/typescript/{%packagejson%}
```diff
@@ -32,31 +32,29 @@
...
-    "@types/react-i18next": "^7.8.2",
     "@wix/fedops-logger": "^5.0.0",
     "@wix/wix-bootstrap-testkit": "latest",
     "@wix/wix-config-emitter": "latest",
     "husky": "^4.2.1",
     "jest-yoshi-preset": "^4.1.0",
     "lint-staged": "^10.0.7",
     "puppeteer": "^1.1.0",
-    "typescript": "~3.6.0",
+    "typescript": "^3.7.5",
```
ranyitz: typescript is breaking on minors, please leave the `~`.
```suggestion
    "typescript": "~3.7.5",
```

packages/create-yoshi-app/templates/fullstack/typescript/{%packagejson%}
```diff
@@ -32,31 +32,29 @@
...
   },
   "devDependencies": {
     "@testing-library/react": "^9.3.0",
-    "@types/express": "~4.0.0",
-    "@types/i18next": "^8.4.5",
-    "@types/jest": "^24.0.0",
-    "@types/node": "^8.0.0",
+    "@types/express": "^4.17.2",
+    "@types/jest": "^25.1.1",
+    "@types/node": "^13.7.0",
```
ranyitz: Let's leave `@types/node` on lts for now

```suggestion
    "@types/node": "^12.0.0",
```

packages/create-yoshi-app/templates/fullstack/typescript/{%packagejson%}
```diff
@@ -32,31 +32,29 @@
...
     "yoshi-template-intro": "^4.1.0"
   },
   "devDependencies": {
     "@testing-library/react": "^9.3.0",
-    "@types/express": "~4.0.0",
-    "@types/i18next": "^8.4.5",
-    "@types/jest": "^24.0.0",
-    "@types/node": "^8.0.0",
+    "@types/express": "^4.17.2",
+    "@types/jest": "^25.1.1",
```
ranyitz: We're still using jest 24. In any case, the types are provided by the preset, so it can be removed.

https://github.com/wix/yoshi/blob/9a0df21288ece86556bcded22cfc8535b200c220/packages/jest-yoshi-preset/package.json#L23

packages/create-yoshi-app/templates/fullstack/typescript/{%packagejson%}
```diff
@@ -32,31 +32,29 @@
...
     "axios": "^0.19.0",
     "bootstrap-hot-loader": "^3.19.3",
-    "express": "~4.15.0",
-    "i18next": "^11.6.0",
-    "react": "16.8.6",
-    "react-dom": "16.8.6",
-    "react-i18next": "^7.11.0",
+    "express": "^4.17.1",
+    "i18next": "^19.1.0",
+    "react": "^16.12.0",
```
ranyitz: Please leave the react and react-dom version fixed, we want them to be the same ones you have on the `ejs` file.

sidoruk-sv:
> Hey @sidoruk-sv!
> 
> Thanks for updating those templates!
> 
> See the [note in the documentation](https://react.i18next.com/latest/trans-component#important-note)
> 
> > While the Trans components gives you a lot of power by letting you interpolate or translate complex react elements - the truth is - in most cases you won't need it.
> 
> We want to keep the templates small, let's remove the usage of `Trans` component from the PR.
> 
> Everything including the update of the React version is great.
> 
> In general we prefer keeping the PRs focused, so it would be a bit better to get update to React in one PR and translations with another.

I could agree until I see a lot of production code, where one sentence is broken into three or more translation keys, even if i18next is upgraded to the latest version.

So developers just don't look into i18next documentation and don't know about the ability to organize translation keys in a more meaningful way.

For example for our project, not too big, but we used this component 45 times and translators were impressed about the ability to understand the full sentence in keys for example `Advanced: <0>Transfer your domain</0> registration to Wix. <1>Learn More</1>` otherwise in Wix codebase it is usually devided into 3 different translation keys, and understanding of such keys becomes hard.

ranyitz:
@sidoruk-sv I agree that it can be useful, I'm not sure that the templates would be the best way to promote it.

sidoruk-sv:
@ranyitz agreed, found this issue:
https://github.com/wix-private/fed-handbook/issues/84

And seems like it is better to be explained there.

ranyitz:
@sidoruk-sv Sounds like a great idea.

sidoruk-sv:
@ranyitz so this PR is polished, your turn on accepting or commenting more)

sidoruk-sv:
@ranyitz i18next will fail if there would no Suspense, they rely on its presence.

PR #1930, opened 2020-02-06
prevent error when project is opened at the first time 

sidoruk-sv:
### 🔦 Summary

For now, when I open newly generated project I got a warning from WebStorm, that:
`'null' not allowed here` for `max_line_length`


packages/create-yoshi-app/templates/business-manager-module/javascript/.editorconfig
```diff
@@ -7,4 +7,4 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 end_of_line = lf
-max_line_length = null
+max_line_length = 120
```
yanivefraim: Thanks for the PR! 
I think that the default Prettier line number (that we use) is 80 (see [here](https://prettier.io/docs/en/options.html#print-width)). 
In this case, I believe that this is a better number. 
WDYT?

sidoruk-sv: @yanivefraim I think that in 2019 we could move to at least 120 symbols.

Found an article, which correlates with my thoughts)
https://hackernoon.com/does-column-width-of-80-make-sense-in-2018-50c161fbdcf6

Also in WebStorm, if we don't specify max_line_length, the default option is 120, not 80.

sidoruk-sv: For example question "how to change default 120 to 250"
https://stackoverflow.com/questions/24355976/how-to-change-line-width-in-intellij-from-120-character

ranyitz: @sidoruk-sv maybe we should move to 120 symbols and maybe not, let's keep this PR with the same behavior and discuss changing the line default length in a different issue.

sidoruk-sv: @ranyitz deal)
So I cleared that incorrect `max_line_length = null`, so it would work in the way as "use default until you will specify `max_line_length`"

PR #1931, opened 2020-02-06
create a webworker bundle that runs on server side

roymiloh:
**Why?**

- Content hash is not needed.
- Single chunk over multiple ones, as we do with the regular server bundle.
- Potentially different externals or any other configurations.

**Configuration:**

```
webWorkerServer: {
  entry: WebpackEntry;
};
```


packages/yoshi-flow-monorepo/src/scripts/build.ts
```diff
@@ -135,12 +136,25 @@ const build: cliCommand = async function(argv, rootConfig, { apps, libs }) {
...
+    }
+
     webpackManager.addConfigs(pkg.name, [
       clientDebugConfig,
       clientOptimizedConfig,
       serverConfig,
       webWorkerConfig,
       webWorkerOptimizeConfig,
+      webWorkerServerConfig,
+      webWorkerServerOptimizeConfig,
```
ronami: Not needed, not needed `min`.

packages/yoshi-flow-monorepo/src/scripts/build.ts
```diff
@@ -135,12 +136,25 @@ const build: cliCommand = async function(argv, rootConfig, { apps, libs }) {
...
 
+    let webWorkerServerConfig;
+    let webWorkerServerOptimizeConfig;
+
+    if (pkg.config.webWorkerServerEntry) {
+      webWorkerServerConfig = createWebWorkerServerWebpackConfig(pkg, {
+        isDev: true,
+      });
+
+      webWorkerServerOptimizeConfig = createWebWorkerServerWebpackConfig(pkg);
```
ronami: Same as bellow.

Issue #1408, opened 2019-06-27
rerun `yoshi start` on configuration change.


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior

When configuring Yoshi via `package.json`/`yoshi.config.js` during `yoshi start`, the user needs to manually press `ctrl+c` and then run `yoshi start` again so this new configuration will take effect.

### 🤔 Expected Behavior

When the developer changes a configuration that needs a rerun in order to take effect let's do automatically it for him.
 
### 💁 Possible Solution

During `yoshi start`, listen to changes on `package.json` `yoshi` field and `yoshi.config.js` files and rerun accordingly.

### 🔦 Context

I've seen many developers struggle when trying to change configuration options (like `ssl`) when they don't know that they need to close and rerun Yoshi.

We aim towards something you start once, and doesn't need to touch it again. This feature will take us one step closer.

saarkuriel:
Can I try and take this issue? @ranyitz @ronami

ranyitz:
@saarkuriel Yes of course! Let us know if you need any help :1st_place_medal:

ranyitz:
@Gongreg Asked me if he can take this issue, and suggested to run Yoshi as a subprocess so it will be easy to kill and restart it.

This is a good solution and was my initial goto as well.
The problem with running start in a different process is that we need to wait around 100ms more just because we now want to maybe restart the process (which is not very common thing to do).

I asked @ronami and we thought about how to do it without starting a new process. I wrote some pseudo-code that show how it can be done.

### The changes that needs to be done:

* Dev environment needs to expose a `stop` method, this stop method should be able to stop both [`webpack-dev-server` & `app-server-process`](https://github.com/wix/yoshi/blob/master/packages/yoshi-common/src/dev-environment.ts#L41-L42) which I think already has a `stop`/`end`/`kill` method.

* `Start` command needs to expose a way to stop it, a good way to do it is to return a function which will be called when we want to stop the command.

![image](https://user-images.githubusercontent.com/11733036/72439432-995c7800-37af-11ea-99a8-d7640c062aaf.png)

* The `yoshi-app` file can listen to changes on the config and in case it has been changed, stop the current dev environment and create a new one.

![image](https://user-images.githubusercontent.com/11733036/72439116-f60b6300-37ae-11ea-8a17-df370972e77f.png)

* We can also send a message to the screen that we're currently reloading the `dev-environment` which would be really nice.

[...]

PR #1932, opened 2020-02-06
Restart Dev Environment when yoshi config changes

Gongreg:
### 🔦 Summary

Issue: #1408 

Added a watcher for yoshi config file, which restarts the dev environment on config change.

### 🗺️ Test Plan

Added feature tests for watching config changes.


test/javascript/features/config-watcher/config-watcher.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const packageJSON = path.join(scripts.testDirectory, 'package.json');
+
+const initialPort = 3200;
+const updatedPort = 3300;
+
+describe.each(['dev'] as const)('Start command config watcher [%s]', mode => {
+  it('should restart on config change', async () => {
+    await scripts[mode](async () => {
+      await waitForPort(initialPort, { timeout: 20000 });
```
yanivefraim: this check is redundant since we are doing the same check inside scripts/dev:
https://github.com/wix/yoshi/blob/master/test/scripts.ts#L147

test/javascript/features/config-watcher/config-watcher.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      //Make sure that yoshi is fully loaded
+      await page.goto(`${scripts.serverUrl}`);
+
+      const data = JSON.parse(fs.readFileSync(packageJSON, 'utf-8'));
+      data.yoshi.servers.cdn.port = updatedPort;
+      fs.writeFileSync(packageJSON, JSON.stringify(data));
+
+      await waitForPort(updatedPort, { timeout: 20000 });
```
yanivefraim: It might be nicer to actually navigate to this port and see that we have the cdn up (looking for static assets in html). WDYT?

Gongreg: I've added navigating to cdn and checking that bundle is there.

test/javascript/features/config-watcher/config-watcher.test.ts
Gongreg: Removed the check.

Gongreg:
> Thanks for the PR, nice work! 👑
> 
> I added some comments regarding tests only (I did not review the entire PR).
> 
> I would also add a Typescript test.

I've added a typescript test, it just feel a bit redundant (copy paste).

yanivefraim:
> > Thanks for the PR, nice work! 👑
> > I added some comments regarding tests only (I did not review the entire PR).
> > I would also add a Typescript test.
> 
> I've added a typescript test, it just feel a bit redundant (copy paste).

Yes, this is something that we have right now for (almost) all of our features, but this is a price that we are willing to pay, for now.

Having said that, I hope that in the (near) future we will have a solution for that.

Gongreg:
> Yes, this is something that we have right now for (almost) all of our features, but this is a price that we are willing to pay, for now.
> 
> Having said that, I hope that in the (near) future we will have a solution for that.

No worries, I don't really mind it, it is better to have 2 copies than a complex setup.

I hope the build will pass now, since I was having some typescript issues on the weekend in jest-yoshi-preset package.

test/javascript/features/config-watcher/config-watcher.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+describe.each(['dev'] as const)('Start command config watcher [%s]', mode => {
+  it('should restart on config change', async () => {
+    await scripts[mode](async () => {
+      //Make sure that yoshi is fully loaded
+      await page.goto(`${scripts.serverUrl}`);
+
+      const data = JSON.parse(fs.readFileSync(packageJSON, 'utf-8'));
+
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
yanivefraim: Is this really necessary? It is a bad practice to wait for XXX ms, we always prefer to wait for something to happen/poll for changes (it can lead to flaky tests, and, in theory, this "flakiness" can happen to the user as well).

One possible (better) solution, is to log the "watching" event, and then the test (or `scripts.dev` command) can wait for this event. This way we know for sure that it is working.

Please let me know if I can help with that!

test/javascript/features/config-watcher/config-watcher.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      data.yoshi.servers.cdn.port = updatedPort;
+      fs.writeFileSync(packageJSON, JSON.stringify(data));
+
+      await waitForPort(updatedPort, { timeout: 20000 });
+      await page.goto('http://localhost:' + updatedPort);
```
yanivefraim: nit, use the newer syntax:

```suggestion
      await page.goto(`http://localhost: ${updatedPort}`);
```

test/typescript/features/config-watcher/config-watcher.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      // To make sure that watcher has started we await a little bit
+      await new Promise(resolve => setTimeout(resolve, 1000));
+
+      const data = JSON.parse(fs.readFileSync(packageJSON, 'utf-8'));
+      data.yoshi.servers.cdn.port = updatedPort;
+      fs.writeFileSync(packageJSON, JSON.stringify(data));
+
+      await waitForPort(updatedPort, { timeout: 20000 });
+      await page.goto('http://localhost:' + updatedPort);
```
yanivefraim: nit, use newer syntax:

```suggestion
      await page.goto(`http://localhost: ${updatedPort}`);
```

test/javascript/features/config-watcher/config-watcher.test.ts
Gongreg: I am fully aware that timeout is a terrible solution. Sadly as far as I can see, the server is fully started a bit later than the

```
Compiled successfully!

Your server is starting and should be accessible from your browser.

  Local:            http://localhost:3000/
  On Your Network:  http://192.168.132.174:3000/


Your bundles and other static assets are served from your dev-server.

  Local:            http://localhost:3200/
  On Your Network:  http://192.168.132.174:3200/


Note that the development build is not optimized.
To create a production build, use npm run build.
```

Text is displayed.

There is some additional text:

```
[SERVER]: (node:12631) DeprecationWarning: grpc.load: Use the @grpc/proto-loader module with grpc.loadPackageDefinition instead
```

That is displayed more or less when the server finishes loading, but depending on it doesn't seem like a good solution too.

Do you have any suggestions how could I wait for the server to fully load?

Gongreg: Ignore the above message, it seems like the timeout is redundant. I might have gotten red test because of unrelated thing.

yanivefraim: Yes, I was just looking at your code, I think that you can safely remove this wait (this is the reason for me asking if you had a reason for this) 😄

Gongreg: 1 in ~10 test runs fail. I think the file watcher is not fast enough to start.

A solution could be to keep changing package.json (most of the times it would be 1 change, but it might require another change if the watcher wasn't fast enough) till I get a message that the watcher actually saw a change.

yanivefraim: Seems like this is a different problem (you say that the watcher does not know about a change). In this case, the following line:
```
      await waitForPort(updatedPort, { timeout: 20000 });
```

was supposed to do the trick.
Anyway, it is important to understand where is the problem and fix it (maybe waiting for a dedicated log). Do you have a ling to a failing test?

Gongreg: I will move the watcher setup before we start the server itself, that way there shouldn't be any additional delay. If there is then there is some additional issue that I will have to investigate.

Gongreg: Update:

Finally managed to reproduce flaky test with DEBUG mode on.

packages/yoshi-common/src/dev-environment.ts

```
    // Start Webpack compilation
    await webpackDevServer.listenPromise();
				console.log('STEP 1');
    await compilationPromise;

		console.log('STEP 2');

    // start app server
    await serverProcess.initialize();

		console.log('STEP 3');
    const actualStartUrl = suricate
      ? getTunnelUrl(appName)
      : startUrl || 'http://localhost:3000';

    openBrowser(actualStartUrl);
```

test/javascript/features/config-watcher/config-watcher.test.ts

```
      //Make sure that yoshi is fully loaded
      console.log("TEST 1");
      await page.goto(`${scripts.serverUrl}`);

      console.log("TEST 2");

      const data = JSON.parse(fs.readFileSync(packageJSON, "utf-8"));
      data.yoshi.servers.cdn.port = updatedPort;
      fs.writeFileSync(packageJSON, JSON.stringify(data));

      console.log("TEST 3");
      await waitForPort(updatedPort, { timeout: 20000 });
```


Output:
```
    Starting development environment...

    STEP 1

    Compiled successfully!

    Your server is starting and should be accessible from your browser.

      Local:            http://localhost:3000/
      On Your Network:  http://192.168.132.174:3000/


    Your bundles and other static assets are served from your dev-server.

      Local:            http://localhost:3200/
      On Your Network:  http://192.168.132.174:3200/


    Note that the development build is not optimized.
    To create a production build, use npm run build.

    STEP 2

    TEST 1

    TEST 2

    TEST 3

    STEP 3

```

So the server is started and port is opened before `await serverProcess.initialize();` finishes running.

Checking now why it happens.

Gongreg: Adding `await page.$eval("#loaded", () => {});` after waiting for the port does not help. That means there is an issue inside `await serverProcess.initialize();` Going in deeper now.

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +21,45 @@ import {
...
+
+  const configInfo = searchForConfig();
+  if (configInfo) {
+    watch({ pattern: configInfo.filepath }, async () => {
+      const newConfig = loadConfig();
+
+      const isYoshiConfigChanged = !isYoshiConfigEqual(config, newConfig);
+
+      // Only restart if it is not being restarted at the moment
+      if (stopCommand && isYoshiConfigChanged) {
```
ranyitz: `// Only restart if it is not being restarted at the moment`

This looks to me like an edge case, If it's not critical I would put it in a later PR

Did you suffer from it?

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +21,45 @@ import {
...
+  return isEqual(omit(oldConfig, 'pkgJson'), omit(newConfig, 'pkgJson'));
+}
+
+const configAwareStartCommand: cliCommand = async function(
+  forwardedArgs,
+  initialConfig,
+) {
+  let config = initialConfig;
+
+  let stopCommand: (() => Promise<void>) | null = await start(
```
ranyitz: When stopping and starting there are some problems that occur from the fact that it takes time for the process to be fully killed. This was taken cared of here - https://github.com/Gongreg/yoshi/blob/yoshi-config-watcher/packages/yoshi-common/src/server-process.ts#L119-L134

I suggest we'll move from `stop` + `start` into `restart`.

the restart function will return another `restart` function etc...

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +21,45 @@ import {
...
+    forwardedArgs,
+    config,
+  );
+
+  const configInfo = searchForConfig();
+  if (configInfo) {
+    watch({ pattern: configInfo.filepath }, async () => {
+      const newConfig = loadConfig();
```
ranyitz: We are loading the config here during watch mode. It means that it will throw and the process will exit because of it. I find this experience suboptimal.

Instead, we can `try` loading the config, `catch` the error and log to the user that there is an error in the `config`/`package.json` parsing.

The user should be able to fix it and it should automatically run again.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Yes, if I did multiple changes I couldn't stop a server which was still booting up.

ranyitz: Ok, so I can suggest that we'll create a boolean parameter (`let restarting = false`) which will be changed to `true` before `await restart()` and change back to `false` after.

I think that it's a bit better than using the `cleanup` function for storing this information

Gongreg: alright.

Gongreg: yet the cleanup function can be null according to the type (it is while the first server start has not finished loading), so then we either have to do 
```
if (!isRestarting && stopCommand && isYoshiConfigChanged) {
}
```

or overcome null check with !.

We need to start config watcher before running startCommand to make sure that we don't get flaky tests.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: How would you recommend to approach it? Right now I am recreating the whole DevEnvironment since there are multiple places that require updated config.

I've updated stop command to be aware of child process.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Great suggestion. Updated according to it.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Added null removal `stopCommand!()`

Gongreg:
@ranyitz, I've added additional tests for the functionalities you talked about.

These issues are remaining:

1. Even with using your recommended restart implementation, I am still sometimes getting issue in tests when starting and stopping the servers. Sometimes the Promise is resolved later than the page is loaded and I get an error in the tests.

2. Before there were a lot of failing tests in CI even though they were passing for me locally. (Might be related to #1 since then when test fails the server keeps hanging and it might interrupt other tests).

I would really any suggestions for consistent server boot/stop in the tests. In real use case scenarios this issue will most likely not appear since almost nobody will be changing the configs at exact same moment the server will start.

But I also don't feel like patching up the tests to change configs multiple times till it catches a change is also a good solution.

Gongreg:
Sorry. Missclick.

I've added additional output message after the start command is fully executed `Development environment fully loaded!`. This way I can fully sync the test child process with the test itself.

packages/yoshi-flow-app/src/scripts/start.ts
ranyitz: You are right.
Let's improve the `end`/`kill` method of the server process. Using the same technic/hack introduced here - [serverProcessRestart](https://github.com/Gongreg/yoshi/blob/master/packages/yoshi-common/src/server-process.ts#L119-L134). (Using an interval to verify that the server is down)

Gongreg: Already did that :)

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +20,57 @@ import {
...
+  let stopCommand: (() => Promise<void>) | null = null;
+  let isRestarting = false;
+
+  watch({ pattern: ['package.json', 'yoshi.config.js'] }, async () => {
+    let newConfig: Config | null = null;
+
+    try {
+      newConfig = loadConfig();
+    } catch (e) {
+      console.log('There was an error trying to parse Yoshi config');
```
ranyitz: Would be nice to see the original error (that shows where is the problem, filename + location in the json)

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +20,57 @@ import {
...
+      return;
+    }
+
+    const isYoshiConfigChanged = !isYoshiConfigEqual(config, newConfig);
+
+    // Only restart if it is not being restarted at the moment
+    if (!isRestarting && isYoshiConfigChanged && stopCommand) {
```
ranyitz: We don't need to do it in this PR, but thinking about it, [debounce](http://demo.nimius.net/debounce_throttle/) would be the best experience.

Right now we can "lose" user changes just because the server is restarting.

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +20,57 @@ import {
...
+
+      await stopCommand();
+      stopCommand = await startCommand(config);
+      isRestarting = false;
+      postStart();
+    }
+  });
+
+  stopCommand = await startCommand(config);
+  postStart();
```
ranyitz: I'm not sure why we need `postStart`

Also, if we want to log something after start, we can just put it inside the start function.

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +20,57 @@ import {
...
+      config = newConfig;
+
+      await stopCommand();
+      stopCommand = await startCommand(config);
+      isRestarting = false;
+      postStart();
+    }
+  });
+
+  stopCommand = await startCommand(config);
```
ranyitz: We can run this command when we [initialize the `stopCommand` variable](https://github.com/wix/yoshi/pull/1932/files#diff-7a1a38ab58d77e6643cbb3a7541881a9R38)

ranyitz: It worth to add a comment about the fact that it's here (probably because you want to start the watcher before we start...)

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Good point, I will add it.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Removed this call.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Adding debounce.

packages/yoshi-flow-app/src/scripts/start.ts
Gongreg: Added error output.

Gongreg:
One additional optimization could be to not restart serverProccess when it is not needed, but that would require a bit of config checking (only restart when appName changes), making start.ts aware of it (would require to keep reference to existing serverProcess) and just feel like it is not worth it for now.

Gongreg:
The tests are failing because of: Error: ENOSPC: System limit for number of file watchers reached.

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +20,59 @@ import {
...
+      console.log('Changes in the Yoshi config detected.');
+
+      config = newConfig;
+
+      await stopCommand();
+      stopCommand = await startCommand(config);
+      isRestarting = false;
+    } else {
+      //It means that the server is not ready or is restarting at the moment and we should try again later.
+      configWatcher();
```
ranyitz: Can we try again after a reasonable timeout? Maybe 200ms?

Gongreg: Sure, any timeout is okay :)

packages/yoshi-flow-app/src/scripts/start.ts
```diff
@@ -14,7 +20,59 @@ import {
...
+    }
+  }, 1000);
```
Gongreg:
```suggestion
  }, 200);
```

Gongreg:
The last remaining issue is CI. I am still getting watcher limit error. Do you have any suggestions?

Gongreg:
After talking yesterday with @ranyitz, we've decided to close this PR and the issue.

The complexity we get by introducing this feature far outweighs any benefit we get from it. There are only few config options that could be actually useful, but if we watched only few of those options this functionality will be not self evident. The best thing to do now is to learn from this case and make sure we avoid situation like this in the future.

PR #1943, opened 2020-02-10
Yoshi server: add `bo-auth` middleware

yanivefraim:
1. Adding a wrapper for `bo-auth` middleware.
2. Added typing for `this.context` (`BootstrapContext` from `@wix/wix-bootstrap-ng/typed`), so we will have type completion 

TODO:
- tests
- refactor/code reuse with `require-login`


packages/yoshi-server/src/types.ts
```diff
@@ -1,5 +1,6 @@
 import { Request, Response } from 'express';
 import { JsonValue } from 'type-fest';
+import { BootstrapContext } from '@wix/wix-bootstrap-ng/typed';
```
ronami: I think we should add a small note on the PR for this.

packages/yoshi-server/package.json
```diff
@@ -30,6 +30,7 @@
     "serialize-error": "4.1.0",
     "sockjs-client": "1.4.0",
     "type-fest": "0.8.1",
+    "wix-bootstrap-bo-auth": "^1.0.1888",
```
ronami: Probably pretty small but everyone will need to install that (while only a few will use it). We can later on think how we can optimize it.

PR #1947, opened 2020-02-11
add an ability to modify the public path

roymiloh:
By using `YOSHI_PUBLIC_PATH`, our users will be able to modify the public path in their bundle. For now we'd like to "hide" this feature, so it's being configured by env variable instead of a proper configuration in yoshi config file.


test/javascript/features/public-path/public-path.test.ts
```diff
@@ -0,0 +1,52 @@
...
+          interceptedRequest.respond({ body });
+        } else {
+          interceptedRequest.continue();
+        }
+      });
+    });
+
+    afterEach(() => jestPuppeteer.resetPage());
+
+    afterAll(() => page.setRequestInterception(false));
```
ronami: Is this needed? (I'm asking since we're calling `jestPuppeteer.resetPage()`)

test/scripts.ts
```diff
@@ -250,14 +258,9 @@ export default class Scripts {
...
   async prod(
     callback: TestCallbackWithResult = async () => {},
-    args: Array<string> = [],
+    opts: ScriptOpts = {},
   ) {
-    let buildResult;
-    try {
-      buildResult = await this.build(ciEnv, args);
-    } catch (e) {
-      throw new Error(e);
```
ronami: This looks confusing. I'm not sure if it's needed for something. @yanivefraim what do you think?

test/scripts.ts
```diff
@@ -31,20 +31,27 @@ export type ProjectType =
...
+  private readonly yoshiPublishDir: string;
+  public readonly staticsServerUrl: string;
 
   constructor({ testDirectory }: { testDirectory: string }) {
     this.verbose = !!process.env.DEBUG;
     this.testDirectory = testDirectory;
     this.serverProcessPort = 3000;
     this.staticsServerPort = 3200;
     this.serverUrl = `http://localhost:${this.serverProcessPort}`;
+    this.staticsServerUrl = `http://localhost:${this.staticsServerPort}`;
```
ranyitz: Note that it can be `https` in case the user uses `--ssl` or `https` in the config.

ranyitz: Do you use in somewhere?

PR #1956, opened 2020-02-12
add the spaced-comment lint rule for the monorepo

ranyitz:
### 🔦 Summary

Linting and formatting suppose to help reducing noise in code reviews and let us focus on what's the code is doing rather than on how it looks.

This rule will help the consistency of the codebase while being autofixable and therefore not harmful.

I've configured it specifically to support [this usecase](https://github.com/typescript-eslint/typescript-eslint/issues/600#issuecomment-499979248) of the `///` triple slash comment in `d.ts` files


ranyitz:
I also think that we should add it to `eslint-config-yoshi`. It would be breaking and therefore we can add it as a warning and promote it to error on version 5.

@Gongreg what do you say?

Gongreg:
@ranyitz, if we are thinking to add more breaking changes regarding eslint in v5 then it is a good time to so. Also this specific case is going to be autofixable so there is no harm to add it for a breaking change.

ranyitz:
@Gongreg Agree, we can also add it right now as a warning. It would not break other projects and actually, we don't need to change it into an error because it's not critical.

Do you wanna create a PR?

Gongreg:
Lets create an issue for it now. Stuck on another task atm.

Issue #1259, opened 2019-04-30
Deprecate `--entry-point` and `--server` on `yoshi start`


Schniz:
# 🙋 Feature Request (Breaking change! 💥 )

Deprecate `--entry-point=""` and `--server=""` in `yoshi start`, by using a hard-coded value: `index-dev.js`

### 😯 Current Behavior

`client` template now has the `start` calling `yoshi start --server=dist/server.js`:
https://github.com/wix/yoshi/blob/6ac24ae87243a27a18e4c94c6b330cf6c7768c4d/packages/create-yoshi-app/templates/client/javascript/%7B%25packagejson%25%7D#L11

`fullstack` template uses `index-dev.js`

That makes it hard to know which file to execute in order to start the server. We don't see any use case for having it as a configuration, and we'd like to remove ambiguity between templates and projects.

### 🤔 Expected Behavior

All projects will have `index-dev.js` that will be responsible on starting the project with testkits, etc:

* Projects that use plain `yoshi start`, can have `index-dev.js` just require `index.js`
* Client projects can move `dev/server.js` to `index-dev.js`
* Fullstack projects are already okay!

### 🔦 Context 

This is for implementing #914, `yoshi serve`, which is a stripped-down, "production-like" version of `start`.

cc @ranyitz

ranyitz:
Nice!
I want to add some explanation to **why this is needed.**

In general, if an option is **not exclusive to one command** and does **not need to be changed between different runs**, it should be in the config.
______

Specifically here, doing this change will enable us to use `yoshi-config` in order to understand how we should run the server, and it will give us an easy way to figure out how to run `start` in the closest thing to production mode, or `yoshi serve`, without requiring the user to specify another command in his/her `scripts` property.

Schniz:
Do we want to change this issue to `deprecate --server and --entrypoint`?

Schniz:
So `yoshi start` and `yoshi serve` will both (pseudo code):

```js
if (fs.existsSync('<projectRoot>/index-dev.js')) {
  spawn("node index-dev.js");
} else if (fs.existsSync('<projectRoot>/index.js')) {
  console.error("Deprecation warning: The next version of Yoshi won't run `index.js` files automatically. Please add an `index-dev.js` file");
  spawn("node index.js");
} else if (entrypoint || server) {
  console.error("Deprecation warning: --entrypoint and --server will be removed on the next Yoshi version");
  spawn(`node ${entrypoint || server}`);
} else {
[...]

PR #1960, opened 2020-02-13
Deprecating --entry-point and --server flags

Gongreg:
### 🔦 Summary

Issue: #1259 

At first the entry helper will try to check if --entry-point or --server flag was passed. If it was it will notify about the deprecation.

Then it checks index-dev.(t|j)s (prioritizes ts file).

If none is found it tries to use index.(t|j)s file. If found notifies about deprecation.

It nothing was found an error is thrown.

I've tried to make sure that this deprecation does not break existing functionality.

I've updated the create-yoshi-app templates to make sure that new project do not get any deprecation warnings.

**Important: Didn't work on create-yoshi-app editor template. **

### 🗺️ Test Plan
Added unit tests for the new entry resolver. Updated feature tests to use new index-dev.js entry.


packages/yoshi-flow-legacy/src/commands/start.js
```diff
@@ -61,7 +62,14 @@ const addJsSuffix = suffix('.js');
 const shouldRunTests = cliArgs['with-tests'] === true;
 const debugPort = cliArgs.debug;
 const debugBrkPort = cliArgs['debug-brk'];
-const entryPoint = addJsSuffix(cliArgs['entry-point'] || 'index.js');
+const entryPointCLI = cliArgs['entry-point'];
+
+let entryPoint = 'index.js';
+try {
+  entryPoint = getServerEntry(
```
Gongreg:
```suggestion
  // Legacy flow can start while missing server entry.
  entryPoint = getServerEntry(
```

packages/yoshi-helpers/src/server-entry.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (serverEntryCLI) {
+    console.log(
+      chalk.yellow(
+        `Deprecation warning: --server and --entry-point is going to be removed in Yoshi v5. Please create 'index-dev.${extension}' in your root directory instead of using these flags.`,
+      ),
+    );
+    return serverEntryCLI;
+  }
+
+  const indexDevTS = path.resolve('index-dev.ts');
```
Gongreg: I will look into glob search on Monday, if it allows us to prioritize TS file search before js then I will replace fs.exists with globber, should simplify the code

Gongreg:
Not sure what label to put on.

docs/api/cli.md
```diff
@@ -28,13 +28,13 @@ This will run the specified (server) `entryPoint` file and mount a CDN server.
 
 ### options
 
-#### `--entry-point` ( `-e` )
+#### `--entry-point` ( `-e` ) (deprecated!)
```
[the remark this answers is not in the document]
ranyitz: Or something like that 💅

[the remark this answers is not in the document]
Gongreg: Alright, lets make it pretty

test/scripts.ts
```diff
@@ -276,7 +276,7 @@ export default class Scripts {
       },
     );
 
-    const appServerProcess = execa('node', ['./index.js'], {
+    const appServerProcess = execa('node', ['./index-dev.js'], {
```
Gongreg:
```suggestion
    const appServerProcess = execa('node', ['./index.js'], {
```

ranyitz:
@Gongreg regarding the labels.

We use them in order to automatically generate a changelog, the PRs which are relevant specified in the [`lerna.json`](https://github.com/wix/yoshi/blob/6c83e0124d0f2a3ecabca0218810ec0e9bdd5733/lerna.json#L8-L15)

`PR: chore` is something that we've added because we didn't want some PRs to be on the changelog, like automatic version bump or lock file maintenance.

I do believe that a PR that includes a deprecation has a place in the changelog because it affects users.

I couldn't find a good label as well, so i'm suggestion on adding a new one. What do you say on: 

PR: Deprecation ⚠️

If we agree on it, let's add it as a label and create another PR that adds it to the `lerna.json`.

packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -13,7 +13,7 @@ import { getDevServerUrl } from './utils/suricate';
 
 const isDevelopment = process.env.NODE_ENV === 'development';
 
-const possibleServerEntries = ['./server', '../dev/server'];
+const possibleServerEntries = ['./server', '../dev/server', '../index-dev'];
```
ranyitz: Maybe i'm missing something, but I don't think that `index-dev` should be a `server entry`, those are entries for webpack to bundle. The `index-dev` should bootstrap the environment and run the server using a node platform testkit.

packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -151,8 +151,8 @@ function validateServerEntry({
   if (!serverEntry && !yoshiServer) {
     throw new Error(
       `We couldn't find your server entry. Please use one of the defaults:
-          - "src/server": for a fullstack project
-          - "dev/server": for a client only project`,
+      - "index-dev": for a fullstack project
```
ranyitz: `src/server` is still a viable server entry. (see comment above)

packages/yoshi-helpers/src/server-entry.ts
```diff
@@ -0,0 +1,56 @@
...
+      deep: 0,
+      cwd: path.resolve(process.cwd(), 'dev'),
+    });
+
+    if (devServer.length > 0) {
+      return path.join('dev', devServer[0]);
+    }
+  } catch (e) {}
+
+  const indexDev = globby.sync(['index-dev.ts', 'index-dev.js'], { deep: 0 });
```
ranyitz: Turn into a loop

website/versioned_docs/version-4.x/guides/app-flow.md
```diff
@@ -151,24 +151,10 @@ Start by opting into app flow by changing your `package.json` or `yoshi.config.j
 }
 ```
 
-In app flow, Yoshi looks for the entry file of your server at `/dev/server.(js|ts)`. Move your current local dev server to that location. For example, if your local dev server is in `index.js`:
+In app flow, Yoshi looks for the entry file of your server at `/index-dev.(js|ts)`. If your local dev server is in `index.js`, then you can call it from `index-dev.js`:
```
ranyitz: remove the whole change - it should be autogenerated

PR #1961, opened 2020-02-15
[create-yoshi-app] [client] upgrade 18next and change client templates

sidoruk-sv:
# 🔦 Summary

Same as #1929


sidoruk-sv:
Failed test `test:templates:server` is not related to the current changes

PR #1966, opened 2020-02-16
Fix website publishing: enforce registry to be the public registry for the website

ranyitz:
### 🔦 Summary
The travis website publishing is now broken due to private dependencies on the `yarn.lock` file of website directory.

This PR changes the lock file to contain only public registry locations and enforce it using `.yarnrc` registry.


ranyitz:
image

PR #1973, opened 2020-02-17
fixed bug where empty dependencies won't update

amiryonatan:
### 🔦 Summary
Intro:
This script is syncing between dependencies in package.json into tsconfig.
The scenario that is broken is as follows:
If a package had dependencies before, and now it doesn't the tsconfig dependencies won't update as we don't write empty arrays.
In my solution, we compare initial and final tsconfig, and if a change was introduced, we overwrite the file


amiryonatan:
@ronami don't forget this

packages/yoshi-flow-legacy/src/commands/utils/index.js
```diff
@@ -1,55 +0,0 @@
...
-      }
-    }),
-  );
-
-  console.log();
-};
-
-module.exports = {
-  verifyTypeScriptReferences,
-};
```
ronami: Dead code we probably forgot to delete when we moved to using flows.

amiryonatan:
@ronami how do we merge this ?

PR #1974, opened 2020-02-17
Yoshi server testing

yanivefraim:
Added a test for `yoshi-server-testing`. 

TODO: Add the tests as part of Yoshi Server tests


test/javascript/features/yoshi-server/api-client-server-react/src/components/App/App.spec.js
```diff
@@ -0,0 +1,31 @@
...
+import ReactDOM from 'react-dom';
+import HttpClient from 'yoshi-server-testing';
+import { HttpProvider } from 'yoshi-server-react';
+import eventually from 'wix-eventually';
+import App from './App';
+import { greet } from '../../api/greeting.api';
+
+const mocks = [
+  {
+    request: { fn: greet, variables: ['Yaniv'] },
```
ronami: Not necessarily related to the PR but we changed this to `method`, I think can change this too. Might be a breaking change for the small number of users.

yanivefraim: Cool, I'll open a new issue for that

Issue #1409, opened 2019-06-27
Remove `--ssl` and `--https` cli options 


ranyitz:
# 🐛 Bug Report

### 😯 Current Behavior

There are currently two ways to configure `webpack-dev-server` (e.g. local CDN) to serve assets on `https`/`ssl`.

One way is via a command line parameter (`--ssl`/`--https`) and the other is via the Yoshi config `servers.cdn.ssl`.

The first (with CLI parameters) is confusing because it implies that the app server will run on `https` as well (while that is not true).

Another problem is that Webpack public path is configured via the Yoshi config `servers.cdn.ssl` option, and therefore does not configure correctly (causing dynamic imports and other assets managed by Webpack to fail.)

https://github.com/wix/yoshi/blob/07542070cad3f0eab57e3cf75c4acf6b22363d0b/packages/yoshi/config/webpack.config.js#L83

### 🤔 Expected Behavior

There is one way to configure your CDN to run over `https`. preferably using yoshi config.

Gongreg:
Hey @ranyitz,
I could work on removing the ssl/https flags from cli.
Wouldn't that theoretically be a breaking change?
But as you said the https/ssl option is redundant.
Maybe it should be simply ignored in v4 and removed in v5?

ranyitz:
@Gongreg Thanks for your offer to help with it!

You are correct, that would be a breaking change.
It is currently targeted as a feature for v5, we have a milestone where are keep track on those features.

When we'll start working on version 5 (hopefully soon) we'll create a branch for it and we'll start adding those breaking changes to it while keeping it up to date with the current version.

After the release we'll merge this branch with master and continue development on it.

I'll let you know when it'll be relevant to work on this feature! 🙏

Gongreg:
Pr is released.

PR #1978, opened 2020-02-19
Removing --ssl --https options

Gongreg:
Issue: #1409

### 🔦 Summary

Removed --ssl and --https options, since to properly use https for CDN you either have to manually set cdn.url or enable cdn.ssl. So the cli flags are redundant and not working as a standalone option.

### 🗺️ Test Plan

Removed redundant test, tests testing ssl are already in place.


ranyitz:
@Gongreg So i'm waiting with merging this. But I think that it's ok to merge this to version 4 with a notice in slack (it's not like we're breaking any functionality or that builds will failed due to this change)

Gongreg:
Hey. Alright. I just wanted to have it ready for version 5. :)

Issue #914, opened 2019-01-16
Add a way to serve the application just like in production (yoshi serve)


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior

Emulating production is difficult, there are multiple steps required, which are not necessarily known to the user.

### 🤔 Expected Behavior

There is a simple command, that serves the application just like in production.

### 💁 Possible Solution
Implement `yoshi serve` command.

The command serves the bundled application (from `dist/statics/`) similarly to the CDN in production. If the bundle is not there, it will show a message that `yoshi build` needs to be run.

### 🔦 Context 

There are multiple requests for such feature, that enables testing things just like in production, which `yoshi start --production` does not provide.

Schniz:
This is crucial for us as well, when testing performance for Yoshi apps. I'd like to do `yoshi build && yoshi serve`, and then I won't have any traces of development mode like HMR

moshfeu:
Still no updates? :(

ranyitz:
Help wanted and PRs welcome 😃

Schniz:
Maybe I'll have time for it tomorrow
because I need it too :D

cc @cowchimp

Schniz:
I have written our own, stripped down `yoshi serve`:

* run `yoshi build`
* if `index-dev.js` is found = fullstack project:
  * `env <YOSHI_BOOTSTRAP_ARGUMENTS> node index-dev.js`
* else if `dev/server.js` is found = client project:
  * `env <YOSHI_BOOTSTRAP_ARGUMENTS> node dist/server.js`
* else
  * fail 😄 

This is what @ranyitz [have replied](https://github.com/wix-private/perfer/pull/11#issuecomment-488648739):

> I want to add that I agree, and something like that will work great as a Yoshi feature.
>
> Our only problems are with the --entry-point/--server options.
>
> We may be able to find those [by] parsing package.json's scripts.start and extracting the option (I don't believe that I'm saying this). (It's a bit better than creating a convention we don't really respect (index-dev.js/dev/server.js))
>
> If we do something like that for the sake of zero config, we should also show a deprecation warning that suggests that the user need to switch from using --entry-point to a config option.

Schniz:
I'm filling the blanks here. This feature is _harder_ and _hackier_ (is that a word?) than the simple case I thought it would be.

There are more open questions:

[...]

PR #1980, opened 2020-02-19
Yoshi serve command

Gongreg:
Issue: #914.

### 🔦 Summary

Added serve command to yoshi-flow-app. It runs dev server in production mode and also starts cdn to serve statics created with yoshi build.

### 🗺️ Test Plan

Added feature tests for yoshi serve (to check if it starts and if it properly checks for missing build).

Added unit tests for packageJSON scripts parser (yoshi serve uses same entry point as the one it finds in yoshi start).


packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const serverProcess = await ServerProcess.create({
+    serverFilePath,
+    suricate: config.suricate,
+    appName: config.name,
+  });
+
+  await Promise.all([serverProcess.initialize(), startCDN(config)]);
+
+  const actualStartUrl = config.suricate
```
Gongreg: Duplicated code from devEnvironment.

packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,68 @@
...
+        --debug         Allow app-server debugging
+        --debug-brk     Allow app-server debugging, process won't start until debugger will be attached
+        --url           Opens the browser with the supplied URL
+    `,
+    );
+
+    process.exit(0);
+  }
+
+  console.log(chalk.cyan('Starting the environment...\n'));
```
Gongreg: Maybe we could add a better message?

packages/yoshi-flow-app/src/bin/yoshi-app.ts
```diff
@@ -75,20 +76,22 @@ Promise.resolve().then(async () => {
...
   await verifyDependencies();
 
+  const config = loadConfig();
+
   if (command === 'start') {
     process.env.NODE_ENV = 'development';
     process.env.BABEL_ENV = 'development';
+  }
 
+  if (command === 'start' || command === 'serve') {
```
Gongreg: I found it strange, but I couldn't set the env variables inside the command itself.

It seems like there is some code which is imported before running the command that depends on the variables.

I didn't investigate it too much, but I don't like that inner details of commands is leaking here.

packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,81 @@
...
+    console.log(
+      chalk.red(
+        `${config.clientFilesPath} is missing. Run \`yoshi build\` and try again.`,
+      ),
+    );
+
+    process.exit(1);
+  }
+
+  console.log(chalk.cyan('Starting the environment...\n'));
```
Gongreg: A better message would be better

ranyitz:
Is it still a work in progress? If no, let's change the title.

In any case It will take some time for me to go over it, so i'm sorry in advance 😭

packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,81 @@
...
+    appName: config.name,
+  });
+
+  await Promise.all([serverProcess.initialize(), startCDN(config)]);
+
+  const actualStartUrl = config.suricate
+    ? getTunnelUrl(config.name)
+    : url || 'http://localhost:3000';
+
+  openBrowser(actualStartUrl);
```
ranyitz: remove open browser

packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,81 @@
...
+    suricate: config.suricate,
+    appName: config.name,
+  });
+
+  await Promise.all([serverProcess.initialize(), startCDN(config)]);
+
+  const actualStartUrl = config.suricate
+    ? getTunnelUrl(config.name)
+    : url || 'http://localhost:3000';
+
```
ranyitz: Add log when the both are ready (cdn, app-server), something like.

Use the same messaging as the devEnvironment logger. - https://github.com/wix/yoshi/blob/master/packages/yoshi-common/src/dev-environment-logger.ts#L22

Separate for dev-server and app-server.

packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,81 @@
...
+        `${config.clientFilesPath} is missing. Run \`yoshi build\` and try again.`,
+      ),
+    );
+
+    process.exit(1);
+  }
+
+  console.log(chalk.cyan('Starting the environment...\n'));
+
+  const serverProcess = await ServerProcess.create({
```
ranyitz: either implement `prodcutionServerProcess` or extend the server process to not run anything "dev" related (e.g. socket server for hmr)

packages/yoshi-common/src/server-process.ts
```diff
@@ -91,46 +81,93 @@ export default class ServerProcess {
...
   }
 
-  onMessage(response: any) {
-    this.resolve && this.resolve(response);
+  async close() {
+    // @ts-ignore
+    if (this.child && this.child.exitCode === null) {
+      this.child.kill();
+
+      await new Promise(resolve => {
```
ranyitz: Nice catch!

packages/yoshi-common/src/server-process.ts
```diff
@@ -25,42 +25,33 @@ function notUndefined<T>(x: T | undefined): x is T {
...
   public appName: string;
 
   constructor({
-    cwd,
+    cwd = process.cwd(),
     serverFilePath,
-    socketServer,
-    suricate,
     appName,
+    nodeEnv = {
```
ranyitz: I think that this variable name would be better named `env` => all of the environment parameters for the process, in oppose to `nodeEnv` which has a disambiguation with `NODE_ENV`

packages/yoshi-flow-app/src/scripts/serve.ts
```diff
@@ -0,0 +1,45 @@
...
+      );
+      return;
+    }
+
+    const serverProcess = new ServerProcess({
+      serverFilePath,
+      appName: config.name,
+    });
+
+    const servers = await Promise.all([
```
ranyitz:
```suggestion
    const [, cdn] = await Promise.all([
```

Would be a bit nicer

test/typescript/features/serve/https/serve.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  templateDir: __dirname,
+  projectType: 'typescript',
+});
+
+describe('Yoshi serve https', () => {
+  it('should return prod build when running yoshi serve', async () => {
+    await scripts.build();
+
+    await scripts.serve(async () => {
+      const browser = await launch({ ignoreHTTPSErrors: true });
```
ranyitz: Awesome

packages/yoshi-common/src/server-process.ts
Gongreg: Good point. Will fix it.

packages/yoshi-flow-app/src/scripts/serve.ts
Gongreg: ah, I am used to const [_, cdn] = await Promise.all([ and it didn't like it. Thanks

Gongreg:
@ranyitz Fixed up the final comments.

Regarding the documentation maybe we should create a separate section in yoshi docs? It depends if we are going to have more node API exposed in the future.

PR #1996, opened 2020-02-23
Yoshi Server - change the api

yanivefraim:
We want to support custom arguments to the `yoshi-server-client` ` request` method. This change of api adds the `headers` option.


packages/yoshi-server-client/src/index.ts
```diff
@@ -29,10 +34,15 @@ export default class implements HttpClient {
...
-  async request<Result extends FunctionResult, Args extends FunctionArgs>(
-    { fileName, functionName }: DSL<Result, Args>,
-    ...args: Args
-  ): Promise<UnpackPromise<Result>> {
+  async request<Result extends FunctionResult, Args extends FunctionArgs>({
+    method: { fileName, functionName },
+    args,
+    headers = {},
+  }: {
+    method: DSL<Result, Args>;
```
ronami: We might want to change the wrapper function name from `fn` to `method` for consistency.

PR #2004, opened 2020-02-24
yoshi-create-app: server templates cleanups

hugebdu:
- using TestEnvBuilder#withMainAppConfigEmitter
- removed unused dependences


packages/create-yoshi-app/templates/server/javascript/test/environment.js
```diff
@@ -1,24 +1,13 @@
 // https://github.com/wix-platform/wix-node-platform/tree/master/bootstrap/wix-bootstrap-testkit
 const testkit = require('@wix/wix-bootstrap-testkit');
-// https://github.com/wix-platform/wix-node-platform/tree/master/config/wix-config-emitter
-const configEmitter = require('@wix/wix-config-emitter');
 
 const TestEnv = require('@wix/wix-test-env');
 
 export const env = TestEnv.builder()
-  .withConfigurer(emitConfigs)
+  .withMainAppConfigEmitter(builder => builder) // empty configuration for now
```
ranyitz: Do we have to pass identity function (`builder => builder`) to it?

hugebdu: @ranyitz you have to pass identity function.
it is a bit awkward since the `erb` config file doesn't contain any expressions to fill in.
even the petri scopes are hardcoded.

hugebdu: I can make the argument optional.
in that case the config emitter will be used only to copy & rename, without any erb processing

will it make more sense?

ranyitz: I think that it makes sense it there is a real usecase for it.
In case everyone needs to consume something from the erb (in a regular production app) it worth showing an example to that in the base template

I'm good with any decision that you make there

hugebdu:
@ranyitz 

> So sourceFolders: ['./templates'] & targetFolder: './target/configs' are preconfigured?

indeed. and even respecting [jest workers](https://github.com/wix-platform/wix-node-platform/tree/master/testing/wix-test-env#jest-workers-support)

packages/create-yoshi-app/templates/server/javascript/test/environment.js
hugebdu: @ranyitz does [this](https://github.com/wix/yoshi/pull/2004/commits/76c356bb310223dd2ea6c90f943bae1311e74e8e) make more sense?

Issue #1902, opened 2020-01-28
Provide a standard way to consume Stylable's output


nitayneeman:
# 🙋 Feature Request

Hey,

I've already posted in your Slack channel, but I always prefer GitHub for making discussions :)

## Context
I’m working right now on guidelines for making mobile-friendly web applications using wix-style-react, when one of the aspects is SSR.

I know that you don’t support SSR at this point, but we’d love to hear your opinion anyway.

In wix-style-react, we use both CSS Modules and Stylable to style our components. This means that when an application is rendered using SSR, it makes sense to separate the CSS outputs out of the bundle (and that actually happens defaultly because of `separateCss`, which is great).

## Problem

The thing is, that the separated `app.css` doesn’t include the output of Stylable, whereas it arrives with the JS bundle itself (https://github.com/wix/yoshi/pull/329) and rendered at runtime.

We're looking for a standard way to consume Styleable's output specifically (mainly for SSR) - such we could suggest whole-heartedly. 😉

## Possible Solution 1

Exporting Stylable’s output into a separate file using `enhancedTpaStyle`.

That's what we currently instruct our consumers to do:
https://github.com/wix/wix-style-react/blob/master/docs/FAQ/README.md#components-do-not-have-styling-in-the-first-render-of-server-side-rendering-ssr

### Pros 
- Stylable output doesn't arrive with the JS bundle but in a separate file, beside `app.css`, so that both could be loaded explicitly when rendering with SSR

### Cons
- This isn't the main purpose of `enhancedTpaStyle` but just a side-effect
- Could be more side-effects that we don't actually want

## Possible Solution 2

Exporting Stylable’s output into a separate file using a dedicated configuration such as `stylableSeparateCss`.

Basically it means to expose the following configuration:
https://github.com/wix/yoshi/blob/9b437130ed28b3b26def37922eefec811b54fde2/packages/yoshi-common/src/webpack.config.ts#L229

### Pros 
- Same of the first solution (a separate file)
- We don't have the side-effects of `enhancedTpaStyle`

### Cons
-  Another configuration to maintain and document

## Possible Solution 3

Theoretically, Stylable’s output could be concatenated to `app.css` ahead of time, pretty much like happens for other CSS preprocessors - which means that would happen defaultly when using `separateCss`, instead of ignoring in case of Stylable (https://github.com/wix/yoshi/issues/309).

### Pros 
- All styles are placed in a single stylesheet per entry, that's easily consumable
[...]

PR #2010, opened 2020-02-25
Expose `stylableSeparateCss` configuration

nitayneeman:
### 🔦 Summary
This PR exposes a configuration called `stylableSeparateCss` that allows separating the Stylable output to a different bundle file.  This is a short-term (but simple) solution, as discussed in https://github.com/wix/yoshi/issues/1902.

### 🗺️ Test Plan
Added test that covers this configuration (verifies `app.stylable.bundle.css` is created correctly when `stylableSeparateCss` is true).


packages/yoshi-flow-legacy/config/webpack.config.js
```diff
@@ -23,7 +23,7 @@ const computedSeparateCss =
     ? inTeamCity || isProduction
     : projectConfig.separateCss;
 
-const stylableSeparateCss = projectConfig.enhancedTpaStyle;
+const separateStylableCss = projectConfig.enhancedTpaStyle || projectConfig.separateStylableCss;
```
nitayneeman: Two things:

1. Renamed this internal variable to `separateStylableCss`, just because it sounds better
2. This variable is conditioned by `projectConfig.enhanced TpaStyle` as well, to not degrade the current behavior

packages/yoshi-flow-legacy/test/webpack-config.spec.js
```diff
@@ -237,6 +237,23 @@ describe('Webpack basic configs', () => {
     });
   });
 
+  describe('Separate stylable css', () => {
```
nitayneeman: Let me know if there's a more appropriate spec file to place it

packages/yoshi-flow-legacy/config/webpack.config.js
```diff
@@ -44,7 +44,7 @@ const defaultOptions = {
   devServerUrl: projectConfig.servers.cdn.url,
   separateCss: computedSeparateCss,
   keepFunctionNames: projectConfig.keepFunctionNames,
-  stylableSeparateCss,
+  separateStylableCss,
```
nitayneeman: Currently it was added to `app` and `monorepo` flows, beside the `legacy`. Not sure if it's supposed to be added to the rest at this moment?

docs/api/configuration.md
```diff
@@ -120,6 +120,16 @@ Set to true to build with enhanced TPA style.
 
 - ![status experimental](https://img.shields.io/badge/status-experimental-ff69b4.svg)
 
+## separateStylableCss
+
+By default, the Stylable output will be bundled to the JS bundle (and rendered at runtime). You can bundle this output into a separate `app.stylable.bundle.css` file instead, by adding the following to your `package.json`:
```
ranyitz:
```suggestion
Output the stylable css into `app.stylable.bundle.css` file.

> By default, the Stylable CSS output will be bundled to the JS bundle (using the [`includeCSSInJS` option](https://github.com/wix/stylable/tree/master/packages/webpack-plugin#plugin-configuration-options)).
```

packages/yoshi-flow-legacy/test/webpack-config.spec.js
ranyitz: This is the old format of tests that checks the legacy flow.

I would prefer adding a test in the new [integration tests format](https://github.com/wix/yoshi/blob/master/CONTRIBUTING.md#integration-tests), you can create a stylable directory treating stylable [as a feature](https://github.com/wix/yoshi/tree/master/test/typescript/features). (Both testing the webpack config + yoshi config)

Please add a similar test in the new format instead of this one.

ranyitz:
@nitayneeman please run `yarn lint --fix` to fix prettier problems

packages/yoshi/types.d.ts
```diff
@@ -81,6 +81,13 @@ declare module '*.svg' {
   export default src;
 }
 
+declare module '*.st.css' {
```
nitayneeman: We need this to avoid type mismatches - otherwise the declaration is wildcarded by the following:
https://github.com/nitayneeman/yoshi/blob/feat/separate-stylable-css/packages/yoshi/types.d.ts#L91

nitayneeman:
@ranyitz - all comments were implemented, and hopefully the CI would pass this time.

PR #2016, opened 2020-02-25
Yoshi serve documentation

Gongreg:
Added documentation about yoshi serve. We are exposing it as node api so we need a separate document for it.


docs/api/node.md
```diff
@@ -0,0 +1,28 @@
...
+id: node
+title: Node API
+sidebar_label: Node API
+---
+
+## Yoshi Serve
+
+`yoshi-flow-app` exposes `serve` functionality through Node API.
+
+Serve runs your dev server with `NODE_ENV=PRODUCTION` and serves your statics in a local CDN.
```
ranyitz: I think that dev server is an ambiguate term, also it's ok to just say the conventions here, so it'll be clear.

```suggestion
Serve runs your `index-dev`/`dev/server` file with `NODE_ENV=PRODUCTION` and serves your `dist/statics` directory as a local CDN.
```

docs/api/node.md
```diff
@@ -0,0 +1,28 @@
...
+
+> Note: You need to build the statics before running serve. You can do that with `yoshi build`.
+
+`serve` method returns a Promise.
+
+Example of usage:
+
+```javascript
+require("yoshi-flow-app/scripts/build/serve")
```
ranyitz: Now when reading it, it feels like we should re-export this (so it will be clear that it's an public API)
```ts
import serve from "yoshi-flow-app/serve"
```
or maybe
```ts
import { serve } from "yoshi-flow-app"
```

Issue #2013, opened 2020-02-25
[RFC] Yoshi flow editor: Serve editor and settings apps


yavorsky:
Hey guys, we are working hard on new editor flow and for now I want to see some comments related to serving editor and settings apps.

From what I got, a lot of teams, who working on out of iframe apps has own solutions for it. We'll try to go through pros and cons of getting html files from cdn and compare it with serving settings and editor apps by app developers.

### Get html files from cdn
<img width="1596" alt="Screen Shot 2020-02-25 at 12 48 28 PM" src="https://user-images.githubusercontent.com/1521229/75240560-2fcb8280-57cd-11ea-83e1-f8f08850eb55.png">

The easiest one. Currently, our flow is not support any html files output. So people generating it using custom scripts from vm files. This is obviously bad:
 - users should spend some time on writing html generators.
 - spreads a lot of similar, but not reusable implementations
 - it's not documented

If we understand that this solution is working and optimal for everyone, we can add support to yoshi flow. I had a chance to play around it in feature branch: https://github.com/wix/yoshi/commit/2d89ba5987aacd6c981f5f5f2285c768139e7ea0

I can see some props of this idea in general:
- It's simple. The same way we are providing viewerScript bundle as a file, we can provide editor and settings html files.
- Easier to configure. We don't need to configure fryingpan, etc.
- Faster builds. Since we don't have any docker steps, the CI builds will take less time.
- (debatable) Getting html files from CDN is faster than receiving response from singe server configured via k8.

Cons will be:
- Less control over flow. Serving the app by ourselves gives us more control: load experiments before and serve stuff according to it, load translations.
- Flow is a bit different between local development and prod. To enable HMR, we need to have a local server. For prod, it will be just html.
- e2e tests. Could be really hard to make an e2e tests having just html files.

### Serve apps using node server infrastructure
<img width="1630" alt="Screen Shot 2020-02-25 at 12 47 47 PM" src="https://user-images.githubusercontent.com/1521229/75240537-28a47480-57cd-11ea-83f1-a5bf5982b7d4.png">

Obviously pros here will be cons from html solution. I will just add that experience should contain less boilerplate. We can hide everything related to platform and server from the flow. We can generate Dockerfile and other node related stuff during the build and hide it as much as possible.
User will just GA the app. Experiments, translations, etc will be passed to the Widget the same way for viewer and editor cases.

[...]

PR #2017, opened 2020-02-25
Output html files for editor apps (widget wrapper + settings)

yavorsky:
Screen Shot 2020-02-25 at 12 48 28 PM

### 🔦 Summary
This is a follow-up for https://github.com/wix/yoshi/issues/2013

Currently, I don't have strong opinion about generating html files instead of serving own app on prod. But this is an optional way to help teams already implemented OOI with old flow to migrate to new one. Moreover, we need to gather more feedback and understand if we should to invest more time to server implementation on current stage.

Basically here we just generating `statics/editor` and `statics/settings` directories with html and vm content. Then we can just take it's url from CDN (like a lot of teams currently doing) and include in dev center as an editor and settings apps.

Here, we also fixed css urls, since it was hardcoded to localhost and had no chance to work on prod.


packages/yoshi-flow-editor-runtime/src/WidgetWrapper.tsx
```diff
@@ -24,13 +25,13 @@ const PublicDataProvider: typeof React.Component =
...
 
-const WidgetWrapper = (UserComponent: typeof React.Component) => (
+const WidgetWrapper = (UserComponent: typeof React.Component, name: string) => (
   props: IHostProps & IFrameworkProps & IControllerContext,
 ) => {
   return (
     <div>
       <link
-        href="https://localhost:3200/todoViewerWidget.css"
+        href={`${window.__STATICS_BASE_URL__}${name}ViewerWidget.css`}
```
yavorsky: I'll make a better solution for css files, it's just a quick fix better than hardcoded localhist :/

packages/yoshi-flow-editor/src/server/vm.ts
```diff
@@ -3,9 +3,21 @@ import { Engine } from 'velocity';
...
+    >;
+  } = {},
+) =>
   new Engine({ template: url }).render({
     ...velocityData,
     ...velocityDataPrivate,
     ...data,
+    clientTopology: {
+      ...(velocityData.clientTopology ? velocityData.clientTopology : {}),
+      ...(data.clientTopology ? data.clientTopology : {}),
```
yavorsky: We need this to merge nested objects.

packages/yoshi-flow-editor/src/commands/htmlGenerator.ts
```diff
@@ -0,0 +1,52 @@
...
+import { STATICS_DIR } from 'yoshi-config/build/paths';
+import renderVM from '../server/vm';
+import { FlowEditorModel } from '../model';
+
+type vmType = 'editor' | 'settings';
+
+const templatesPath = path.resolve(__dirname, '../server/templates');
+
+const vmPaths: Record<vmType, string> = {
+  editor: path.join(templatesPath, './editorApp.vm'),
```
bordi: **TBD as part of https://github.com/wix/yoshi/issues/2013**
There will be more templates in few cases:
- separate Settings Panel for mobile
- separate Editor App for mobile

Hence, should be glob pattern be used here?

packages/yoshi-flow-editor/src/commands/htmlGenerator.ts
```diff
@@ -0,0 +1,52 @@
...
+    clientTopology: {
+      staticsBaseUrl: '../',
+    },
+  });
+  fs.copyFileSync(vmPath, path.join(destinationDir, `${widgetName}.vm`));
+  fs.writeFileSync(path.join(destinationDir, `${widgetName}.html`), rendered);
+};
+
+const getOutputFilename = (type: vmType) => {
+  return path.resolve(path.join(STATICS_DIR, type));
```
ranyitz: Easier to read as
```suggestion
  return path.join(process.cwd(), STATICS_DIR, type);
```

packages/yoshi-flow-editor/src/commands/htmlGenerator.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const generateHTML = (type: vmType, widgetName: string) => {
+  generateHTMLFromVM(getOutputFilename(type), vmPaths[type], {
```
ranyitz: `getOutputFilename` could be better named `getOutputDirectory`

But maybe it could just be an implementation detail of `generateHTMLFromVM`

packages/yoshi-flow-editor/src/commands/htmlGenerator.ts
```diff
@@ -0,0 +1,52 @@
+import path from 'path';
+import fs from 'fs-extra';
+import { STATICS_DIR } from 'yoshi-config/build/paths';
+import renderVM from '../server/vm';
+import { FlowEditorModel } from '../model';
+
+type vmType = 'editor' | 'settings';
+
+const templatesPath = path.resolve(__dirname, '../server/templates');
```
ranyitz: `path.resolve(arg)` is `path.join(process.cwd(), arg)`. We don't need the `process.cwd()` resolution here because `__dirname` is absolute.

While it's not critical, I would use `path.join` here.

sdoomz:
I like the idea of creating index.html during the build phase and putting it to the statics folder. 
In this case, you don't even need to think about the server and all your client app is loaded from a CDN. 

Pros:   
1. Content is delivered a bit faster. Because of DNS resolution and 0 server processing time. 
2. Reliability. `index.html` is quite stable and doesn't require additional monitoring tools or health checks. 

On the other hand, global variables as `__STATICS_BASE_URL__` are still injected and it's possible to output `index.html` file either with `app.js` or `app.min.js` based on your `NODE_ENV`.

In Search App we used a similar, but less flexible, approach - [just html file](https://github.com/wix-private/site-search/blob/master/packages/search-results-widget/src/editorWidget.html) and it took us some time to implement it. 

Summarizing everything, it would be really nice to have this behaviour implemented in the new `editor-flow` out of the box. 

Thanks Artem.

packages/yoshi-flow-editor/src/commands/htmlGenerator.ts
```diff
@@ -0,0 +1,52 @@
...
+    widgetName,
+  });
+};
+
+const syncHTMLDirectory = (type: vmType) => {
+  fs.mkdirpSync(path.join(STATICS_DIR, type));
+};
+
+export const generateEditorHTMLFiles = (model: FlowEditorModel) => {
+  syncHTMLDirectory('editor');
```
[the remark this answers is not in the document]
yavorsky: It isn't :(

packages/yoshi-flow-editor-runtime/src/WidgetWrapper.tsx
yavorsky: Not sure I got it. You mean `__webpack_public_path__` ?

yavorsky:
I've added a basic README file, which explain the basic use case.
I'm going to add more docs, with use cases for dev-center registration and point current doc here.

packages/yoshi-flow-editor-runtime/src/WidgetWrapper.tsx
ranyitz: I mean that we know what would be the public path, and I think that it's equivalent to the statics base url, so we can inject it in build time instead of taking it from the window.

In any case, if it's working right now we can continue using this methodology and update it in a later PR

PR #2018, opened 2020-02-25
Yoshi Serve in yoshi-common

Gongreg:
### 🔦 Summary

Moved out `yoshi serve` to `yoshi-common`.

Documentation: #2016

### 🗺️ Test Plan

Added test to check that we quit when user tries to run it in monorepo.
Other tests were already in place.


test/javascript/features/serve/monorepo/monorepo.test.ts
```diff
@@ -0,0 +1,17 @@
+import Scripts from '../../../../scripts';
+
+const scripts = Scripts.setupProjectFromTemplate({
+  templateDir: __dirname,
+  projectType: 'javascript',
+});
+
+describe('Yoshi serve monorepo', () => {
+  it('should notify is yoshi build was not run', async () => {
```
ranyitz:
```suggestion
  it('should throw an error that monorepo flow is not supported', async () => {
```

test/scripts.ts
```diff
@@ -247,8 +247,7 @@ export default class Scripts {
     resolve: TestCallback,
     reject: (reason: string) => void = () => {},
   ) {
-    const serve = require('../packages/yoshi-flow-app/build/scripts/serve')
-      .default;
+    const serve = require('../packages/yoshi-common/serve');
```
ranyitz: Is there a reason not to `import` it in the root?

test/typescript/features/serve/monorepo/monorepo.test.ts
```diff
@@ -0,0 +1,17 @@
+import Scripts from '../../../../scripts';
+
+const scripts = Scripts.setupProjectFromTemplate({
+  templateDir: __dirname,
+  projectType: 'javascript',
+});
+
+describe('Yoshi serve monorepo', () => {
+  it('should notify is yoshi build was not run', async () => {
```
ranyitz:
```suggestion
  it('should throw an error that monorepo flow is not supported', async () => {
```

PR #2019, opened 2020-02-25
Yoshi Server HMR

yanivefraim:
The first step of adding HMR support for Yoshi Server.

TODO:
Add support for adding/deleting files (API/route)


packages/yoshi-server/src/routes/api.ts
```diff
@@ -12,24 +13,42 @@ import { route } from '..';
...
-  return {
-    ...acc,
-    [filename]: chunk,
+  return functions;
+}
+
+if (process.env.NODE_ENV === 'development') {
+  const socket = new SockJS(
+    `http://localhost:${process.env.HMR_PORT}/_yoshi_server_hmr_`,
+  );
```
ronami: I think we can re-use the code that connects to the websocket since it's used in a few places already.

test/javascript/features/yoshi-server/hmr/hmr.test.ts
```diff
@@ -0,0 +1,71 @@
...
+const scripts = Scripts.setupProjectFromTemplate({
+  templateDir: __dirname,
+  projectType: 'yoshi-server-javascript',
+});
+
+describe.each(['dev'] as const)('yoshi-server route hmr [%s]', mode => {
+  const routeFilePath = path.join(scripts.testDirectory, 'src/routes/app.js');
+  let originalContent: string;
+  beforeEach(() => {
+    originalContent = fs.readFileSync(routeFilePath, 'utf-8');
```
ronami: I don't think we need a `beforeEach()` here

yanivefraim: I wanted to be able to get back to the previous state because we can have multiple tests. Another option is to have try/catch/finally. WDYT?

test/javascript/features/yoshi-server/hmr/hmr.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    fs.writeFileSync(apiFilePath, originalContent);
+  });
+  it('api function', async () => {
+    await scripts[mode](async () => {
+      await page.goto(`${scripts.serverUrl}/app`);
+      const title = await page.$eval('h1', elm => elm.innerHTML);
+      expect(title).toBe('hello from yoshi server Yaniv');
+
+      const editedContent = originalContent.replace(
+        'greeting: `${name}`,', // eslint-disable-line
```
ronami: Why ESLint complains? Can we disable only a specific rule?

test/javascript/features/yoshi-server/hmr/hmr.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    'src/api/greeting.api.js',
+  );
+  let originalContent: string;
+  beforeEach(() => {
+    originalContent = fs.readFileSync(apiFilePath, 'utf-8');
+  });
+  afterEach(() => {
+    fs.writeFileSync(apiFilePath, originalContent);
+  });
+  it('api function', async () => {
```
ronami: Since both route and API function have the same setup, can they both be under the same `describe` block?

packages/yoshi-server/src/routes/api.ts
yanivefraim: done, thanks

test/javascript/features/yoshi-server/hmr/hmr.test.ts
yanivefraim: Fixed, thanks

test/javascript/features/yoshi-server/hmr/hmr.test.ts
yanivefraim: fixed

PR #2023, opened 2020-02-26
add stylable dependencies (Support for stylable 3)

ranyitz:
### 🔦 Summary
This PR should effectively add support for `stylable@3` as an opt in while keeping `stylable@1` as the default.

Yoshi will bring `stylable@1` as it has done before, but the user will be able to upgrade its stylable version to version 3 by importing `yoshi-stylable-dependencies`.

___________

`yoshi-stylable-dependencies` is a bundle of dependencies that includes a webpack plugin, node require hook and jest transformer. All of those dependencies are on version 3.

While the API for the users has changed from version 1 to 3 of stylable, the API to yoshi left similar, with some minor changes.

Yoshi will check for the existence of `yoshi-stylable-dependencies` package, and in case it exist it will use its stylable dependencies.

_______________

Most of yoshi's users aren't using stylable directly, but instead the different wix UI libraries, like [wix-style-react](https://github.com/wix/wix-style-react) are using it. Using the new method, they'll be able to upgrade their code to stylable 3 and add `yoshi-stylable-dependencies` to their `package.json` dependnecies. By doing that, yoshi will use the new version, which means that it leaves the control on the stylable version to the code that actually uses it.

We're assuming here 2 things:

1. A certain user uses a single component library.
2. Our users do not use stylable directly.

This change should ease the migration by letting the component libraries to update their version without upgrading a major version to Yoshi.


barak007:
It will be nice to have a tool that checks for duplicate `yoshi-stylable-dependencies` or library that requires different stylable versions e.g `wix-ui-core` .

shlomitc:
@ranyitz - nice work. Is this fix merged also the yoshi@3?

ranyitz:
It's only implemented for Yoshi 4

shlomitc:
I don’t have the numbers, but I guess that there are still many yoshi users using v3 which will need to upgrade both the relevant component library and their yoshi version.
If this number is small - maybe it’s ok to keep this feature only in v4. If not - what is the effort to port this to v3?

ranyitz:
@shlomitc the last version of yoshi 3 was released [more than a year ago.](https://github.com/wix/yoshi/blob/master/CHANGELOG.md#3305-2019-02-21) We're not adding features to old versions, only critical bug fixes. This is not a critical bug fix, but a feature that enables you to upgrade the version of stylable.

In reality the migration from 3 to 4 should be very easy, in case it takes more than 15 minutes, we're here to help.

I don't have exact numbers today, but those are the numbers from November 18 2019:
```
wix-style-react@7 yoshi versions:
4 => 60
3 => 8
2 => 1

wix-style-react@6 yoshi versions:
4 => 109
3 => 43
2 => 1

Aggregated:
169 - yoshi@4
51 - yoshi@3
2 - yoshi@2
```

I assume that today, after almost half a year the number of yoshi 3 users should be much smaller.

shlomitc:
cool, thanks a lot

Issue #1210, opened 2019-04-04
Transpile all files in the project


ranyitz:
# 🙋 Feature Request

### 😯 Current Behavior

When looking at a fullstack project, there are multiple files that are not being transpiled. It means that for example, you can't write esm there (`import` & `export`), you can write them in TypeScript etc...

### 🤔 Expected Behavior

We would be able to write in all files in the same syntax.

### 💁 Possible Solution 

- [ ] Transpile `jest-yoshi.config.js` file, the transpilation of the `global-setup` file in jest would enable transpilation to the yoshi config file, currently done via https://github.com/wix/yoshi/pull/1176. We should change this file in TypeScript projects to `.ts` and change from `require` to `import` in babel projects.

- [ ] Transpile `index.js` file - We can bundle the index.js file (instead of the `server.js` file), we just need that the `node platfrom` will accept a function instead a path to a function, we would be able to `require/import` it and it will be a part of the bundle. (app-flow) (cc @ronami)

- [ ] Transpile `index-dev.js` file - We need to run some `ts-node/babel-register` before loading it and then it will be transpiled. We need to keep note that the `dist/server.js` is not transpiled, we can actually ignore the whole `dist` directory when configuring `babel`/``TypeScript

### 🔦 Context

When one file is being transpiled and other file doesn't it creates a weird behavior that the developer needs to think about. It also creates inconsistencies in the codebase which i find annoying.

ranyitz:
Related https://github.com/wix/yoshi/issues/888

PR #2029, opened 2020-03-01
add support for typescript/babel transpilation for server process

ranyitz:
partially `fixes` #1210

### 🔦 Summary
This PR adds support for TypeScript/babel transpilation for server process file (e.g. `index-dev.ts`/`dev/server.ts`). It's now possible to write those files with the same syntax as the rest of the app.

* [x] TypeScript runs without failing on TypeChecking (`fast`) because it will happen in the IDE and that's not critical to fail the development build on type errors during `npm start` (which doesn't run in the CI)

* [x] Modify templates and remove the deprecated `--server` so they'll take `dev/server.ts` instead.

* [x] Add link to fixtures in contributing

### 🗺️ Test Plan
I've noticed we don't have a specific test for `app-server` but instead, all other integration tests are testing that it works. The easiest thing was to change `index-dev.js` to `index-dev.ts` in templates tests to verify that it works.

Added a test for the TypeScript case, where we test that the server doesn't fail on type errors.

(I've also tested it manually, and in client projects in templates)


packages/yoshi-common/src/server-process-worker.js
```diff
@@ -0,0 +1,16 @@
+// We use this file in order to have a `fork` with `ipc`
+// and also have `ts-node` register for TypeScript support
+// for the user's server file
+
+const path = require('path');
+
+const serverFilePath = process.env.__SERVER_FILE__;
+
+if (path.extname(serverFilePath) === '.ts') {
```
ronami: I think we should check for the existence of a `tsconfig.json` (`isTypeScriptProject()`, can be passed as an env var from outside), since some projects may use `allowJs` and have only some of their setup code in TypeScript.

packages/yoshi-common/src/server-process-worker.js
```diff
@@ -0,0 +1,16 @@
...
+  require('ts-node').register({
+    fast: true,
+  });
+}
+
+// eslint-disable-next-line
```
ronami: Why is that needed? Can we disable only a specific rule instead of all of them?

packages/yoshi-common/src/server-process-worker.js
ranyitz: Cool, so we have `ts-node` configured even if the initial file is a `js` file.

Good idea! I'll fix that

packages/yoshi-common/src/server-process-worker.js
ranyitz: It's because of the `require` call, i'll add the specific rule

ranyitz: Actually it's not needed, my bad, removing

packages/yoshi-common/src/server-process-worker.js
ronami: Thinking about it again, maybe run `@babel/register`/`ts-node`, depending on the project type? We still support Babel and only having this feature for ts project will be annoying for existing Babel projects.

packages/yoshi-common/src/server-process-worker.js
ronami: I don't think a dynamic `require` call should show any linting errors since it cannot be done with an import. It's interesting to know what rule is failing.

packages/yoshi-common/src/server-process-worker.js
ranyitz: I've modified the code to handle both cases, I can easily add a test as well

CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ Test each feature on a dedicated installed project.
 
 How it works:
 
-1. We copy base fixture template to `.tmp` folder (basic files like `package.json` etc.)
+1. We copy [base fixture template](./test/fixtures/typescript/base-template/package.json) to `.tmp` folder (file which are neede in most tests, like `package.json`)
```
ranyitz:
```suggestion
1. We copy [base fixture template](./test/fixtures/typescript/base-template/package.json) to `.tmp` folder (file which are needed in most tests, like `package.json`)
```

packages/yoshi-common/src/server-process-worker-with-transpilation.js
```diff
@@ -0,0 +1,8 @@
+// We use this file in order to have a `fork` with `ipc`
+// We setup either `babel/register` or `ts-node`
+// for the user's server file to have transpilation enabled
+const userServerFilePath = process.env.__SERVER_FILE_PATH__;
+
+require('../build/require-hooks').setupRequireHooks();
```
ronami: I'm not sure if we need this since this file (`server-process-worker-with-transpilation.js`) will be in `/build` anyway. Can it be a TypeScript file with an import? That way, if we use it wrong or import something unknown we'll get a compilation error.

ranyitz: modified

ranyitz:
@ronami Thanks for the great review! All code review comments have been fixed. Let me know if anything is missed.

packages/yoshi-common/src/server-process.ts
```diff
@@ -57,7 +57,13 @@ export class ServerProcess {
       cwd: this.cwd,
     });
 
-    this.child = child_process.fork(this.serverFilePath, [], {
+    const userServerFilePath = path.join(process.cwd(), this.serverFilePath);
+    const serverProcessWorker = path.join(
+      __dirname,
+      'server-process-worker-with-transpilation.js',
+    );
```
ronami: Super tiny (let's fix in a future commit) but can just be `require.resolve()`

test/typescript/features/app-server/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "lib": ["dom", "esnext", "webworker"],
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "strict": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true
+  },
+  "files": [
+    "/Users/rany/Projects/yoshi/packages/yoshi/types.d.ts",
+    "/Users/rany/Projects/yoshi/node_modules/jest-yoshi-preset/types.d.ts"
```
ronami: Oops `/Users/rany/Projects/yoshi/packages/yoshi/types.d.ts`

ronami: Use `{%rootPath%}`

ranyitz: Opps 😨

ranyitz:
@ronami Good points, new review comments fixed as well! Again, thanks for the review ❤️

PR #2035, opened 2020-03-03
Remove website lockfile

ranyitz:
### 🔦 Summary
This PR removes the lockfile from the website installation.

When the lockfile of the website is updated with the private npm registry, travis can't reach it and fail. While we were able to solve it for [the case of release creation](https://github.com/wix/yoshi/pull/2014) renovate which attempts to do lockfile maintenance currently don't support different registries for different lock files, or even to update a single lockfile and don't update the other. (I checked the source code...) 

Lockfile can help with determinism, but unfortunately in our case (as a second lock file) creates more problems.


scripts/createVersion.js
```diff
@@ -63,16 +63,6 @@ Promise.resolve()
       console.log();
     }
 
-    console.log();
```
yanivefraim: This part is supposed to prevent errors during new version creation, in case someone did `yarn install` on Yoshi's root, without installing the website. Are you sure that we want to remove it?

scripts/createVersion.js
```diff
@@ -63,16 +63,6 @@ Promise.resolve()
...
-    console.log('installing website...');
-    console.log();
-
-    execa.sync('yarn install --pure-lockfile', {
```
yanivefraim: This part is supposed to prevent errors during new version creation, in case someone did `yarn install` on Yoshi's root, without installing the website. Are you sure that we want to remove it?

ranyitz:
@yanivefraim Yes, now that there is no lock file, this `install` would be redundant for `99%` or more of the cases. I think that if we want to handle a case where someone releasing for the first time we can do it with an error message and introduce much less problems for the day to day workflow

ranyitz:
We should probably create this custom error handling for this, but that's an edge case, because there aren't many people that create versions in yoshi

ranyitz:
Also, while the error is not clear. It's still not user facing, and seeing missing dependency is not so far from understanding that install is needed.

PR #2051, opened 2020-03-05
support for TypeScript/Babel transpilation for server process

ranyitz:
This is reopening of the origin PR - https://github.com/wix/yoshi/pull/2029 After it was reverted on - https://github.com/wix/yoshi/pull/2047 due to causing flakiness in tests (which may lead to start time degradation for end-users)


ranyitz:
Some debugging information:

Those are the flaky tests:
image

ranyitz:
**TL;DR - We won't include the index-dev.js (JS) file using babel in this PR because it may be breaking for users**

Transpilation `index-dev.ts` is a non breaking change, because it's not possible to have such file using currently working API. But if we'll transpile `index-dev.js` using `@babel/register` that would be a possible breaking change.

Also, one of the `hmr` tests is failing due to this change (only on `js`) which means we'll need to debug it once we'll want to enable `index-dev.js` transpilation, probably in the next major version.

PR #2054, opened 2020-03-06
extract `prettier-config-yoshi` package

bodia-uz:
### 🔦 Summary

To share one prettier config between different projects, we need to extract yoshi prettier config to a separate package (`prettier-config-yoshi`).

It will allow us to use it in existing `eslint-config-yoshi-base`, `tslint-config-yoshi-base` projects, and other projects that want to be aligned with yoshi.

### 🗺️ Test Plan

its just config file. Not sure if we need to test it in some way.


packages/prettier-config-yoshi/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "prettier-config-yoshi",
+  "version": "4.4.3",
+  "description": "prettier configuration for projects using yoshi",
+  "main": "index.js",
+  "keywords": [
+    "prettier",
+    "yoshi"
+  ],
+  "author": "Ronen Amiel & Ran Yitzhaki",
```
zemlanin: Bohdan, don't be shy 🙃

bodia-uz:
```suggestion
  "author": "Ronen Amiel & Ran Yitzhaki & Bohdan Lyzanets",
```
;)

packages/prettier-config-yoshi/README.md
```diff
@@ -0,0 +1,59 @@
+# prettier configuration for yoshi based projects
+
+`prettier.config.js` or `.prettierrc.js`:
+
+```
+const prettierConfig = require('prettier-config-yoshi');
+
+module.exports = {
+  ...prettierConfig,
+  semi: false
```
yavorsky: I think the idea of having a common place for prettier config is to just reuse it, not extend. I mean, prettier configuration is not to big just less than 10 options. I won't encourage people to extend it in docs.
We are extending eslint-config using *-base* approach just b/c we want to add new rules (specific to project environment), not override existing ones.

packages/prettier-config-yoshi/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+eslint config:
+
+```js
+const prettierConfig = require("prettier-config-yoshi");
+
+module.exports = {
+  plugins: ["prettier"],
+  rules: {
+    "prettier/prettier": ["warn", prettierConfig]
```
yavorsky: I would also add to docs that if you already have `.prettierrc` or other prettier configuration in the project, you don't need to pass prettierConfig to eslint rule.

packages/prettier-config-yoshi/index.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  singleQuote: true,
+  trailingComma: 'all',
```
yavorsky: Not sure about this, but maybe adding other options that are set by default (like printWidth: 80, bracketSpacing: true) will help to highlight other fields in editor during import.

packages/prettier-config-yoshi/README.md
bodia-uz: I copied a few examples from prettier docs and adapt the, but you a right, better to not provoke people to do it=)

packages/prettier-config-yoshi/index.js
bodia-uz: Maybe. @ranyitz what do you think?

packages/prettier-config-yoshi/README.md
bodia-uz: Updated this section. Now it's less detailed and recommends the user to use `eslint/tslint-config-yoshi-base`

bodia-uz:
@ranyitz the main purpose of this PR is to allow Wix projects that are not under Yoshi, to be aligned with Yoshi prettier config without thinking what is inside this config. We planned to use it in our project.
For now, we just added the same config in our repo. It's ok for us and we can live with it, or we can import it from `prettier-config-yoshi` if you find it reasonable to add it.

ranyitz:
@bodia-uz I understand. 

> allow Wix projects that are not under Yoshi, to be aligned with Yoshi prettier config without thinking about what is inside this config

If you'll be aligned with `eslint-config-yoshi` you'll be aligned with the prettier config as well.

____________

What's the use-case for a non Yoshi project in Wix?

zemlanin:
> What's the use-case for a non Yoshi project in Wix?

Projects not _(yet?)_ on yoshi, like https://github.com/wix-private/santa-editor

bodia-uz:
@ranyitz as Anton said we want to use in our project `santa-editor`, which is not under yoshi now, and not ready yet, to use `eslint-config-yoshi`.

ranyitz:
@bodia-uz @zemlanin If that's the case, I would go with duplicating
```
singleQuote: true,	
trailingComma: 'all',
```

Paying the price of the duplication in this case is lower than the price of maintaining another package.

zemlanin:
I'm not much in favor of having a package with two-keys-object, but duplicating doesn't seem great to me also…

Would using `require('packages/eslint-config-yoshi-base').rules['prettier/prettier']` in `santa-editor` be _so_ bad?..

ranyitz:
@zemlanin that's totally ok
```
require('eslint-config-yoshi-base').rules['prettier/prettier']
```

In my opinion (which I understand can be controversial) repeating yourself in some cases is totally ok, and just a matter of tradeoff. It depends on how many times the code is repeated, how much code is repeated, how much this code is likely to be changed in the future, how much effort it takes to not repeat it and more.

In this case, both solutions seems fine to me, but managing another package for prettier config is not something I want to maintain.

bodia-uz:
Cool.
We will go with duplication for now. Then (when ready) migrate to `eslint --fix` using `eslint-config-yoshi` config.

Thanks for the discussion.

PR #2056, opened 2020-03-08
thunderbolt - Add `editor-elements-registry` to bundling check

NitayRabi:
### 🔦 Summary
Allow bundling thunderbolt-elements correctly through re-export (see https://github.com/wix-private/editor-elements/pull/221)


packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -148,7 +149,7 @@ export function createServerWebpackConfig(
 ): webpack.Configuration {
   const defaultOptions = createDefaultOptions(pkg);
 
-  const customThunderboltElements = pkg.name === 'thunderbolt-elements';
+  const customThunderboltElements = isThunderboltElementModule(pkg);
```
giladsegal: looks like this flag is redundant, you can invoke `isThunderboltElementModule` instead

.gitignore
```diff
@@ -10,4 +10,5 @@ build
 tsconfig.tsbuildinfo
 *_backup
 .tmp
-.history
\ No newline at end of file
+.history
+yarn-error.log
```
ranyitz:
```suggestion
```

PR #2057, opened 2020-03-08
Fix yoshi serve: app-server use `NODE_ENV="development"`

ranyitz:
### 🔦 Summary
Using serve would ideally run the app-server with `NODE_ENV="production"`.

In reality the node_platform configures things a bit differently for production

image

So until we'll dive in an see how we can fix it, we'll start with using `NODE_ENV="development"` for the server process which effectively makes this pass


packages/yoshi-common/src/server-process.ts
```diff
@@ -37,7 +37,7 @@ export class ServerProcess {
     serverFilePath,
     appName,
     env = {
-      NODE_ENV: 'production',
+      NODE_ENV: 'development',
     },
```
ronami: Not related to the PR but I think it's less confusing, by default, to set the value of `NODE_ENV` to development if it's not defined instead of having a default env (that only works if it wasn't passed at all).

This way, we get the default `NODE_ENV` in both of the following cases:

```js
new ServerProcess()
```

And also:

```js
new ServerProcess({ env: { HELLO: 'world' } })
```

ranyitz: #2061

PR #2058, opened 2020-03-08
Provide 404 and 500 errors on Yoshi Server Routes

### LINKED ISSUE #2049: Yoshi Server - use production 404 / 500 pages
In production, we want to have the default `node_platform` 404 / 500 error pages (we currently have a non-production 404 / 500 pages)

yanivefraim:
Fixes #2049 

We want to have proper error pages on production (using `node-platform`'s error pages). 

- In case of a server error - we will call `next` with a custom error. This way `node-platform` will catch the error, log it to `new-relic`, and show a proper error page.

- In case of a 404 - if Yoshi Server did not find a match, it will pass the request using `next()`. If the user did not supply another endpoint (after his custom server), `node-platform` will catch the request and log it to `new-relic` + display a proper 404 page.


test/fixtures/yoshi-server-typescript/base-template/tsconfig.json
```diff
@@ -10,7 +10,9 @@
   },
   "files": [
     "{%rootPath%}/packages/yoshi/types.d.ts",
-    "{%rootPath%}/node_modules/jest-yoshi-preset/types.d.ts"
+    "{%rootPath%}/node_modules/jest-yoshi-preset/types.d.ts",
+    // a workaround until we publish the types
```
yanivefraim: Do we want to publish the types in this PR?

packages/yoshi-server/src/server.ts
```diff
@@ -19,6 +21,13 @@ export type Route = {
   ) => void;
 };
 
+class InternalServerError extends wixBusinessError(1011, HttpStatus.INTERNAL) {
+  // default error code: -100, default http status: 500.
+  constructor(msg?: string, cause?: Error | string) {
+    super(msg, cause);
+  }
+}
+
```
ronami: Super small but I think this can go into an `httpErrors.ts` file

packages/yoshi-server/src/server.ts
```diff
@@ -56,8 +65,11 @@ export default class Server {
         }
       }
     } catch (error) {
-      if (process.env.NODE_ENV === 'production') {
-        return send(res, 500, 'internal server error');
+      if (
+        process.env.NODE_ENV === 'production' ||
+        process.env.IS_INTEGRATION_TEST_PROD === 'true'
+      ) {
+        next(new InternalServerError('internal server error', error));
```
ronami: should probably `return`

packages/yoshi-server/src/server.ts
```diff
@@ -66,6 +78,11 @@ export default class Server {
       return send(res, 500, html);
     }
 
+    if (process.env.NODE_ENV === 'production') {
+      // If Yoshi Server did not find anything, pass the request on
+      next();
```
ronami: should probably `return` too

packages/yoshi-server/src/server.ts
yanivefraim: nice catch, thanks. Done.

packages/yoshi-server/src/server.ts
yanivefraim: nice catch, thanks. Done.

packages/yoshi-server/src/server.ts
yanivefraim: done

PR #2060, opened 2020-03-08
Internal: move some integration tests from `.js` to `.ts`

yanivefraim:
Found some integration tests written in `.js`...


test/fixtures/yoshi-server-typescript/base-template/tsconfig.json
```diff
@@ -10,7 +10,9 @@
   },
   "files": [
     "{%rootPath%}/packages/yoshi/types.d.ts",
-    "{%rootPath%}/node_modules/jest-yoshi-preset/types.d.ts"
+    "{%rootPath%}/node_modules/jest-yoshi-preset/types.d.ts",
+    //Will the use also need this one???
```
yanivefraim: I guess that we will want a better solution here. @ronami , WDYT?

PR #2069, opened 2020-03-09
Deprecate `yoshi release` command

ranyitz:
### 🔦 Summary
This PR is deprecating the command `yoshi release`
This will be a part of the domain of `fed-ci-scripts`

See https://github.com/wix-private/wix-fed-scripts/pull/37

It's important to note that this PR only relevant to publishing `npm` modules. For some reason `yoshi release` was configured in all templates. This PR also removes them.

This is the message that will be seen when running `yoshi release` after this PR will be merged.

image


ranyitz:
@netanelgilad Thanks! Fixed ⚓️

PR #2075, opened 2020-03-09
Align npm scripts in templates

### LINKED ISSUE #1689: Align generators npm scripts
# NPM Scripts in Generators

Some generators specify unconventional (?) npm scripts, for example:

https://github.com/wix/yoshi/blob/a77be2e2823e659f7325f91cde1b09f2dd5c3e72/packages/create-yoshi-app/templates/business-manager-module/typescript/%7B%25packagejson%25%7D#L10-L16

Since the generators are basically setting the standard for newly created projects at Wix, we should provide a better starting point. 
The current setup raises eyebrows.

* I don't think we need to have a `release` phase in most templates (probably only in `library`).
* We need to have a `build` script, which will be referenced in `pretest`, or not.

ranyitz:
# 🔦 Summary
This PR changes the `pretest` build script into just `build`. (and Fixes #1689 )

Today, many people try to run `npm run build` or `yarn build` and waste time understanding that they actually need to run `npm run pretest`.

We used to run the build before we run the test, today, the CI [supports running the build script if it exists](https://github.com/wix-private/wix-fed-scripts/blob/master/src/npm-ci-scripts/build.ts#L63). When running the test locally we advice to not build, but instead `start` so we'll have an incremental build, see the [DEV mode](https://wix.github.io/yoshi/docs/guides/jest-setup#dev-mode) with Yoshi.

________

I also added `start` command for libraries, because even libraries would like to have incremental build.


packages/create-yoshi-app/templates/server/typescript/{%packagejson%}
```diff
@@ -10,8 +10,8 @@
   },
   "scripts": {
     "start": "yoshi start",
-    "pretest": "yoshi build",
-    "test": "yoshi test --mocha",
```
yanivefraim: Tests in the server template are still mocha. Was this changed by mistake?

ranyitz: Yes, totally a mistake :)

packages/create-yoshi-app/templates/server/typescript/{%packagejson%}
```diff
@@ -10,8 +10,8 @@
...
+    "build": "yoshi build",
+    "test": "yoshi test",
```
ranyitz:
```suggestion
    "test": "yoshi test --mocha",
```

PR #2078, opened 2020-03-09
Yoshi Server: Support nested default routes

### LINKED ISSUE #2071: Yoshi Server: default route (`index.js`) does not work when not in root
Use case:

```
- routes
-- app
---- index.js
```

There is a bug here:
https://github.com/wix/yoshi/blob/master/packages/yoshi-server/src/server.ts#L96
`routePath` will be `'app/index'`, and the condition for `routePath === '/index'` will be falsy.

yanivefraim:
Fixes #2071 - we want to support nested default routes.


packages/yoshi-server/src/utils.ts
```diff
@@ -36,3 +36,14 @@ export function pathMatch(route: string, pathname: string | undefined) {
     };
   }, {});
 }
+export function buildRoute(routesBuildDir: string, absolutePath: string) {
+  const relativePath = `/${relativeFilePath(routesBuildDir, absolutePath)}`;
+  // Change `/users/[userid]` to `/users/:userid`
+  const routePath = relativePath.replace(/\[(\w+)\]/g, ':$1');
+  // TODO: refactor
+  return routePath === '/index'
```
yanivefraim: This might need a refactor. I just wrote it quick and dirty so it will work...

packages/yoshi-server/src/server.ts
```diff
@@ -89,11 +89,9 @@ export default class Server {
 
     return serverChunks.map(absolutePath => {
       const chunk = importFresh(absolutePath) as RouteFunction<any>;
-      const relativePath = `/${relativeFilePath(routesBuildDir, absolutePath)}`;
-      // Change `/users/[userid]` to `/users/:userid`
-      const routePath = relativePath.replace(/\[(\w+)\]/g, ':$1');
+      const route = buildRoute(routesBuildDir, absolutePath);
```
ronami: since `routesBuildDir` doesn't change, I think we can move it over to `buildRoute` instead of passing the same value as a parameter.

packages/yoshi-server/src/utils.ts
```diff
@@ -36,3 +36,14 @@ export function pathMatch(route: string, pathname: string | undefined) {
...
+export function buildRoute(routesBuildDir: string, absolutePath: string) {
+  const relativePath = `/${relativeFilePath(routesBuildDir, absolutePath)}`;
+  // Change `/users/[userid]` to `/users/:userid`
+  const routePath = relativePath.replace(/\[(\w+)\]/g, ':$1');
+  // '/index' => '/'
+  // '/app/index' => '/app'
+  // else return routePath as is
+  return path.basename(routePath) === 'index'
+    ? routePath.replace('/index', '') || '/'
+    : routePath;
```
ronami: I think this can be easier to understand if we split it to multiple `if ()` statements with a comment on top of each. I think splitting the `||` part will also make it clearer on what edge case each check handles.

PR #2080, opened 2020-03-10
add public path to prod source maps; use plugin directly

roymiloh:
Up until now, at the bottom of each bundle file, there was a source map url that pointed to the source map file, relative to the path of the bundle. Say the bundle url was `x.com/bundle.js`, the source map url that was written is `bundle.js.map`.

However, there are some cases where we load the bundle directly to the html (inline it), so the url is not relative anymore. In order to make things work, we have to add the absolute url so the browser will identify the source maps correctly.

I also got rid of [webpack's devtool](https://webpack.js.org/configuration/devtool/#devtool) option as it doesn't allow us to add a custom public path. I basically did what `devtool` [does under the hood](https://github.com/webpack/webpack/blob/a291f8b55f023142c1feec6fb5bda92c869f8465/lib/WebpackOptionsApply.js#L242-L271).


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -356,9 +357,6 @@ export function createBaseWebpackConfig({
             chunkFilename: `${SERVER_CHUNKS_DIR}/[name].js`,
             libraryTarget: 'umd',
             globalObject: "(typeof self !== 'undefined' ? self : this)",
-            // Point sourcemap entries to original disk location (format as URL on Windows)
-            devtoolModuleFilenameTemplate: info =>
-              path.resolve(info.absoluteResourcePath).replace(/\\/g, '/'),
```
ronami: Can't we keep this?

PR #2091, opened 2020-03-11
[yoshi template "library"]: use `maven-v4.0.0.xsd` in pom.xml

sidoruk-sv:
### 🔦 Summary
This tiny improvement eliminates a redirect, cause http://maven.apache.org/maven-v4_0_0.xsd will redirect to http://maven.apache.org/xsd/maven-4.0.0.xsd


sidoruk-sv:
> I'm not sure what the effect of it, but it looks correct, Thanks! 👍

Just consistency with other templates and minimising redirects on this resource (if it requested anywhere)

PR #2106, opened 2020-03-16
Yoshi server: init server

yanivefraim:
We want to be able a one-time setup for the server, which will run only when the server starts (today, we have to add this for each request).

The convention is `src/init-server.ts/js`, which exposes a function:

```js

export default async context => {
	// return init data;
};
```

Init data will be available on `this`:

```js
import { method } from 'yoshi-server';

export const greet = method(function(name) {
  return {
    greeting: `hello ${name} ${this.initData}`,
    name,
  };
});

```


packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -101,10 +101,18 @@ function createServerEntries(context: string, cwd: string = process.cwd()) {
       })
     : [];
 
+  const initFunction = fs.pathExistsSync(path.join(cwd, SRC_DIR))
```
yanivefraim: todo: no need for globby?

packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -101,10 +102,20 @@ function createServerEntries(context: string, cwd: string = process.cwd()) {
       })
     : [];
 
+  let initServerPath;
+  try {
+    initServerPath = resolve.sync(path.join(cwd, SRC_DIR, 'init-server'), {
```
ronami: Super small: instead of using both `resolve()` and `path.join()` we can only have `resolve()` and use the `basedir` option (https://github.com/browserify/resolve#resolveid-opts-cb).

packages/yoshi-server/src/server.ts
```diff
@@ -45,6 +47,22 @@ export default class Server {
...
+
+  public initServer: () => Promise<void> = async () => {
+    const initServerPath = path.resolve(BUILD_DIR, 'init-server.js');
+
+    if (await fs.pathExists(initServerPath)) {
+      const chunk = importFresh(initServerPath) as InitServerFunction;
+
+      this.initData = await chunk(this.context);
+    }
+  };
```
ronami: Super small: I think we can inline this function into `create()`, if we keep it separate we can change to `private`.

Issue #2043, opened 2020-03-04
[RFC] Yoshi flow editor: Develop your app locally.


yavorsky:
This RFC related to really important part of development process. By developing the app we want to not rely on additional environment not related to our widget, but  on the same time, be sure app will work as expected after we publish it.

Currently, the best way to develop OOI app is to run it in viewer and editor environment, provide HMR (in progress) and point platform part to download `viewerScript`, `editorScript`, editor/settings apps endpoints to local ones. On the same time, we should have production version pointed to stable builds from CDN.

Currently, from what I saw we have 2 different flows from team to team:
### Maintain 2 Apps in Dev Center.
One app is configured to local environment, second - to prod one. It's really easy to set-up. You just need to go to dev center, copy paste values you need, create same components. Then you probably need to have at least 2 websites for each app to verify everything is working fine.
Of course it will create some inconveniences and potential issues:
- You should maintain 2 apps on dev center. After you added new widget, you need to add it to second app. After 
- You should maintain 2 apps in the project. Currently, you should handle additional logic for viewerScript:
*example from achievements app*

```ts
const prodAppConfig: IAppConfig = {
  widgetId: 'achievements_dev',
  pageId: 'challenge_page',
  widgetType: 'ef4e59bd-625a-484e-9ba1-650d4aecf03b',
  pageType: '847c2f5d-017f-499f-9c3e-069164fc116f',
};

const devAppConfig: IAppConfig = {
  widgetId: 'challenges_dev_widget',
  pageId: 'challenges_dev_page',
  widgetType: '4224a7c4-3783-466b-be3e-b0826d2225da',
  pageType: '642217d5-f1e7-476a-8c55-7241da8e2267',
};

export const APP_CONFIG: { [key: string]: IAppConfig } = {
  [APPS.Prod]: prodAppConfig,
  [APPS.Dev]: devAppConfig,
};

// Later in viewerScript
const appConfig = APP_CONFIG[config.appParams.appDefinitionId];
```
Moreover, with new editor flow, where we want to maintain `.component.json` file for each widget, we should extend it to be something like

*.component.json*
```js
{
  "local": {
      "id": "LOCAL_APP_ID",
    },
  "prod": {
      "id": "PROD_APP_ID",
    }
}
```
 so we'll configure viewerScript to map widgets correctly under the hood.
![excalidraw-202034141034](https://user-images.githubusercontent.com/1521229/75879132-823b1d80-5e23-11ea-9274-0bf290b41c05.png)

[...]

PR #2110, opened 2020-03-17
Override viewerScript, viewerWidget, editor and settings urls for local mode.

yavorsky:
This is a c/p from https://github.com/wix/yoshi/issues/2043 which fully cover the intention of this PR.

This PR allows `yoshi start` auto open viewer and editor urls with needed query params to override. Moreover, we can override multiple widgets.

Our flow will match requirements that I see really convenient:
- We don't need to maintain second app
- We don't need to maintain Client Spec Map Experiments, creating new spec for each project, creating new experiment etc.
- Overrides are flexible according to local project structure. After adding / removing widget locally, new url will represent relevant override map.
- We can create a tunnel to localhost or verify ip is static, public and use this value instead of localhost for viewer urls override. This will solve issue when SSR fallbacks to CSR or use public version on for SSR and local for CSR.

So for now, we can use
- `tpaWidgetUrlOverride` and `tpaSettingsUrlOverride` to override editor app and settings pannel apps. 
- `widgetsUrlOverride` to override viwerWidget bundle.
- `viewerPlatformOverrides` to override viewerScript bundle.

Platform changes:
[Santa PR](https://github.com/wix-private/santa/pull/46987) (allow override wigetUrl)
[Editor PR](https://github.com/wix-private/santa-editor/pull/26890) (allow pass query params to editor)
[Document management PR](https://github.com/wix-private/document-management/pull/1881)

excalidraw-202034141034 (4)


packages/yoshi-flow-editor/src/commands/start.ts
```diff
@@ -34,7 +36,7 @@ const start: cliCommand = async function(argv, config, model) {
 
   const {
     '--help': help,
-    '--url': url,
+    '--url': urlFromCliArgs,
```
yavorsky: Initially I decided to remove this option at all and use `startUrl` from config. But for now I can see that it will allow quickly running specific environment with installed app.
Still thinking about it and we can remove it in future.

Open for discussion.

packages/create-yoshi-app/templates/flow-editor/javascript/.application.json
```diff
@@ -0,0 +1,3 @@
+{
+  "appDefinitionId": "fda9208a-85a4-4482-9f43-59599773e998"
```
yavorsky: We should generate it in future.

packages/yoshi-config/src/config.ts
```diff
@@ -16,6 +16,8 @@ type CdnConfig = {
   ssl: boolean;
 };
 
+export type StartUrl = string | Array<string> | null;
```
yavorsky: We need it in flow-editor :)

packages/create-yoshi-app/templates/flow-editor/javascript/.application.json
ranyitz: You mean, generate a new App in dev-center?

packages/yoshi-flow-editor/src/utils.ts
```diff
@@ -0,0 +1,77 @@
+import path from 'path';
+import { URL } from 'url';
+import { StartUrl } from 'yoshi-config/build/config';
+import { FlowEditorModel, ComponentModel } from './model';
+
+const devServerUrl = 'https://localhost:3000';
```
ranyitz: How do you know it will be on `3000`?

ranyitz:
@yavorsky Please note that the tests are failing (`editor-flow`)

packages/create-yoshi-app/templates/flow-editor/javascript/.application.json
yavorsky: As a first step, we can allow users to fill it when creating new app.

packages/yoshi-flow-editor/src/utils.ts
yavorsky: Good point, I'll find a way to receive it.

yavorsky: Not sure about solution, but seems like we just take it from process.env.PORT, which is being generated according to `PORT` from `yoshi-commin/utils/constants`.

ranyitz: Yes

PR #2112, opened 2020-03-17
[create-yoshi-app:dev]: Improve symlink strategy to ignore development types

ranyitz:
### 🔦 Summary
This PR fixes a problem that we've had with `@types` clashes when working on `create-yoshi-app:dev` experience.

In order to understand the problem you need to understand the background.

We want to work locally and test our template and our modules, for that, we want to generate a project and symlink all of Yoshi's local modules (`yoshi-flow-app`, `yoshi-config` `yoshi-common` etc...) and install the rest. While this is handles beautifully, those modules also have dependencies. Those dependencies are being symlinked by a heuristic of symlink yoshi's `node_modules` directory.

The problem is, that we also symlink yoshi's dev dependencies, which includes `@types`.

In my specific case, is seems that TypeScript cannot handled a duplicated declaration for a global variable (which makes sense) because it finds that those definitions are different.

image

The user should have never get `jest-environment-puppeteer`'s types, and therefore I wrote some code that makes sure to symlinks only the `@types` which are defined as dependencies of yoshi's modules.

this is a heuristic and would not catch 100% of the cases. But it catches 100% of the cases we have right now and therefore works for our use-case.

I would love to enhance it in the future, but there is no need for that right now.

It also transforms the `symlinkModules.js` to TypeScript


ranyitz:
@yavorsky & @amitdahan I guess that `editor-flow` and `bm-flow` will encounter similar problems.

Please update if this strategy doesn't hold, and we'll need to replace it with something smarter :)

amitdahan:
Looks good! Finally we can rejoice in our dev flow! 😄

PR #2113, opened 2020-03-17
add specific test command for library flow

ranyitz:
### 🔦 Summary
This PR adds a specific test command for library flow.

This is the minimum required to run `jest` and bootstrap the `cdn` for the `e2e` tests.


packages/jest-yoshi-preset/jest-preset.js
```diff
@@ -221,4 +222,8 @@ const config = {
   ],
 };
 
+if (inTeamCity()) {
+  config.testResultsProcessor = require.resolve('jest-teamcity-reporter');
```
ranyitz: We're adding this in legacy test command, but in my opinion it can be a feature of the preset as well, it's not breaking.

packages/yoshi-common/src/cdn.ts
```diff
@@ -2,10 +2,9 @@ import fs from 'fs';
 import http, { IncomingMessage, ServerResponse } from 'http';
 import https from 'https';
 import serveHandler from 'serve-handler';
-import { Config } from 'yoshi-config/build/config';
 import { STATICS_DIR } from 'yoshi-config/build/paths';
 
-export async function startCDN(config: Config) {
+export async function startCDN({ port, ssl }: { port: number; ssl: boolean }) {
```
ranyitz: I want to use this method without committing to a specific config structure.

ranyitz:
@yanivefraim Thanks! We'll need to add tests for the whole library flow. The integration test infra should fit like a glove!

PR #2117, opened 2020-03-18
[flow-library]:  Bundle name will be  [name].umd.js

ranyitz:
### 🔦 Summary

This PR rewrites the `umd` bundle name to be `[name].umd.js` which is much clearer than `app.bundle.js`

We also need to remove the organization (e.g. `@wix/package-name` => `package-name`)


packages/yoshi-flow-library/src/config/normalize.ts
```diff
@@ -20,6 +20,7 @@ export default (initialConfig: InitialConfig, pkgJson: PackageJson): Config => {
     url: cdnUrl,
     umd,
     bundle: !!initialConfig.bundle,
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
amitdahan: Is this needed here? Where's the non-null thing?

ranyitz: Nice catch, I'll remove it.

I want to disable this rule all over the codebase, we are using `!` broadly, and I don't think it's such a bad practice in those places.

packages/yoshi-flow-library/src/webpack.config.ts
```diff
@@ -46,9 +52,13 @@ export function createClientWebpackConfig(
     separateCss,
   });
 
-  clientConfig.entry = isSingleEntry(entry) ? { app: entry as string } : entry;
+  clientConfig.entry = isSingleEntry(entry)
+    ? { [stripOrganization(name)]: entry as string }
+    : entry;
 
   clientConfig.externals = config.externals;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
ranyitz:
```suggestion
```

packages/yoshi-flow-library/src/config/normalize.ts
ranyitz:
```suggestion
```

PR #2132, opened 2020-03-22
CDN deployment conditional: Bring back `BUILD_VCS_NUMBER`

Schniz:
### 🔦 Summary
Removing `BUILD_VCS_NUMBER` in #2059 broke Perfer CI in Thunderbolt.
Let's return it and discuss afterwards whether we want to add another environment variable to Perfer CI:

* Adding a fake `ARTIFACT_VERSION` env var can do strange things to some projects' build pipelines
* Adding a fake `SRC_MD5` is very weird

### 🗺️ Test Plan
After releasing a new version, Perfer builds will pass again.


Schniz:
cc @NitayRabi is there a problem with that?

ronami:
cc @ranyitz  @netanelgilad

netanelgilad:
It's true that not all CI builds have SRC_MD5 and we need to fallback to BUOLD_VCS_NUMBER but that should be the order, SRC_MD5 should take precedence.

packages/yoshi-helpers/src/utils.ts
```diff
@@ -193,8 +193,8 @@ export const getProjectArtifactVersion = () => {
   return (process.env.ARTIFACT_VERSION
     ? // Dev CI
       process.env.ARTIFACT_VERSION.replace('-SNAPSHOT', '')
-    : // PR CI won't have a version, only SRC_MD5
-      process.env.SRC_MD5) as string;
+    : // PR CI won't have a version, only VCS_BUILD_NUMBER or SRC_MD5
+      process.env.SRC_MD5 || process.env.VCS_BUILD_NUMBER) as string;
```
Schniz: @netanelgilad SRC_MD5 is before

netanelgilad: yep github mobile got me confused :D

Schniz:
We're adding SRC_MD5 in Perfer CI

PR #2133, opened 2020-03-22
Don't transpile with ts-node/register files ignored by git

ronami:
### Summary

This PR adjusts the changes in #2051 and makes `ts-node` ignore any file ignored by git.

This is needed since Thunderbolt (and possibly other projects) consume output from their `dist` or other directories from their startup script. This causes the startup script to work slower (since we're transpiling JavaScript bundles that don't need it) but also breaks source maps.

Because `ts-node` only supports regex patterns for ignoring specific paths, (and because it's not simply ignoring the `dist` directory for Thunderbolt), I created custom `require.extension`s that filter files according to `.gitignore` files in the project and pass it over to `ts-node`.


packages/yoshi-common/src/require-hooks/ts-node-register.ts
```diff
@@ -1,4 +1,9 @@
-require('ts-node').register({
+import globby from 'globby';
+import { register } from 'ts-node';
+
+const originalJsHandler = require.extensions['.js'];
+
+const { extensions } = register({
```
ronami: returns the extensions used by `ts-node`, can be either `js` (if `allowJs` is true), `ts`, and `tsx` (if it's enabled)

packages/yoshi-common/src/require-hooks/ts-node-register.ts
```diff
@@ -9,3 +14,20 @@ require('ts-node').register({
     target: 'es2017',
   },
 });
+
+// don't transpile with ts-node/register files ignored by git
+const shouldIgnore = globby.gitignore.sync({ cwd: process.cwd() });
```
ronami: awesome globby has that feature, was about to write it myself

packages/yoshi-common/src/require-hooks/ts-node-register.ts
```diff
@@ -9,3 +14,20 @@ require('ts-node').register({
...
+
+// `ts-node` only supports regex ignore patterns, use custom extensions so functions can
+// be used
+extensions.forEach(ext => {
```
ronami: should probably be a flag in `ts-node` just like it is in `globby`

Issue #2146, opened 2020-03-24
[RFC] BoB apps support for yoshi-flow-editor


yavorsky:
#### This issue related to best-of-both apps support for editor flow

Terminology (will be updated):
**bob** - Best-of-both. Best of platform and app builder. We can build platform apps using app-builder and import it in editor.
**ooi**  - Out of iframe. Apps that's running in viewer not wrapped in iframe. Still wrapped in iframe in editor environment.

Seems like we can reuse a lot of logic between platform apps and ooi apps. Since we have different UI part (widgets), we still can use controllers, which is automatically collected and output as `viewerScript.bundle.js`.
Moreover we're working on `editorScript` support. At first iteration, it will be just an entry-point for `editorScript.bundle.js`. Users will control what this module is exporting, and what additional logic contains. The only thing - we need to prevent breaking changes in future to prevent people pinning `yoshi-flow-editor` version 👀and always get fresh feedback. So it will be our wrapper, which is basically pass everything from user's module.

Here is proposed update for current template. It's very raw and I can miss some things. But the idea is to make everything simple and consistent. Moreover, it would be really cool if some teams that has few components wanted to migrate from OOI to platform components can use the same app and do it consequently.

```
src
   components
        some-platform-component
           .component.json
            Settings.ts
            viewer.controller.ts
            editor.controller.ts
        some-ooi-component
           .component.json
            Settings.ts
            controller.viewer.ts (former controller.ts)
            Widget.ts
editor.app.ts
viewer.global.ts
.application.json
```
We've tried to find what could be common for these types of components.
Let's take [conferencer-bob app](https://github.com/wix-private/conferencer-bob/tree/master) as an example.

#### viewerScript (controller.viewer.ts)
[Widgets directory](https://github.com/wix-private/conferencer-bob/tree/0ca75a0c56d4ddf3603277fc31aa35647e787b30/src/widgets/speakersWidget) contains 2 types of widgets `.viewer.ts` and `.editor.ts`. Initially I thought abut it as a widget replacement. But actually it's controllers. Widget part is hidden from us at all.
It doesn't render anything, but just configure in declarative (editor) or imperative(viewer) our widgets. We even load viewer controller via `viewerScript.bundle` [inside createControllers function](https://github.com/wix-private/conferencer-bob/blob/0ca75a0c56d4ddf3603277fc31aa35647e787b30/src/viewerApp/viewer.js#L20)

[...]

PR #2151, opened 2020-03-25
Add editor script support for yoshi-flow-editor

yavorsky:
### 🔦 Summary
This PR allows user to create `src/editor.app.ts`, which will be applied on the app level.
It doesn't contain any breaking changes, but adds new `editorScript.bunle.js` in the `dist` if user specified `editor.app` file.
Moreover was added new query param to override editorScript on the platform side.

Platform PRs:
- santa-editor - https://github.com/wix-private/santa-editor/pull/27120 ✅
- document-managment -https://github.com/wix-private/document-management/pull/3118 ⌛️

> I'm going to include this experience in PR related to documentation update.

This is also a part of https://github.com/wix/yoshi/issues/2146 implementation

Real-app usage PR: https://github.com/wix-private/strategic-products/pull/3664


packages/yoshi-flow-editor/src/buildEditorEntires.ts
```diff
@@ -2,6 +2,7 @@ import path from 'path';
 import writeComponentWrapping from './wrappers/componentWrapping';
 import writeEditorAppWrapping from './wrappers/editorAppWrapping';
 import writeSettingsWrapping from './wrappers/settingsWrapping';
+import writeCommonEditorScriptWrapping from './wrappers/commonEditorScriptWrapping';
```
yavorsky: It's called common to be:
- consistent with viewerScript wrapper
- in future we are going to add each wrapping per widget for bob apps support

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -56,16 +58,27 @@ export async function generateFlowEditorModel(
     throw new Error(`artifact id not provided.
     Please insert <artifactId>yourArtifactId</artifactId> in your "pom.xml"`);
   }
+  const fileExtension = isTypescriptProject() ? 'ts' : 'js';
```
yavorsky: Not related to the topic, but just a small enhancement.

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -56,16 +58,27 @@ export async function generateFlowEditorModel(
...
 
   const rootPath = process.cwd();
-  const initApp = resolveFrom(path.join(rootPath, 'src'), 'app');
+  const srcPath = path.join(rootPath, 'src');
+  let initApp = resolveFrom(srcPath, 'init.app');
+  if (!initApp) {
+    initApp = resolveFrom(srcPath, 'app');
```
yavorsky: Since we already have 1 prod app, I don't want to break it :)

packages/yoshi-flow-editor/src/wrappers/templates/CommonEditorScriptEntry.ts
```diff
@@ -0,0 +1,7 @@
+import t from './template';
+
+type Opts = Record<'editorEntryFileName', string>;
+
+export default t<Opts>`
+  export * from '${({ editorEntryFileName }) => editorEntryFileName}';
```
yavorsky: Instead of just specifying user's file as an entry-point, I would like to use our wrapper even if it is doing nothing now. Then it will be easier to add more stuff here.

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -1,15 +1,17 @@
...
 import { getProjectArtifactId } from 'yoshi-helpers/utils';
+import { isTypescriptProject } from 'yoshi-helpers/build/queries';
 import resolve from 'resolve';
 import fs from 'fs-extra';
 import { Config } from 'yoshi-config/build/config';
 
 export interface FlowEditorModel {
   appName: string;
-  appDefId: string;
+  appDefId: string | null;
```
ranyitz: Can it be `null`, when?

yavorsky: yeah, when people for some reason didn't specify appDefinitionId in `.application.json` or doesn't have this file. Since it's needed only for overrideUrl, we can just warn people to create it.

yavorsky: We will be more strict in future when things like appDefId will be generated by us.

ranyitz: Makes sense.

I think it's a bit better to have a normalizer for it.
Which returns the EditorFlowModel with all the values that needs to be there. (no `null` values for things that are required in the flow)

When loading it, it will throw an error in case something isn't right.

Not super important, but I think we'll make the management of the codebase a bit nicer

PR #2154, opened 2020-03-26
re-structure the config for library flow

ranyitz:
### 🔦 Summary
This PR changes the config to look as follows:
```js
{
	bundle: InitialBundleConfig | boolean,
	storybook: boolean
}
```
```js
InitialBundleConfig = {
  library?: Library;
  entry?: WebpackEntry;
  externals?: WebpackExternals;
  port?: number;
  https?: boolean;
};
```

If the user specify `bundle: false` or doesn't specify bundle at all, they would not get a `umd` bundle.

In case they do specify it with `bundle: true` they get all the defaults which takes the name of the package as the `library` name which will be exposed on the global.


packages/yoshi-flow-library/src/commands/test.ts
```diff
@@ -2,16 +2,20 @@
...
-import {
-  hasBundleInStaticsDir,
-  hasE2ETests,
-} from 'yoshi-helpers/build/queries';
+import { hasE2ETests } from 'yoshi-helpers/build/queries';
+import { STATICS_DIR } from 'yoshi-config/build/paths';
+import globby from 'globby';
 import { cliCommand } from '../cli';
 
+const hasBundleInStatics = () => {
```
yavorsky: Maybe it's worth to make `hasBundleInStaticsDir` from `yoshi-helpers` more flexible instead of creatine another helper?

packages/yoshi-flow-library/src/config/normalize.ts
```diff
@@ -1,28 +1,44 @@
...
-  const cdnUrl = `${cdnHttps ? 'https:' : 'http:'}//localhost:${cdnPort}/`;
+  const bundleDefaults: InitialBundleConfig = {
+    port: 3300,
+    https: false,
+    entry: 'index.ts',
+    externals: [],
+    library: pkgJson.name,
+  };
+
+  let bundleConfig: BundleConfig | false = false;
```
yavorsky: Maybe let's convert user's `false` to internal `null`? I mean here `bundleConfig` is something we are handling, but `bundle: false` is an user-friendly option that should be converted to `null` the same way we are converting `true` to default config.

packages/yoshi-flow-library/src/config/types.ts
```diff
@@ -6,27 +6,35 @@ type WebpackEntry = string | Array<string> | Entry | EntryFunc;
...
   storybook?: boolean;
 };
 
+type PartialWithMember<T, M> = {
+  [P in keyof T]?: T[P];
+} &
+  M;
+
+export type BundleConfig = RequiredRecursively<
+  PartialWithMember<InitialBundleConfig, { url: string }>
```
yavorsky: Maybe instead of custom helper,

`Partial<InitialBundleConfig> & { url: string }` ?

packages/yoshi-flow-library/src/commands/test.ts
ranyitz: We can and we should, for now I prefer having this piece of logic duplicated, and fix that later on.

packages/yoshi-flow-library/src/config/types.ts
ranyitz: Makes sense, thanks!

packages/yoshi-flow-library/src/config/normalize.ts
ranyitz: Now when I read your comment, having this with a `null` makes more sense.

PR #2158, opened 2020-03-27
fix(yoshi-flow-editor-runtime): Fix ssr by checking `window` presence.

yavorsky:
### 🔦 Summary
Currently ssr is not working for new flow.
@bordi  and me debugged it and figured out that the issue is that we don't have `window` variable in ssr environment. Moreover, `react-dom-server` we have on bolt side is not supporting `React.Suspense`.

> Also note. We shouldn't fix it now, but consider in future:

We also have an issue with HMR environemnt for bolt's SSR. If we are going to use suricate to create a tunnel to local environemnt for SSR part, it's going to be broken since bundle will contain HMR stuff that is not compatible with node environment we are running ssr in.


bordi:
Seems legit

PR #2163, opened 2020-03-29
[library-flow]: Fix line and column order for link to issue feature

ranyitz:
### 🔦 Summary
This PR fixes a bug where the link generated by the formatter was lead to the wrong location on the file due to the wrong line and column order.

This also removes the white background from the file name because it is more identifies as a link, also vscode integrated terminal has a hover effect which cannot be seen with the white background.

The state after the fixes can be seen in the following gif:

ts formatter bug

___________

There is another important change added to this PR.

There are some cases, when TypeScript doesn't supply the regular format of the error, and therefore we can't parse the error. In those cases, I prefer showing something to the user (even if it's not formatted very nicely). Also, this prevent the watcher from exiting on error, which is the right thing to do here.


packages/yoshi-common/src/typescript/formatter.ts
```diff
@@ -35,7 +35,7 @@ export function formatTypescriptError(errorString: string) {
     isWarningSeverity: () => false,
   };
 
-  const errorLocation = chalk.bgWhite.black(`${fileName}(${column},${line})`);
+  const errorLocation = chalk.underline(`${fileName}(${line},${column})`);
```
yavorsky: 🦄

PR #2167, opened 2020-03-30
feat(yoshi-flow-editor): Add migraiton guide

yavorsky:
### 🔦 Summary
This PR contains migration guide for user's who already created OOI apps using old flow, but want to be cool and modern and use new flow.

Also added our **brand new** marketing wallpaper.

E52C44BF-55C7-407E-AA9C-0AE07E1B7A71


packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
+## Editor flow migration guide
+
+#### Intro
+![E52C44BF-55C7-407E-AA9C-0AE07E1B7A71](https://user-images.githubusercontent.com/1521229/77897090-e2d24480-7281-11ea-9cbf-e3fdc4d8c509.PNG)
```
ranyitz:
```suggestion
![E52C44BF-55C7-407E-AA9C-0AE07E1B7A71](https://user-images.githubusercontent.com/1521229/77897090-e2d24480-7281-11ea-9cbf-e3fdc4d8c509.PNG)

#### Intro
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+
+This is a migration guide for apps created via `out-of-iframe` option using `create-yoshi-app`.
+New flow removes a lot of boilerplate, so prepare your `CMD` and `DELETE` buttons on the keyboard 😉.
```
ranyitz:
```suggestion
Editor flow removes a lot of boilerplate, so prepare your `CMD` and `DELETE` buttons on the keyboard 😉.
```

ranyitz: > prepare your `CMD` and `DELETE` buttons

Oh yeah!! 💯

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+
+The project structure will be updated a bit also. We are going to group everything related to a single widget by directories.
```
ranyitz:
```suggestion
The project structure will be updated as well. We are going to group everything by widgets, All things related to a single widget will be in a single directory.
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+This guide supposes you preserve structure provided by ooi template, so if you created additional directories not related to current flow, please verify it's needed and update paths in imports.
```
ranyitz:
```suggestion
This guide supposes you preserve the structure provided by the `ooi` template, so if you created additional directories, please verify they're needed and update the relevant paths in `import` statements.
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+
+Ok, let's start 🚀
+
+##### Update package.json
+- Remove `yoshi` dependency
```
ranyitz:
```suggestion
- Remove `yoshi` devDependency: `npm uninstall --save-dev yoshi`
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+- Add `yoshi-flow-editor` devDependency: `npm install --save-dev yoshi-flow-editor`
+- Add `yoshi-flow-editor-runtime` dependency: `npm install --save yoshi-flow-editor-runtime`
+- Update `yoshi` to `yoshi-flow-editor` in `scripts`
```
ranyitz: It's not related to the PR, but we have to decide about this because we're creating both `yoshi-library`, `yoshi-app`, `yoshi-bm` and now `yoshi-flow-editor`.

Let's get to a decision so we won't confuse everyone.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
++  "yoshi-flow-editor": "^4.0.0"
+},
+"dependencies": {
++  "yoshi-flow-editor-runtime": "^4.0.0"
+...
+} 
+```
+
+##### Remove all boilerplate files:
+- Go to `viewerApp` directory. Move all test files to `__tests__` directory in the root. Remove `viewerApp` directory. yoshi-flow-editor will generate viewerScript for you!
```
ranyitz:
```suggestion
- Go to `viewerApp` directory. Move all test files to `__tests__` directory in the root. Remove `viewerApp` directory. `yoshi-flow-editor` will generate the `viewerScript` for you!
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+- Provide the same steps for `editorApp` and `settingsPanel` directories. *We are going to generate the `editor app`/`settings panel` wrappers under the hood.*
+- Remove `templates` directory. *New flow will render editor app and settings panel w/o template files.*
```
ranyitz:
```suggestion
- Remove `templates` directory. *Editor flow will render editor app and settings panel w/o template files.*
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+##### Group files by components:
+If you have widget-separated monorepo configured we recommend collecting all projects into a single one.
```
ranyitz:
```suggestion
If you have widget-separated monorepo configured we recommend combining all projects into a single one.
```

I think we should add a "please talk to us in case this setup doesn't work for you and you want to keep the widget separated."

We know that there are, and we would like to know about those cases.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+- In `src/components` directory consider every directory represent each widget. For single-widget apps, you need to rename your `Widget` directory to the name of the component. Try to create simple name related to the app in general. As an example, if your app is Pizza 🍕 instead of `PizzaSauce` and `PizzaArtichoke`, call it `Sauce` and `Artichoke`.
```
ranyitz:
```suggestion
- Every directory in `src/components` represents a single widget. For single-widget apps, you need to rename your `Widget` directory to the name of the component. Try to create directory names that don't include the app name. For example, if your app is called "Pizza" 🍕 instead of `PizzaSauce` and `PizzaArtichoke`, call it `Sauce` and `Artichoke`.
```

> For single-widget apps, you need to rename your `Widget` directory to the name of the component

I'm not sure I understand what to do, maybe provide an example?

>  Try to create directory names that don't include the app name

Maybe it worth explaining why. Does this effect the name in the devCenter? just a general best practice?

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+- Verify your component's directory has 3 main files (or directories with `index.ts`):
+  - `Widget` - React component which will be rendered in editor and viewer.
```
ranyitz:
```suggestion
  - `Widget` - `export default` A React component which will be rendered in editor and viewer.
```

A believe that this makes less place for mistakes

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+  - `controller.ts` - Component's logic, which will be executed in Web Worker environment.
```
ranyitz: I would add what we expect this file to contain. (An async function?)

ranyitz: And maybe we should add links to the general docs that shows what should be in this file

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+  - `Settings` - Settings panel. Will be rendered in editor after double-clicking on component or picking "Settings" tab.
```
ranyitz: It worth saying like in the other two above, that it's a default export to a react component that renders the settings panel...

Also here, a link to example/Types would be good

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+- Create `.component.json` file for each component (in component's dir) and paste you widget id as an `"id"` field. You probably can take this value from the former `viewerApp` directory. Usually you need it to match `createController` function with the platform's `controllerConfig`. Also, you can take it from http://dev.wix.com
```
ranyitz:
```suggestion
- Create `.component.json` file for each component (in component's dir) and paste your widget id as an `"id"` field. You probably can take this value from the former `viewerApp` directory. Usually you need it to match `createController` function with the platform's `controllerConfig`. Also, you can take it from http://dev.wix.com
```

It would be good to have a link to an example, I know this is simple, but an example can guarantee this would be smooth.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,56 @@
...
+- Create `.application.json` file in the project's root. Add `"appDefinitionId": YOUR_APP_DEFINITION_ID` here.
+
+##### Done ✅
```
ranyitz: I think it worth mention that they should probably open this under an experiment first... Or, how to gradually test it in production.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
yavorsky: Yeah, as I said `yoshi-editor` will confuse a lot of people, b/c it could be perceived as editor for yoshi.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
yavorsky: Added `Making all changes smooth and safe` section

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
yavorsky: Added. Not sure about example, since it's just simple React component the same people already have.

yavorsky:
I don't see any difference between `page` and `widget`. The only thing is a bit different configuration in dev center. But it's the same react component with settings and controller.
I would like to deprecate it in future and not mention anywhere that we can have `Page.ts` instead of `Widget.ts` since for us and for users it will be no difference.

yavorsky:
For widget vs component:
I think a lot people using old dev center and old ooi template more familiar with `widget` terminology. But new flow clarifies it a bit: you can have components directory with both widgets and pages.

But I don't see any difference between page and widget. The only thing is a bit different configuration in dev center. But it's the same react component with settings and controller.
I would like to deprecate it in future and not mention anywhere that we can have Page.ts instead of Widget.ts.

Maybe it's worth to add `type = 'widget'` to `.component.json` config file in future.

ranyitz:
@yavorsky I believe you, that for what we know there is not difference. I'm willing to remove this, but as making changes now is relatively easy (1 active user). It will be harder once we'll have 2 or more.

Maybe it makes sense deprecated it and do what you suggest.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,84 @@
...
+- *viewerWidget*: `baseStaticsUrl/{name}ViewerWidget.bundle.min.js`. You'll have each bundle per component. Here `name` is a simply name of the component's directory. For example for directory called `Artichoke`, bundle name will be `ArtichokeViewerWidget.bundle.js`.
+- *editorApp*: `baseStaticsUrl/editor/{name}.html`. For example `editor/Artichoke.html`
+- *settingsPanel*: `baseStaticsUrl/settings/{name}.html`. For example `settings/Artichoke.html`
+
+Update your components in dev center to match requirements above. As a baseStaticsUrl, you can usually use static.parastorage by knowing you `artifactId`.
+For example `https://static.parastorage.com/services/{artifactId}/1.0.0/viewerScript.bundle.min.js`
+
+##### Making all changes smooth and safe
+Since any migration is a risk, we can greatly reduce chances to break production using some of these approaches:
+- Instead of updating current app, create new one with own artifact. Add new app in Dev Center. Verify it's working both locally and on the production. Please note, you'll still have `DRAFT` type of the app. Since you probably don't want to publish this app to the real world app market.
```
ranyitz:
```suggestion
- **Dry run:** Instead of updating current app, create new one with own artifact. Add new app in Dev Center. Verify it's working both locally and on the production. Please note, you'll still have `DRAFT` type of the app. Since you probably don't want to publish this app to the real world app market.
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,84 @@
...
+- [Create new experiment](https://github.com/wix-private/fed-handbook/blob/master/EXPERIMENTS.md) and add it to [Client Spec Map Experiments](https://bo.wix.com/client-spec-map-experiment-server/). Override all urls to the new artifact and enable it to limited audience. Please check [the Client Spec Map Experiments guide](https://github.com/wix-private/meta-site/tree/master/wix-meta-site-server/app-store-service/client-spec-map-experiment-server#change).
+- Just run `npm start` locally and verify everything is working fine. This will override all needed bundles with your local one.
```
ranyitz: This is duplicated with the `done` section

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,84 @@
...
+
+Update your components in dev center to match requirements above. As a baseStaticsUrl, you can usually use static.parastorage by knowing you `artifactId`.
+For example `https://static.parastorage.com/services/{artifactId}/1.0.0/viewerScript.bundle.min.js`
+
+##### Making all changes smooth and safe
+Since any migration is a risk, we can greatly reduce chances to break production using some of these approaches:
+- Instead of updating current app, create new one with own artifact. Add new app in Dev Center. Verify it's working both locally and on the production. Please note, you'll still have `DRAFT` type of the app. Since you probably don't want to publish this app to the real world app market.
+- [Create new experiment](https://github.com/wix-private/fed-handbook/blob/master/EXPERIMENTS.md) and add it to [Client Spec Map Experiments](https://bo.wix.com/client-spec-map-experiment-server/). Override all urls to the new artifact and enable it to limited audience. Please check [the Client Spec Map Experiments guide](https://github.com/wix-private/meta-site/tree/master/wix-meta-site-server/app-store-service/client-spec-map-experiment-server#change).
```
ranyitz:
```suggestion
- **Gradually open it using `cliene-spec-map`:** [Create new experiment](https://github.com/wix-private/fed-handbook/blob/master/EXPERIMENTS.md) and add it to [Client Spec Map Experiments](https://bo.wix.com/client-spec-map-experiment-server/). Override all urls to the new artifact and enable it to limited audience. Please check [the Client Spec Map Experiments guide](https://github.com/wix-private/meta-site/tree/master/wix-meta-site-server/app-store-service/client-spec-map-experiment-server#change).
```

yavorsky:
@ranyitz Let's remove it. But for now we don't mention it anywhere. And in this doc especially. So to `remove` it, we just need to not consider `Page.ts` when building our app model. But it's a bit different change.

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
yavorsky: Just a general best practice. Full url will already contain artifact, so it's not needed. I can add it to doc

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
yavorsky: Removed

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -75,7 +75,7 @@ For example `https://static.parastorage.com/services/{artifactId}/1.0.0/viewerSc
...
-- [Create new experiment](https://github.com/wix-private/fed-handbook/blob/master/EXPERIMENTS.md) and add it to [Client Spec Map Experiments](https://bo.wix.com/client-spec-map-experiment-server/). Override all urls to the new artifact and enable it to limited audience. Please check [the Client Spec Map Experiments guide](https://github.com/wix-private/meta-site/tree/master/wix-meta-site-server/app-store-service/client-spec-map-experiment-server#change).
+- **Gradually open it using `cliene-spec-map`:** [Create new experiment](https://github.com/wix-private/fed-handbook/blob/master/EXPERIMENTS.md) and add it to [Client Spec Map Experiments](https://bo.wix.com/client-spec-map-experiment-server/). Override all urls to the new artifact and enable it to limited audience. Please check [the Client Spec Map Experiments guide](https://github.com/wix-private/meta-site/tree/master/wix-meta-site-server/app-store-service/client-spec-map-experiment-server#change).
```
ranyitz:
```suggestion
- **Gradually open it using `client-spec-map`:** [Create new experiment](https://github.com/wix-private/fed-handbook/blob/master/EXPERIMENTS.md) and add it to [Client Spec Map Experiments](https://bo.wix.com/client-spec-map-experiment-server/). Override all urls to the new artifact and enable it to limited audience. Please check [the Client Spec Map Experiments guide](https://github.com/wix-private/meta-site/tree/master/wix-meta-site-server/app-store-service/client-spec-map-experiment-server#change).
```

packages/yoshi-flow-editor/docs/OOI_TO_EDITOR_FLOW_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,93 @@
...
+
+![E52C44BF-55C7-407E-AA9C-0AE07E1B7A71](https://user-images.githubusercontent.com/1521229/77897090-e2d24480-7281-11ea-9cbf-e3fdc4d8c509.PNG)
+
+
+#### Table of contents
+- [Intro](#intro)
+- [Update package.json](#update-packagejson)
+- [Remove all boilerplate files:](#remove-all-boilerplate-files)
+- [Group files by components:](#group-files-by-components)
+- [Update your urls in Dev Center](#update-your-urls-in--dev-center)
```
ranyitz:
```suggestion
- [Update your urls in Dev Center](#update-your-urls-in-dev-center)
```

PR #2168, opened 2020-03-30
Strip the organization name from the default library option

ranyitz:
### 🔦 Summary
The `library` config option (see webpack's [output.library](https://webpack.js.org/configuration/output/#outputlibrary)) controls the global variable name which the library will be exposes one.

Up until now, on every generated project it was `window['@wix/<name>']`.

This PR makes it so even if your `package.json` is `name: "@wix<name>"` the `umd` generated bundle will expose its main file exports under `window['<name>']`

________

While this is a possible breaking change, we still don't have users to it


packages/yoshi-flow-library/src/commands/start.ts
```diff
@@ -64,7 +64,7 @@ const start: cliCommand = async function(argv, config) {
   const devEnvironment = await DevEnvironment.create({
     webpackConfigs: webpackConfigs as [webpack.Configuration?],
     // @ts-ignore - we won't need this port if we won't have clientConfig
-    webpackDevServerPort: config.bundleConfig.port,
+    webpackDevServerPort: config.bundleConfig?.port,
```
ranyitz: This is not related, but needed for a case of `bundleConfig = "false"`

PR #2170, opened 2020-03-31
Support bundling `st.css` files in SSR

NitayRabi:
### 🔦 Summary

Add Stylable webpack plugin to handle `*.st.css` files in SSR with the following config - 

- Do not load actual styles into the bundle
- Do not output css (loading of css will be handled by client bundle)
- Do not use shared runtime

Once this is merged, editor-elements & thunderbolt can integrate with Stylable

@ronami - Would you like me to put it under experimental flag or specific case for Thunderbolt - I don't see how this will break current consumers.

@barak007 - Anything we missed here? is there another configuration we can/should use ?


giladsegal:
Nice job!
A single question - you are applying this change to node users - what would happen for existing projects with stylable that uses SSR? (do we have any?)

barak007:
This looks right for the current approach. 

1 it relies on the fact that some other flow will build the css bundle and include it in the served html
2. it assume css bundle is built with same stylable config.
3. since all 3rd party are external in this build it requires the ssr server to use the stylable require hook with same config.

maybe for the future better approach will be to reuse the client bundle in node. it will also be faster to bootstrap the server and will require less config (3rd parties included). also use html-webpack-plugin so all generated assets from webpack will be automatically injected to the html

NitayRabi:
@barak007 

> 3. since all 3rd party are external in this build it requires the ssr server to use the stylable require hook with same config.

Why do we need the hook? When I tried this configuration in thunderbolt it worked (the Client bundle generates the css, and thunderbolt is making sure it gets into the DOM)

barak007:
As far as I know there is no 3rd party stylable in thunderbolt.

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -66,6 +66,18 @@ function prependNameWith(filename: string, prefix: string) {
   return filename.replace(/\.[0-9a-z]+$/i, match => `.${prefix}${match}`);
 }
 
+const getCommonStylbleWebpackConfig = (name: string) => ({
+  filename: '[name].stylable.bundle.css',
```
ronami, line 70: Super small but I think that's probably specific to client bundle

test/typescript/features/stylable/ssr/ssr-st-css.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  projectType: 'typescript',
+});
+
+describe('Stylable SSR', () => {
+  it('Should allow importing *.st.css files in ssr', async () => {
+    await scripts.build();
+    const serverBundle = require(`${scripts.testDirectory}/dist/server.js`);
+    expect(serverBundle.style).toBeTruthy();
+    expect(Object.keys(serverBundle.style.classes).includes('root')).toBe(true);
+  });
```
ronami, line 16: Instead of testing the contents of the bundle as a string, we try to test it the same way our users will: By running it and expecting certain results.

Since we're trying to support SSR, I think we can set up a small server that will render (with `react-dom/server`) a react component that uses an `st.css` file. The client-side of that app will also render (hydrate) the same component. We can then make a request to that server from a browser (puppeteer) and see that everything renders correctly with no errors.

You can see how we do it in other tests: https://github.com/wix/yoshi/tree/master/test/javascript/features/env-vars.

Also, since we still support both JavaScript and TypeScript, we try to test new features with both transpilers. Can you please add the same test and setup to `test/javascript`?

test/typescript/features/stylable/ssr/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+  "compilerOptions": {
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "jsx": "react"
+  },
+  "files": [
+    "./src/test-types.d.ts"
+  ]
+}
```
ronami, line 11: I think this isn't needed. We use a base fixture for all TypeScript tests with a basic `tsconfig.json`: https://github.com/wix/yoshi/blob/master/test/fixtures/typescript/base-template/tsconfig.json. All `ts`/`d.ts` files inside the `src` directory are automatically included. Adding a `tsconfig.json` or any other file will override the default from the fixtures (if it exists).

ronami:
Looks great 🏆 

I don't think we have any users that use app-flow and import their own `st.css` files from the server bundle. Currently, if you try to import an `st.css` file from the server and use app flow you would get a compilation error. You will need to use a require hook if you require `node_modules` (like `wix-style-react`) that come with their own `st.css` files since by default we treat all server 3rd parties as external.

I don't think this is a breaking change since currently, projects that try to use it this way get a compilation error.

Issue #2155, opened 2020-03-26
Bundle size dashboard


Gongreg:
# Bundle size dashboard

Create some centralized dashboard which shows:
* bundles sizes of all applications/widgets. 
* all of the benchmark result of all of the applications/widgets

so that we can easily look at all data and understand where we need to improve.

# Investigation

## Webpack Bundle Analyzer

To get the sizes of bundles we could use already existing yoshi functionality (`yoshi build --analyze`). It gives us additional information about the bundle that we could later on track.

The time implications of using this plugin are small (10-20 additional seconds), the output that is generated shouldn't be more than few MB. 

There are few possible ways we could use it:

* Generate html output that we upload with relevant build information to the server. This would be the quickest win since it would also solve the rendering part in the dashboard. We just upload html, build info and simply render the html in the dashboard.

* Generate json that we save and then render ourselves. A bit more involving process, but would allow us to store information in a more accessible way. I couldn't find a renderer which would understand this output, so we would have to do additional work to support it.

## Webpack Stats Plugin

As an alternative we can use webpack stats plugin. This plugin allows us to reach all of the data that we want out of the webpack stats object. The plugin is also quite performant and quickly writes information to the json output (While checking I didn't feel big differences in duration). Also it is flexible and we can choose which information we wish to save. It is important to only save relevant information because the stats object itself can be over few hundred MB in a bigger project.

The negative sides are that using this plugin would involve biggest amount of work. Also it only return bundle sizes from the webpack stats, it does not show gzipped size (which is done by analyzer).

Overall it feels like just uploading html output to be displayed from dashboard is a fastest solution and a quickest win. Later on we could decide what additional information we wish to get out of stats.json and use WebpackStatsPlugin for that. 

## Build duration
There is something fishy happening while using `yoshi build --analyze`, it does not allow for the build to finish even when the  Webpack Bundle Analyzer plugin is removed. This didn't allow me to do proper comparison. Now I found that by skipping --analyze the build properly finishes, so we can do a full comparison if it is needed. But overall the difference was about 10-20 seconds on a 90 second build.

ranyitz:
@Gongreg This is a great idea.

I think I have an idea how it's can be done:
1. We can upload all webpack stats from all builds to the CDN. How? We'll just emit them in `dist/statics` and they will be there.
[...]

PR #2171, opened 2020-03-31
Webpack stats output to a file by default on CI

Gongreg:
### 🔦 Summary
Issue: #2155 

Added outputting of Webpack stats to a file.
To make sure that we have an escape hatch I've added an ENV flag disable check that we will be able to control from ci.

I've set the stats path to the same currently used one: `target/webpack-stats.json`

Also I've moved stats generation with flag `--stats` to use the same plugin 

### 🗺️ Test Plan
Tested manually on a project.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -54,6 +56,9 @@ const inTeamCity = checkInTeamCity();
 
 const disableModuleConcat = process.env.DISABLE_MODULE_CONCATENATION === 'true';
 
+const enableStatsOutput = process.env.ENABLE_WEBPACK_STATS_OUTPUT === 'true';
+const statsOutputPath = process.env.WEBPACK_STATS_OUTPUT_PATH;
```
ranyitz: I prefer having that as a convention.

Something like
```
dist/webpack-stats.json
```

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -54,6 +56,9 @@ const inTeamCity = checkInTeamCity();
 
 const disableModuleConcat = process.env.DISABLE_MODULE_CONCATENATION === 'true';
 
+const enableStatsOutput = process.env.ENABLE_WEBPACK_STATS_OUTPUT === 'true';
```
ranyitz: I prefer having that as an `opt-out` flag. For cases when it's problematic to do

packages/yoshi-common/src/webpack.config.ts
Gongreg: we don't want to save the file to dist though. Since we don't want to clutter the artifact with unnecessary data. So something like `../../webpack-stats.json` or `../../stats/webpack.json`. I've talked with with CI and they would like that we remove the stats.json file after we upload necessary info to the server.

packages/yoshi-common/src/webpack.config.ts
Gongreg: That would mean that we enable the stats writing before we are actually doing anything with it. Also CI wanted to have a gradual release if possible. (Both things can be done as an opt-out flag, but then we have to already create a flag in CI).

Gongreg:
> Today we have the `--stats` option. I prefer deprecating it as the stats writing mechanism here is preferable.

I would like to keep this option just for the ci if possible. Or we could reuse the implementation while still supporting `--stats`?

But maybe we should do it in a separate PR?

packages/yoshi-common/src/webpack.config.ts
tqmukas: What if there is no such variable? What happens then? The path would be `undefined`.

packages/yoshi-common/src/webpack.config.ts
ranyitz: Configuring a variable parameter in the CI (before releasing the feature in yoshi) makes sense to me in any case, and shouldn't be hard.

packages/yoshi-common/src/webpack.config.ts
ranyitz: Just to verify, we agree on it being a convention but we still discussing the path and file name.

> Since we don't want to clutter the artifact with unnecessary data

What do you mean clutter the artifact?

> I've talked with with CI and they would like that we remove the stats.json file after we upload necessary info to the server.

You mean, to not package it in the docker image? Is there a location we can put it which will not be included in the docker image?

ranyitz:
@Gongreg 

> I would like to keep this option just for the ci if possible. Or we could reuse the implementation while still supporting --stats?
> But maybe we should do it in a separate PR?

What do you mean "just for the ci"?
We want to always generate stats in the CI by default.

Maybe we can do the same thing we're doing with sourcemaps. We're always generating sourcemaps on CI, if you want to have it locally as well use the [--source-map option](https://wix.github.io/yoshi/docs/api/cli#--source-map)

The reason for this is that generating sourcemaps takes time, also writing stats. It is usually not needed locally and therefore we do this tradeoff, but sometime you want to debug it, and therefore there is a CLI option for it.

ranyitz:
Also, we'll need to think about thunderbolt usecase, they are currently generating stats and put them in `dist/statics`. This PR makes it so we'll generate duplicated stats.

@ronami What do you think?

packages/yoshi-common/src/webpack.config.ts
ranyitz: Actually yoshi [write the stats today](https://wix.github.io/yoshi/docs/guides/bundle-analyze) on `target/webpack-stats.json`

I don't see a good reason to break this API, for people that already uses `yoshi build --stats` in their CI build, this would not change a thing. Moving the file would be breaking for them.

Gongreg: Set to `target/webpack-stats.json`

Gongreg:
@ranyitz, for now we will ignore the 2nd part of the question. It doesn't involve the generation of stats in the target directory.

I've moved old stats generation to use the plugin.

ranyitz:
@Gongreg It seems that the `--stats` test is failing (in the integration suite) Can you have a look?

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -636,7 +636,7 @@ export function createBaseWebpackConfig({
       (inTeamCity && isProduction && !isDev && !disableStatsOutput)
         ? [
             new StatsWriterPlugin({
-              filename: STATS_FILE,
+              filename: path.join('../../', STATS_FILE),
```
ranyitz, line 639: Actually, it's not straight away clear.

I think that we should resolve it with `process.cwd()`.

AdirAmsalem:
Hi guys, I searched whether my `stats.json` is being stored somewhere out-of-the-box and found this PR. More context here: https://wix.slack.com/archives/CHRSK2K63/p1587051978091400

Just making sure - it's ended up being emitted to `target`, right? i.e. there's no way to access the stats files generated in CI?
(As opposed to https://github.com/wix/yoshi/pull/1811, but which is only applicable to thunderbolt/mono-repo flow and is irrelevant for me)

Thanks

ranyitz:
@AdirAmsalem That's correct. Currently, it's only open for specific projects. It's done by configuring [`DISABLE_WEBPACK_STATS_OUTPUT = true`](https://github.com/wix/yoshi/pull/2171/files#diff-1a4b00e421a5af8f4f5d2ffdb4f1e51bR60) in the CI for most projects.

@Gongreg Can you share some information regarding when this will be open for everyone?

Gongreg:
Hey. We are still working on the server part.

When we will have a working storage solution, we will start to open the yoshi part for everyone.

PR #2173, opened 2020-04-01
Fix BM `moduleParams` type breakage

amitdahan:
### 🔦 Summary
Mainly - `locale` is now an optional `string`.
Defaulting to `'en'` sounds like a good enough plan.


amitdahan:
Seems another template test joined the failing gang. Seems unrelated right @ranyitz ?

PR #2184, opened 2020-04-02
Prevent throwing errors for `publicData` getter when value is not loaded yet.

yavorsky:
### 🔦 Summary
This PR fixes our `PublicData` Provider to not throw errors when value is not loaded yet. It's pretty ok to rerender it after it's loaded with some value.

Currently if we are making something like `publicData.get('title')` and it wasn't loaded yet, we just throwing an error.

This PR propose to return `undefined` since we are going to renender it anyway when value will be loaded.

Second change related to timeout we are checking for `publidData` request was succeed. Sometimes when connection is not good, it's really doesn't fit 500ms requirements.


packages/yoshi-flow-editor-runtime/src/react/PublicDataProviderViewer.tsx
```diff
@@ -15,7 +15,7 @@ export class PublicDataProviderViewer extends React.Component<
     try {
       return this.props.data[scope][key];
     } catch (e) {
-      console.error(new Error(`key "${key}" wasn't found on public data:`));
```
yavorsky, line 18: Seems like it was typo :/

packages/yoshi-flow-editor-runtime/src/react/PublicDataProviderEditor.tsx
```diff
@@ -78,7 +79,7 @@ export class PublicDataProviderEditor extends React.Component<
   handleSetParam = (key: string, value: any) => {
     const { Wix } = this.props;
 
-    if (!this.state.ready || !this.state.data) {
```
yavorsky, line 81: We are going to have data as an object anyway to prevent excess checks for `null`.

packages/yoshi-flow-editor-runtime/src/react/PublicDataProviderEditor.tsx
```diff
@@ -68,8 +69,8 @@ export class PublicDataProviderEditor extends React.Component<
   }
 
   handleGetParam = (key: string) => {
-    if (!this.state.ready || !this.state.data) {
-      throw this.state.readyPromise;
+    if (!this.state.ready) {
+      return undefined;
```
yavorsky, line 73: We are going to return `undefined` instead of throwing an error, which is pretty ok for getter.

packages/yoshi-flow-editor-runtime/src/react/PublicDataProviderEditor.tsx
```diff
@@ -21,17 +21,18 @@ export class PublicDataProviderEditor extends React.Component<
...
     const { Wix } = this.props;
 
     const publicDataPromise = new Promise((resolve, reject) => {
       setTimeout(function() {
-        reject(new Error('timeout of 500ms reached on Wix.Data.Public.getAll'));
-      }, 500);
+        reject(
+          new Error('timeout of 1500ms reached on Wix.Data.Public.getAll'),
+        );
+      }, 1500);
```
yavorsky, line 35: sometimes 500 ms is not enough.

packages/yoshi-flow-editor-runtime/src/WidgetWrapper.tsx
```diff
@@ -23,7 +23,8 @@ interface IFrameworkProps {
 }
 
 // Bolt ssr environment doesn't contain window.
-const globalWix = typeof window !== 'undefined' ? window.Wix : undefined;
+const inBrowser = typeof window !== 'undefined';
```
yavorsky, line 26: This change is not related to PR general idea, just small clarification for this check.

packages/yoshi-flow-editor-runtime/src/react/PublicDataProviderEditor.tsx
ranyitz: I would put this timeout in a constant.

PR #2185, opened 2020-04-02
[editor-flow] Add BoB apps support and allow users override platform hooks.

yavorsky:
### 🔦 Summary
This PR is an initial implementation of the [BoB apps support RFC](https://github.com/wix/yoshi/issues/2146)

> I won't dive deep into the basic concept, since it's written above. It's also doesn't contain breaking changes for current users (actually 1 🥴).  

What's included here:

#### Update convention of controllers naming
Currently we are going to support 2 types of controllers:
- `editor.controller.ts`
- `viewer.controller.ts` (or`controller.ts`).

##### `editor.controller.ts`
In current PR we are just adding support for this type of controllers: not considering it still when building entry-point for editorScript, but adding it to be a part of model. We can use it in future, when we'll know exactly what we can generate here.

Second option why do we need it on current stage - components validation. We will throw an error when both `Widget.ts` (app is OOI) and `editor.controller.ts` (app is BoB) is not present.

##### `viewer.controller.ts` vs `controller.ts`
Currently for ooi apps we are going to support only single controller, which will be a part of viewerScript. So it's completely looks good to me to still support `controller.ts` files.

#### Consider only components that have `.component.json` file
This is an devex enhancement not fully related to BoB apps, but will help users create new components more easily w/o breaking current flow.
Currently, when you want to create new component it's pretty enough to create new directory in `components` directory. Then you need to register this widget in dev center and fill id, create Widget and/or controller and Settings.

But in practice, in most cases you don't want to break yoshi process by just adding a directory. You can continue working on you app untill you'll have and widgetId, which means this component is pretty ready to use and you can create `.component.json` with this id.

This will also make our model interface more strict (no components w/o ids, additional runtime conditions, etc).

#### Support `viewer.app.ts`. (Former `app.ts`/`init.app.ts`).
Here we are going to change the idea of this file. Currently, we just loading it before each `createController` call and pass some props from here. Seems like it's not the correct scenario. After discussion with @ranyitz we understood that this thing is being used only in controller.
Also the `export default` function here is not saying anything. So we decided to create lifecycle hooks that platform supports and add few more. Then we are going to override this hooks with our magic, but considering user's actions.

So currently `viewer.app.ts` is a file from the `src` that allow user to override:
- `initAppForPage` (platform). `export const initAppForPage` will fully pass it to the platform.
- `mapPlatformStateToAppData` (editor-flow). `export const mapPlatformStateToAppData` will except function with arguments provided by the platform that returns some useful for controller fields. The same `mapStateToProps` allow you to pass properties you need from the whole app to component, we will also allow to pass properties from the whole platform to controller as an `appData` field.
- We will add more fields to override when we see it's currently using by someone.

#### Widget.ts/Page.ts changes
- It's possible for now component to now have Widget.ts or Page.ts files, only if `editor.controller.ts` is present.
- Allow to have `index.ts`. Since we are going to not use `Page.ts` and it's better people to have common shape for all ooi components, I would like to allow create `index.ts` file as an alternaive for Widget.ts. I think for ooi apps, Widget.ts is a main part and represents the idea of component's directory. So using index.ts looks more clean to me. (Doesn't have strict opinion about it).


packages/yoshi-flow-editor/src/constants.ts
```diff
@@ -1 +1,11 @@
 export const CI_CONFIG = '.ci_config';
+export const WIDGET_FILENAME = ['index', 'Widget', 'Page'];
+export const VIEWER_CONTROLLER_FILENAME = ['viewer.controller', 'controller'];
+export const EDITOR_CONTROLLER_FILENAME = ['editor.controller'];
+export const SETTINGS_CONTROLLER_FILENAME = ['Settings'];
+export const COMPONENT_CONFIG_FILENAME = ['.component'];
+export const APPLICATION_CONFIG_FILENAME = ['.application'];
+export const VIEWER_APP_FILENAME = ['viewer.app', 'app'];
```
yavorsky, line 8: Filename is array, so it will be more flexible to have optional names w/o additional conditions

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -31,6 +48,7 @@ export interface AppConfig {
 }
 export interface ComponentConfig {
   id: string;
+  type?: ComponentType;
```
yavorsky, line 51: User can specify `widget` | `page` in config. Currently we are not using it, but we will in future.

packages/yoshi-flow-editor/src/model.ts
```diff
@@ -92,57 +120,86 @@ For more info, visit http://tiny.cc/dev-center-registration`);
...
+      /* If no config specified, we are going to completely ignore this component and
+      warn users to add it. */
+      if (!componentConfig) {
+        console.warn(`Seems like you didn't add "${formatPathsForLog(
+          COMPONENT_CONFIG_FILENAME,
+          'json',
+        )}" to some of you components: ${componentPathRelativeToRoot}.
+For more info, visit http://tiny.cc/dev-center-registration`);
+        // Ignore components w/o config file.
+        return processedModels;
```
yavorsky, line 148: We will warn user that some component doesn't contain .component.json, but on the same time won't break flow

packages/yoshi-flow-editor/src/wrappers/templates/CommonViewerScriptEntry.ts
```diff
@@ -40,15 +40,18 @@ const controllerConfigs = t<{
...
 export default t<Opts>`
-  import {createControllersWithDescriptors, initAppForPage as initAppForPageWrapper} from '${({
+  import {createControllersWithDescriptors, initAppForPageWrapper} from '${({
     viewerScriptWrapperPath,
   }) => viewerScriptWrapperPath}';
-
-  import userInitApp from '${({ initAppPath }) => initAppPath}';
   ${({ controllersMeta }) => importsForControllers({ controllersMeta })}
+  import * as viewerApp from '${({ viewerAppFileName }) => viewerAppFileName}';
+  const importedApp = viewerApp;
```
yavorsky, line 48: Ok, here is a trick.. webpack will throw an error when we are importing module using `import * as all` and then address it `all.someModule`.
image

packages/yoshi-flow-editor/src/constants.ts
ranyitz: 😍

packages/yoshi-flow-editor/src/wrappers/templates/CommonViewerScriptEntry.ts
ranyitz: What are we trying to achieve?

yavorsky: We want to import everything from user's script and wrap these modules in ours

PR #2187, opened 2020-04-04
Bump express to ~4.17.0 in server templates

amitdahan:
### 🔦 Summary
`@wix/wix-express-csrf` [started using an `express@>=4.17.0` feature](https://wix.slack.com/archives/C0C89GRRB/p1585817764395300) which is why server templates started failing.

This should fix it.

### 🗺️ Test Plan
Template tests


amitdahan:
Question: Should we pin `express` and similar deps to make sure we're also handling them in our bulk-bumps?

amitdahan:
Closed in favor of #2189

PR #2193, opened 2020-04-05
Move transpilation target to ES5 (e.g. support IE11)

ranyitz:
### 🔦 Summary
We still need to support `IE11` and therefore the `target` compiler option should be set to `ES5`


amitdahan:
So in fact, only `module: 'esnext'` was originally required for `esm`?

PR #2196, opened 2020-04-06
Yoshi Server: handle non json response

### LINKED ISSUE #2190: Yoshi Server Client - improve error message in case of html error response
Use case - a call to an api function returns an html error page. This will cause an error here:
https://github.com/wix/yoshi/blob/master/packages/yoshi-server-client/src/index.ts#L76

```
FetchError: invalid json response body at http://localhost:1601/_api_ reason: Unexpected token < in JSON at position 0
```

We should either:
- make sure there's no such case (of an html error page for api function)
- make sure not to try and `await res.json()` in this case

Reproduction:
Cause an error by using an old express version.

yanivefraim:
Fixes #2190 (see more details inside the issue)


packages/yoshi-server-client/src/index.ts
```diff
@@ -73,13 +73,26 @@ export default class implements HttpClient {
     });
 
     if (!res.ok) {
-      const error = await res.json();
+      if (res.headers.get('content-type')?.includes('application/json')) {
```
ronami, line 76: @yanivefraim and I discussed the possibility of later on adding a custom response header (`x-wix-yoshi-server`) to signal that the response came from Yoshi Server (as opposed to something else).

packages/yoshi-server-client/src/index.ts
```diff
@@ -73,13 +73,26 @@ export default class implements HttpClient {
     });
 
     if (!res.ok) {
```
ronami, line 75: I think the response can also be successful but not be from Yoshi Server (or be a valid JSON)

PR #2197, opened 2020-04-06
bump wnpm-ci to 8.0.121

netanelgilad:
### 🔦 Summary
bump wnpm-ci to 8.0.121


ranyitz:
This should enable the fix so `wnpm` won't bump in case the version contents are the same

PR #2199, opened 2020-04-06
Verify dependency tree for dependency conflicts

### LINKED ISSUE #650: Verify package tree for multiple versions of the same package
# 🙋 Feature Request

For some dependencies (Webpack, ESLint, Jest) having multiple versions installed are likely to break project setup with errors that are hard to understand. We can verify that there is only one version of each of these and fail with a clear error otherwise.

See [create-react-app's](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/utils/verifyPackageTree.js#L18) implementation as an example.

amitdahan:
### 🔦 Summary
Closes #650.
Currently only checks for `webpack`.
Can be disabled via `SKIP_PREFLIGHT_CHECKS=true`.

### 🗺️ Test Plan
Template tests will now fail when something changes (for example a foreign webpack appears indirectly in one of our usual deps).
I think think is a good thing.


packages/create-yoshi-app/templates/client/typescript/{%packagejson%}.json
```diff
@@ -39,7 +39,8 @@
     "typescript": "~3.7.5",
     "velocity": "~0.7.0",
     "yoshi": "^4.1.0",
-    "yoshi-style-dependencies": "^4.1.0"
+    "yoshi-style-dependencies": "^4.1.0",
+    "sled-test-runner": "1.0.434"
```
amitdahan, line 43: `sled-test-runner@1.0.434` brings `webpack@5` as a dep.
I'll revert once everything is green except Client TS template.

packages/yoshi-flow-legacy/src/cli.js
```diff
@@ -11,7 +11,7 @@ module.exports = async command => {
   const appDirectory = fs.realpathSync(process.cwd());
   const action = require(`./commands/${command}`);
 
-  await verifyDependencies();
+  await runPreflightChecks();
```
amitdahan, line 14: This will needlessly run `verifyNodeVersion` a second time for the legacy flow.
While technically still correct, I thought it'd have a negligible effect on performance, I thought this refactor was worth it, and it gave me the env var check for free for (my new check).
Does anyone object?

packages/create-yoshi-app/templates/client/typescript/{%packagejson%}.json
```diff
@@ -39,7 +39,8 @@
     "typescript": "~3.7.5",
     "velocity": "~0.7.0",
     "yoshi": "^4.1.0",
-    "yoshi-style-dependencies": "^4.1.0"
+    "yoshi-style-dependencies": "^4.1.0",
+    "webpack": "5.0.0-beta.14"
```
amitdahan, line 43: `webpack` was still not deduped to the top so this should force it and finally fail.

packages/yoshi-common/src/verify-dependency-tree.ts
```diff
@@ -0,0 +1,87 @@
+import { PackageJson } from 'type-fest';
+import chalk from 'chalk';
+
+type DepConflict = {
+  dep: string;
+  local: string;
+  root: string;
```
amitdahan, line 7:
```suggestion
  localVersion: string;
  rootVersion: string;
```

amitdahan:
OOI templates broke, PR is green otherwises.

PR #2202, opened 2020-04-07
Refactor Editor Flow to load Editor SDK dynamically and incapsulate it

yavorsky:
### 🔦 Summary
After discussion with @maoryosef2 we decided that for platform apps, the platform should pass the sdk version we need for current app.

##### How it's working currently
Currently, we are including `<script src="FIXED_SDK_VERSION" />` in vm file and rendering it for local experience as `localhost:3000:[settings|editor]/:widgetName` and for prod creating html file, for example `editor/button.html`.

Later in all our scripts we can access editor sdk via `window.Wix`.
This is bad: we can't load it dynamically according to some specific version and we can't mock it normally for tests, since mocks is conflicting with original sdk source.

##### How to access `Wix` now?
With this change, we just pass Wix as a property to Settings and Widget. Also, we are verifying it's not loaded before sdk ready (the same it works now), since a lot of users' initialization is based on statement that sdk is available on component mounting.

##### How we are deciding which sdk version to load?
Currently, latest sdk is not compatible with one most of ooi apps using now. (1.190.0).
So if we have `wix-sdk-version` query param passed by our platform part (via editor script), we are loading relevant version, If not - static version as today.

##### Wix SDK mocks loading changes.
This update also provide a bit better experience for loading mock for tests.
So currently, for `yarn start` we included `wix-mock` in the structure and loaded it.

**Develop scenario - no mock**.With this change we don't need to load it for cases when we overriding urls and sending requests from editor or viewer part to localhost. So we won't load mock for most of development cases.

**Test scenario - no original sdk**. And if request is going directly to localhost by user, we don't want to load original sdk and use only mocked version.

We can distinguish these cases by query params platform sends (for example `instance`).

Since this PR has small breaking change - if you are using window.Wix outside of the component, it will be `undefined` initially.
So I opened PR to challenges to move bi events to render props: https://github.com/wix-private/strategic-products/pull/3808


packages/yoshi-flow-editor/src/server/server.ts
```diff
@@ -46,12 +46,18 @@ app.get(
 
 app.use('/editor/:widgetName', (req, res) => {
   const { widgetName } = req.params;
-  res.send(renderVM(editorTemplate, { widgetName }));
+  const { instance } = req.query;
+  res.send(
+    renderVM(editorTemplate, { widgetName, usePrivateSDKMock: !instance }),
```
yavorsky, line 51: Only platform pass instance. So we can `usePrivateSDKMock` only in cases when we directly going to `localhost/settings/widget`, but load real sdk for loading app in editor environment.

packages/yoshi-flow-editor-runtime/src/utils.ts
```diff
@@ -10,5 +10,11 @@ export function objectPromiseAll(target: Record<string, any>) {
   }, Promise.resolve({}));
 }
 
+export const getQueryParams = (
+  search = window.location.search,
+): URLSearchParams => {
+  return new URLSearchParams(search);
```
yavorsky, line 16: Screen Shot 2020-04-07 at 11 36 56 AM

yavorsky: AFAIK - we don't support IE in editor

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
```diff
@@ -0,0 +1,100 @@
...
+        this.queryParams.get('wix-sdk-version') || DEFAULT_EDITOR_SDK_SRC;
+
+      loadSDK(editorSdkScriptSrc).then(() => {
+        this.editorSDK = window.editorSDK;
+        this.Wix = window.Wix;
+        this.setState({
+          isSDKLoaded: true,
+          Component: WrappedComponentCallback(this.Wix, this.editorSDK),
+        });
+        if (this.editorSDK) {
```
yavorsky, line 72: window.editorSDK is only present in latest versions.

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export type WrappedComponentCallback<P> = (
+  Wix: IWixStatic,
+  editorSDK: EditorSDK,
+) => React.ComponentType<P>;
+
+const loadSDK = (scriptSrc: string) => {
+  // We pass it when rendering app with usePrivateMock param
+  if (window.__USE_PRIVATE_SDK_MOCK__) {
+    return import('../wixPrivateMockPath');
```
yavorsky, line 34: Load mocked version isntead of original one.

packages/yoshi-flow-editor-runtime/src/react/PublicDataProvider.tsx
```diff
@@ -0,0 +1,16 @@
...
+  Wix: IWixStatic | null;
+  data: Record<string, any>;
+  children: React.Component | React.ReactElement;
+}) => {
+  const { Wix, data, children } = props;
+  if (Wix) {
+    return <PublicDataProviderEditor Wix={Wix} children={children} />;
+  }
+  return <PublicDataProviderViewer data={data} children={children} />;
+};
```
yavorsky, line 16: Instead of having a lot of check which Provider to load, I created new component which will handle it.

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
```diff
@@ -10,27 +11,39 @@ const EditorAppWrapper = (
...
+          widgetId: '',
+          getAllPublicData: true,
+        },
+        // overrides: {
+        //   fedopsAppName: CART_ICON_APP_NAME,
+        //   fedopsLogger,
+        //   locale: window.__LOCALE__,
+        //   siteBaseUrl: `${window.__APP_MODEL__.topology.baseDomain}/`,
+        //   platform: {baseUrls: {cartIconBaseUrl: window.__BASE_STATIC_URL__}},
+        // },
```
yavorsky, line 41: I didn't touch comments, since I have a hope it will be useful someday

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
```diff
@@ -10,27 +11,39 @@ const EditorAppWrapper = (
...
+      },
+    );
+  };
+
+  return withEditorSDK(WithViewerScript);
+};
```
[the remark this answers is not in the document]
maoryosef2: Also if `WithEditorSDK` is used to inject for both editor and viewer, maybe its better to call it `WithSDK`

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
```diff
@@ -0,0 +1,100 @@
...
+      const editorSdkScriptSrc =
+        this.queryParams.get('wix-sdk-version') || DEFAULT_EDITOR_SDK_SRC;
+
+      loadSDK(editorSdkScriptSrc).then(() => {
+        this.editorSDK = window.editorSDK;
+        this.Wix = window.Wix;
+        this.setState({
+          isSDKLoaded: true,
+          Component: WrappedComponentCallback(this.Wix, this.editorSDK),
+        });
```
maoryosef2, line 71: this could be a bit problematic, since the wrapped component will be loaded before the SDK finished its configuration (relevant for the new editorSDK)

i guess that this should be 
```suggestion
       if (this.editorSDK) {
         this.editorSDK.panel.onEvent(({ eventType, eventPayload }) => {
            if (eventType === 'startConfiguration') {
              this.setState({
                config: eventPayload,
                Component: WrappedComponentCallback(this.Wix, this.editorSDK)
              });
            }
       } else {
        this.setState({
          isSDKLoaded: true,
          Component: WrappedComponentCallback(this.Wix, this.editorSDK),
        });
        }
```

packages/yoshi-flow-editor-runtime/src/utils.ts
maoryosef2: in `EditorAppWrapper.tsx` you are wrapping also the viewer script, is this wrapped only for editor time?
if not, then you are using WithEditorSDK in viewer flow which means that you are trying to use URLSearchParams in the viewer - in this case, we do need to support IE

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
maoryosef2: this will only load a mock for the old SDK, 
where is this mock used?

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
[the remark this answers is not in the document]
yavorsky: > This is running for OOI widgets or also for BoB?
for BoB it's redundant to inject the SDK in viewer time

No, if we don't have `Widget.ts` in the component directory, this will be ignored for these components. So this is only relevant for ooi apps

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
yavorsky: LGTM!

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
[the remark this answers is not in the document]
yavorsky: But we are injecting it in editor environment, since widgets are being rendered here. Also the variable name is `editorSDK` 🙂

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
yavorsky: For e2e tests. When you going to `localhost:3000/settings/componentName` it's being rendered with mock.

packages/yoshi-flow-editor-runtime/src/utils.ts
yavorsky: Yeah, it's only for editor part.

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
maoryosef2: so this is only relevant for old sdk e2e
need to comeup with a way to e2e new sdk

packages/yoshi-flow-editor-runtime/src/utils.ts
yavorsky: EditorAppWrapper use WidgetWrapper, which is part of the viewer. But we are not doing anything for WidgetWrapper

packages/yoshi-flow-editor-runtime/src/react/withEditorSDK.tsx
ranyitz: It worth putting this comment in the code, above this `if` statement,

maoryosef2:
>Even though both editorSDK and wixSdk comes from the same script (only in latest versions)
This is incorrect, editorSDK and Wix sdk are 2 different SDKs and comes from 2 different places
(editorSDK.js & js-sdk.js)

in platform apps you are getting the editorSDK injected in editorReady,
but when you open a panel which is a seperate IFrame you need to import editorSDK

yavorsky:
> Why HOC and not Provider? I would prefer this to be like anything else, both for our code and for the users

I think that `Wix` is something we are using in any ooi app, so It's pretty ok to pass it via props for any Settings and Editor's Widget component.
With provider approach we will allow users to prevent loading additional scripts, but on the same time we need it for `ViewerScriptWrapper`, so we will load it in any case. And as a result we'll just force users to add additional HOCs (and hooks in future) to the project

> If the SDK already exists on the window, let's use it. If not let's use the one from the platform

We control window in all cases, so here we'll have 2 different flows - sync and async. As a result - people will use in some place `window.Wix`, in some `props.Wix`. It will create additional confusion.
If performance downgrade is present here, that's the different case and we can discuss it more. I think we need to check http/2 is present here to verify it's true.

> but when you open a panel which is a seperate IFrame you need to import editorSDK

@maoryosef2 Do we have cases when we have both platformSdk and WixSDK at the same time?

From what I understood, it has different names, but one is supposed to replace another? So basically it's the same sdk, which has different names, different APIs and was renamed.

Maybe solution with HOCs is good, but since we know in which cases which type of SDK to use, I think controlling it on our side and passing via props something that can be used in user's component w/o forcing to import some HOCs sounds better to me.
Discussion is opened 🙂

ranyitz:
> I think that Wix is something we are using in any ooi app, so It's pretty ok to pass it via props for any Settings and Editor's Widget component.
With provider approach we will allow users to prevent loading additional scripts, but on the same time we need it for ViewerScriptWrapper, so we will load it in any case. And as a result we'll just force users to add additional HOCs (and hooks in future) as an abstraction to access data, but not to control performance by reducing scripts to load

One of the reason to put this on the context is so they they'll be able to access it anywhere in their app without transferring it between components via props. I think that they will need to put it on the context themselves to enable this and I prefer to do it for them in advance as it more comfortable. Also with a hook we can give them the types without them needing to import and declare them.

e.g.
```
const wix = useWixSdk()
// The Wix variable is typed
```
We're using this approach for everything in their app and I think that if both approaches are equal than we should go with similarity of the API (with BM and other things that we get through the runtime)

maoryosef2:
>@maoryosef2 Do we have cases when we have both platformSdk and WixSDK at the same time?

Not that I am aware of

ranyitz:
> We control window in all cases, so here we'll have 2 different flows - sync and async

What do you mean? Why we can't have all sync (by understanding the wanted flow type (OOI/BOB) on build time?)

Maybe i'm missing the reasoning for the async flow.

> As a result - people will use in some place window.Wix, in some props.Wix.

I'm not sure, but in case they will, let's add a lint rule. I think that there is a case to have an eslint config for the `editor-flow`. It would be a good way to prevent things like that.

> It will create additional confusion.

If doing `useWixSdk` in order to access the typed `wixSdk` will be documented as the best practice and everybody will use it. Why would it confuse them?

> If performance downgrade is present here, that's the different case and we can discuss it more. I think we need to check http/2 is present here to verify it's true.

(maor)
> you will still need to import the script in runrime, so you wont have much performance gain

You right that it's not a massive performance gain, but if I have 2 solutions which gives me the same outcome and one of them has better performance I don't see why not to go with it.

> I think controlling it on our side and passing via props something that can be used in user's component w/o forcing to import some HOCs sounds better to me.

Regarding controlling it on our side, I totally agree.

Regarding forcing it to import some HOCs/Render Props/Hook. We can always enable both APIs if we believe it's the right thing.

(Both as props and on the context for getting it down the line)

We just need to remember that we create 2 ways to get the same thing, which is something that I prefer not to do. I realize that having HOC/Render Props is not so nice, but IMHO a hook looks better, I believe that soon enough we'll be able to use `hooks` in the viewer and the syntax will improve.

yavorsky:
> What do you mean? Why we can't have all sync (by understanding the wanted flow type (OOI/BOB) on build time?)

B/c on production we just exporting `.html` file generated on build time. It's easy to generate it for OOI, since the SDK version is static. But for BoB apps we are getting it from query params.
Lile `editor/widget.html?editorSDKVersion=some-version`. In this case SSR will help and implementing server will solve it, but on the same time will create more difficulties with supporting it. And will probably block BoB progress.

> I'm not sure, but in case they will, let's add a lint rule. I think that there is a case to have an eslint config for the editor-flow. It would be a good way to prevent things like that.

But if we can remove it from `window.Wix`, which I think everyone agree is a bad practice, why should we support it at all? That's another reason why I don't like idea to load it via `<script>` in html.

> You right that it's not a massive performance gain, but if I have 2 solutions which gives me the same outcome and one of them has better performance I don't see why not to go with it.

B/c 2 solutions = `complexity*2`. We'll need to support both solutions and create 2 different flows for it. People could face new issues during migration from ooi to bob. For example - we are saying that `Settings.ts` looks pretty similar between BoB and OOI, but on the same time, for OOI you have `window.Wix` in global scope, in BoB - not. For BoB - SDKProvider will fetch SDK, but for OOI - not. Maybe it's details and we shouldn't spend some time on it. I'm trying to find similarities between this types of project and provide common approach - that's why we'll have viewerScript processing the same way for both apps. It's not easy, but creating new branch of logic every time we can do it, will finally result in 2 different flows branching on the entry point and having nothing in common, just covered by single one.


I agree with points related to Context.Consumer. I think we should generate Provider by ourselves and use it, but allow users to use consumer to get SDK instance.
Hook examples looks really nice, but I really not sure when it will be supported by Bolt and won't rely on it.

Updating PR to support Consumers + RenderProps. Let's discuss more importing sdk via `<script>` in ooi html.

packages/yoshi-flow-editor-runtime/src/react/SDK/SDKRenderProp.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+export const EditorSDK: React.FC<IEditorSDKConsumer> = props => {
+  const { children } = props;
+
+  // We don't have Editor SDK for viewer part and going to return `{ editorSDK: null, editorSDKConfig: null }`
+  if (!isEditor()) {
+    return children(defaultEditorSDKContext);
+  }
```
maoryosef2, line 96: when will this be rendered outside of the editor?

packages/yoshi-flow-editor-runtime/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_EDITOR_SDK_SRC =
```
maoryosef2, line 1: this is the old SDK,
please call it `DEFAULT_JS_SDK_SRC` or `DEFAULT_WIX_SDK_SRC`

packages/yoshi-flow-editor-runtime/src/utils.ts
```diff
@@ -10,5 +10,27 @@ export function objectPromiseAll(target: Record<string, any>) {
...
+export const loadScript = (scriptSrc: string) => {
+  return new Promise(resolve => {
+    const script = document.createElement('script');
+    script.src = scriptSrc;
+    script.async = false;
+    script.onload = resolve;
+
+    document.body.appendChild(script);
+  });
+};
```
maoryosef2, line 31:
```suggestion
export const loadScript = (scriptSrc: string) => 
  new Promise(resolve => {
    const script = document.createElement('script');
    script.src = scriptSrc;
    script.async = false;
    script.onload = resolve;

    document.body.appendChild(script);
  );
};
```

packages/yoshi-flow-editor-runtime/src/react/SDK/SDKRenderProp.tsx
yavorsky: For OOI apps. Here we have `Widget` React component, which looks the same for user, but being rendered in both viewer and editor. By using our Render Prop, user should understand that it could be null for viewer case, and I configured types to be `sdk | null` for it for Widget

maoryosef2: Though its the editorSDK wrapper, which OOI probably doesn't use, so I am not sure that it is needed

yavorsky: Yeah, removed it foe EditorSDK render prop

PR #2216, opened 2020-04-12
Generators: update the usage of `wix-test-env` to better support experiments setup

yanivefraim:
We want to be able to mock experiments inside our `index-dev` file, **before** starting the server. 
For this, we have to use the `withConfigurer`:

Environment file:

```
TestEnv.builder()
...
.withConfigurer(({ petri }) => {
    petri.onAnyConducts(() => ({
      'specs.crash.IsTitleGilad': 'true',
    }));
  })
...
.build();
```

The problem is that we cannot do this inside the environment file, because it will affect tests as well.

Solution:

Export `TestEnv` without running `.build()`. In this case, we can run the `withConfigurer` and then `build()` inside `index-dev`.

Full example will be added to fed-handbook.


yanivefraim:
Closing, for now

PR #2217, opened 2020-04-12
separate wrapping for ooi widgets and platform widgets

maoryosef2:
### 🔦 Summary
Currently all widgets are being wrapped with a `setProps` helper function, which is injected as the first parameter to `pageReady`.
But `setProps` is only relevant for OOI widgets, this change includes a couple of things

- A more explicit definition of the widget type (by adding `ooi_widget` and `platform_widget`)
- Wrapping the controller method according to the widget type
  - for `ooi_widget` wrap with the `setProps` and state handler
  - for the rest wrap with a "default" wrapper, which simply invokes the controller method

@yavorsky I am working on top of your branch instead of master as it is includes other changes required for me to work locally on this change.

UPDATE: changed to work against master now that https://github.com/wix/yoshi/pull/2202 is merged

### 🗺️ Test Plan
- will add tests for `platform_widget` types and see that their template is generated correctly


packages/yoshi-flow-editor/src/model.ts
```diff
@@ -170,8 +172,13 @@ For more info, visit http://tiny.cc/dev-center-registration`);
       const editorControllerFileName = resolveFromComponents(
         EDITOR_CONTROLLER_FILENAME,
       );
-      const widgetFileName = resolveFromComponents(WIDGET_FILENAME);
-      if (!widgetFileName && !editorControllerFileName) {
+
+      const isOOIWidget = componentConfig.type === OOI_WIDGET_COMPONENT_TYPE;
```
yavorsky, line 176: I think we can accept this confuguration as an override option, bit on the same time we can pretty easy automatically  set which type of component is in use.

I would switch approach here to:
- Check if `Widget.ts` present with `resolveFromComponents`.
- If nothing found, that's platform app. If found - OOI.

Since we already asking to follow some conventions and reducing configuration mental load, I would use option we can definitely know by default.

packages/yoshi-flow-editor/src/wrappers/commonViewerScriptWrapping.ts
```diff
@@ -14,6 +14,7 @@ const toControllerMeta = (
   return {
     controllerFileName: component.viewerControllerFileName,
     id: component.id,
+    type: component.type,
```
yavorsky, line 17: Since we'll have 2 kind of controllers - `viewer.controller.ts` and `editor.controller.ts`, I think it could be a bit confusing to have `ooi` and `platform` inside type field. Maybe it's worth to rename it to `componentType` or something more strict?

maoryosef2: we can go with `widgetType`

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -3,9 +3,12 @@ import { createInstances, objectPromiseAll, fetchFrameworkData } from './utils';
 
 let frameworkData: any;
 
+type WidgetType = 'ooi_widget' | 'platform_widget';
```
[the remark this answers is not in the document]
maoryosef2: yes... intended to do it (see the TODO on line 7)

packages/yoshi-flow-editor/src/model.ts
maoryosef2: I agree with the approach of less configuration, 
the reason I made this change was this line 
https://github.com/wix/yoshi/pull/2217/files/d14308c31571a0df464c8fe4db17e39a3a04593d#diff-17fb83743fd8618675a37595513fa758L174

for someone not familiar with the flow, it will be hard to understand the connection between 
`widgetFileName` and `editorControllerFileName`
So I thought we can make it more explicit.

I can convert it to something like this

remove the default component type assignment
```javascript
      if (!componentConfig.type) {
        componentConfig.type = DEFAULT_COMPONENT_TYPE;
      }
```
if we have a `widgetFileName` set component type to `OOI_WIDGET_COMPONENT`
if we have an `editorControllerFileName` set component type to `PLATFORM_WIDGET_COMPONENT`

if we have both / none throw the appropriate error

maoryosef2:
> For example when we are going to add translations or experiments, maybe we'll have a lot of things in common and we have to just maintain 2 different flows with a lot of things in common.

where would those be added?
currently the first parameter passed is `setState`, so in any case you will need to make breaking changes in order to support that flow, we can consider making the first argument an object that will contain context APIs (and for OOI widgets it will have `setState`)

packages/yoshi-flow-editor/src/model.ts
maoryosef2: changed to assume the widget type if it wasn't given

yavorsky:
> > For example when we are going to add translations or experiments, maybe we'll have a lot of things in common and we have to just maintain 2 different flows with a lot of things in common.
> 
> 
> 
> where would those be added?
> 
> currently the first parameter passed is `setState`, so in any case you will need to make breaking changes in order to support that flow, we can consider making the first argument an object that will contain context APIs (and for OOI widgets it will have `setState`)

It’s not first parameter. It’s context, so you can access it via this.

maoryosef2:
> It’s not first parameter. We are using .call, so it’s context. you can access it via this.

wasn't talking about the `.call`, I was referring to the wrapping of `pageReady`
https://github.com/wix/yoshi/blob/c6965d15ae1cabdfc8e15c8c4727dd44a809610d/packages/yoshi-flow-editor-runtime/src/viewerScript.ts#L87

passing `setProps` as the first argument there, its redundant for platform widgets, and actually caused an issue since the script expected `$w` as the first argument.

yavorsky:
@maoryosef2 ah, I got it.. ok, I think we can consider extending or merging it in future when we really need.. for now it's looking really great

Issue #2213, opened 2020-04-12
[yoshi-server] can't use wix-api-gw-full-testkit with yoshi test --mocha


morsh:
# 🐛 Bug Report

### 😯 Current Behavior
When running tests that use `@wix/wix-api-gw-full-testkit` all tests fail because the testkit can't start.
Slack thread: https://wix.slack.com/archives/C0NJPS8N5/p1586681815102500
Error: Timeout exceeded 30000 ms

### 🤔 Expected Behavior

Tests should run.

### 💁 Possible Solution 
Enable increase timeout for mocha tests, which is currently hardcoded:
https://github.com/wix/yoshi/blob/master/packages/yoshi-flow-legacy/src/commands/test.js#L108

### 💻 Code Sample

`environment.ts`:
```ts
import * as testkit from '@wix/wix-bootstrap-testkit';
import * as ApiGwTestkit from '@wix/wix-api-gw-full-testkit';

export class Environment {
  public apiGwServer: ApiGwTestkit.ApiGwClientTestkit;

  private bootstrapServer() {
    return testkit.app('./index', {
      env: {
        MOUNT_POINT: '/search-everything',
        DEBUG: 'wix:*'
      }
    });
  }

  constructor() {
    this.apiGwServer = ApiGwTestkit.createTestkit();
  }

  async start() {
    this.app = this.bootstrapServer();
    await this.apiGwServer.start();
    await this.app.start();
  }

  async stop() {
    await this.apiGwServer.stop();
    await this.app.stop();
  }
}
```

### 🌍 Your Environment

```
  System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: Not Found
    npm: 6.14.4 - ~/.nvm/versions/node/v10.15.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 80.0.3987.163
    Firefox: Not Found
    Safari: 13.1
  npmPackages:
    @storybook/core: Not Found
    @storybook/react: Not Found
    babel-preset-yoshi:  4.51.1 
    eslint-config-yoshi:  4.53.0 
    eslint-config-yoshi-base:  4.53.0 
    jest-yoshi-preset: Not Found
    storybook: Not Found
[...]

PR #2218, opened 2020-04-12
Enable setting mocha timeout as an environment variable

morsh:
Resolving issue: https://github.com/wix/yoshi/issues/2213
FYI: @ranyitz


packages/yoshi-flow-legacy/src/commands/test.js
```diff
@@ -100,12 +100,14 @@ module.exports = runner.command(
         specsPattern.push(...globby.sync(globs.e2eTests));
         await bootstrapCdn();
       }
+      
+      const mochaTimeout = process.env.MOCHA_TIMEOUT || 30000;
```
ranyitz, line 104:
```suggestion
      const mochaTimeout = process.env.MOCHA_TIMEOUT || '30000';
```

That will probably work, but just to go on the safe/typed side.

morsh:
@ranyitz - updated, let me know if this wording works for you

docs/api/cli.md
```diff
@@ -153,6 +153,8 @@ Run tests with Jest - this is the default
 
 Run unit tests with Mocha
 
+You can set environment variable `MOCHA_TIMEOUT` to set the default timeout for mocha tests.
```
ranyitz, line 156:
```suggestion
You can set environment variable `MOCHA_TIMEOUT` to set the timeout for mocha tests.
defaults to 30000ms
```

PR #2219, opened 2020-04-13
`BROWSER=none` instead of `false` for canceling browser from opening on `start`

ranyitz:
### 🔦 Summary
There is a mistake, `openBrowser` underlying function from `react-dev-utils` is [looking for `BROWSER=none`](https://github.com/facebook/create-react-app/blob/master/packages/react-dev-utils/openBrowser.js#L38)


yavorsky:
I love this direction
image

PR #2236, opened 2020-04-14
Fix lodash path for worker externals

yavorsky:
### 🔦 Summary
There are many ways to use lodash:

```
import lodash from 'lodash'
import { get } from 'lodash'
import get from 'lodash/get'
```

See Sokra's comment: https://github.com/webpack/webpack/issues/10111#issuecomment-565784249

In most cases we know that `lodash` already present in this environment.
But currently when you specify lodash in externals, it will work only for direct import: `from 'lodash'`. But a lot of people are using lodash in a cherry-pick way trying to reduce bundle size. 

But in fact they only increasing it, since externals mechanism we use isn't including importing of methods: `from 'lodash/get'`.

It's related not only to editor flow, but since we are controlling externals on our side, this PR related only to flow editor. Other flows seems to have configured externals on user's side. (Maybe let's consider updating generators)

This is an example of `viewerScript.bundle.js`:

### `import lodash from 'lodash'`

Screen Shot 2020-04-14 at 2 13 22 PM

### `import get from 'lodash/get'`

Screen Shot 2020-04-14 at 7 56 46 PM

This PR use function to match externals. Since we already know it's available via `_` in the worker flow, we can use it to match lodash object with methods we need.


yavorsky:
With this change we reduced bundle size for prod app (challenges) that use our flow

before:
Screen Shot 2020-04-16 at 7 28 19 PM

after:
Screen Shot 2020-04-16 at 7 28 23 PM

Thanks @ranyitz for pointing me to `externalizeRelativeLodash` and @roymiloh for implementing it 😃

PR #2237, opened 2020-04-14
Add flow editor entries to files needed to be transpiled.

yavorsky:
### 🔦 Summary
Follow-up for https://github.com/wix/yoshi/pull/2229 with better solution.
Currently we have wrapper files for editor flow in specific directory inside yoshi. It's pretty good for now, except the fact that these files are not being transpiled.

I can see that we have some exclusions for some projects like wix-style-react and editor-elements, so I really want `yoshi-flow-editor` to join this **VIP club** of dependencies. 👑💵🚁


tuchk4:
image

bodia-uz:
Pls ensure wix-style-react, editor-elements, and other VIP projects are OK with yoshi-editor-flow company. Not sure if it’s opened club

yavorsky:
@bodia-uz flow-editor is a Kanye West in the world of wix packages, so there shouldn't be any issues. I think we need to add it to the docs.

We can write some e2e tests for having `editor-elements` and `wix-style-react` in the same room tho. Wanna open PR? You can use Stanford prison experiment testkit for it.

ranyitz:
Stanford prison experiment testkit 😆

PR #2241, opened 2020-04-15
Re-render module entry on model changes

amitdahan:
### 🔦 Summary
Re-render our secret `yoshi-flow-bm/tmp/module.ts` file on model changes (i.e new page file).

This will still not make `npm start` JustWork™️ since right now - BM environment will need to be restarted with a new BM testkit, and BM testkit is still configured in the user's project, not through `flow-bm` just yet.

This will only generate new bundles for model changes.

To complete this we'll have to provide "our own" BM testkit for `npm start` experience.


packages/yoshi-flow-bm/package.json
```diff
@@ -9,6 +9,7 @@
   "dependencies": {
     "arg": "4.1.3",
     "chalk": "3.0.0",
+    "chokidar": "3.3.1",
```
yavorsky, line 12: Don't we want to give a chance to https://github.com/parcel-bundler/watcher ? :)

yavorsky, line 12: Don't we want to give a chance to https://github.com/parcel-bundler/watcher ? :)

packages/yoshi-flow-bm/src/renderModule.ts
```diff
@@ -49,12 +49,9 @@ createModule({
   config: ${JSON.stringify(moduleConfig)},
 });`;
 
-const renderModule = (model: FlowBMModel) => {
-  const filePath = path.resolve(__dirname, '../tmp/module.ts');
+export const moduleEntryPath = path.resolve(__dirname, '../tmp/module.ts');
```
yavorsky, line 52: I've moved it from `tmp` to hidden dir: `.custom-entries`.  https://github.com/wix/yoshi/pull/2237/files#diff-223545c43cd2f58df056ab79210c6f63R11

Maybe it's worth to move here also to some hidden by default directory

packages/yoshi-flow-bm/package.json
yavorsky: And for chokidar - why the version is pinned?

ranyitz: I tried to use it in the past (parcel-bundle/watcher) and it had many bugs. I'm ok with starting with chokidar, in case we'll prefer parcel-bunbler/watcher we can always change that across the board.

packages/yoshi-flow-bm/src/renderModule.ts
ranyitz: Agree, It needs to be done, we can do this in a dedicated PR.

packages/yoshi-flow-bm/package.json
amitdahan: I thought it'd be better to pin everything, isn't it?

PR #2244, opened 2020-04-16
Import from yoshi-helpers/build

amitdahan:
### 🔦 Summary
...

### 🗺️ Test Plan
...

### DIFF
--- a/packages/yoshi-flow-editor/src/model.ts
+++ b/packages/yoshi-flow-editor/src/model.ts
@@ -1,6 +1,6 @@
 import path from 'path';
 import globby from 'globby';
-import { getProjectArtifactId } from 'yoshi-helpers/utils';
+import { getProjectArtifactId } from 'yoshi-helpers/build/utils';
 import { isTypescriptProject } from 'yoshi-helpers/build/queries';
 import resolve from 'resolve';
 import fs from 'fs-extra';


amitdahan:
Hey @ranyitz & @yavorsky 
Do you think we should remove the `yoshi-helpers/utils` entry-point?
I can continue in this PR if so.

ranyitz:
@amitdahan I agree that we should remove it, because having it there, make it feels like a public API.

Actually, other people thought about it as well

https://github.com/swimmio/wix-style-react/blob/52d2e9a9b1a9c9d688f2254eae9d7f3909d7a324/scripts/bundlify-components/webpack.config.js#L5
https://github.com/wix-platform/wix-serverless/blob/0f59eba67984ae83e4ab14f25deb76f24dd2ec76/serverless-runtime-server/mocha/it.conf.js#L1
https://github.com/wix-private/app-market/blob/ccecaed727f94deb2722ee5a6f47fcc972630241/badges/members-badge-definitions-bm-client/test/transforms/babel.js#L2

They just need to switch it to `yoshi-helper/build/utils` and understand that they are using a private API which will probably break in the future.

We know that we even want to remove `yoshi-helper` package. So that is a good start.

I think it totally makes sense to do so, we just need to open PRs to those projects so they won't suddenly break.

PR #2245, opened 2020-04-16
Use `is-ci` package instead of reading TeamCity's env vars

stavalfi:
it was pretty interesting task.

some notes:

1. https://github.com/watson/is-ci is using https://github.com/watson/ci-info. it checks if `BUILD_NUMBER` is truth to see if we are running on teamcity. but if not, it also checks the env-vars under this file: `https://github.com/watson/ci-info/blob/master/vendors.json` which is: `TEAMCITY_VERSION `.

2. this pr was a bit of a cleanup-pr to make sure yoshi is using the correct env-vars in tests (`test-helpers/env-variables.js`) so the use of `is-ci` will be safe to use in teamcity (or any other ci)

3. `IS_BUILD_AGENT` was removed from the code base because `is-ci` does not support it. I replaced it with `TEAMCITY_VERSION`

4. `packages/yoshi-flow-legacy/test/e2e.spec.js` - I replace `outsideTeamcity` with insideTeamcity in some tests because those test fail in https://pullrequest-tc.dev.wixpress.com/ because in the production code we check if we run on CI using `isCi`, and before this pr, we checked it using `IS_BUILD_AGENT` but didn't pass it in the tests so teamcity override it and thats why the tests passed until now. in short, these tests only worked when they run in CI or locally but with specific chrome version. (Thanks to @netanelgilad for helping me out understand why those tests failed!!)


ohana54:
Can you change this as well:
https://github.com/wix/yoshi/blob/151baa06519be1f5bfc47f244cf6d8c2a0e0ad77/packages/yoshi-helpers/src/queries.ts#L27

stavalfi:
discussion about failing tests: https://wix.slack.com/archives/CAL591CDV/p1587043602204700

stavalfi:
> Can you change this as well:
> https://github.com/wix/yoshi/blob/151baa06519be1f5bfc47f244cf6d8c2a0e0ad77/packages/yoshi-helpers/src/queries.ts#L27

i changed the implementation of this function, but this function has a use case. yoshi added specific teamcity test-reporter when it knows that its running on teamcity.

ranyitz:
@stavalfi Good catch. We do need a separate `isCi` and `isTeamCity` function. Just because teamcity requires specific logger etc.

ranyitz:
@stavalfi Let me know when it's ready for review

stavalfi:
i will. thanks for all the help @ranyitz

ranyitz:
@stavalfi Nice work on the PR! I feel bad that the legacy testing infra is so hard to work with. What's left here, can I help with something?

stavalfi:
@ranyitz thank you so much for trying to help! 2 tests are left. let's take it private

PR #2250, opened 2020-04-16
Yoshi server generators: migrate main route

yanivefraim:
WIP


packages/create-yoshi-app/templates/fullstack/typescript/src/init-server.ts
```diff
@@ -0,0 +1,3 @@
+export default async context => {
```
yanivefraim, line 1: we will move the config to be part of the `context` (almost every project will need it)

PR #2252, opened 2020-04-16
Pass `componentId` to platform's Settings

yavorsky:
### 🔦 Summary
Actually we need to consider not only `editorSDKSrc` query, but also `componentId`, which is a part of url. We will use this value in platform `Settings`.


yavorsky:
@maoryosef2 Sounds good! We've also discussed it with @ranyitz and decided that passing some stuff via props has some cons:
- It's not typed for users. They should check what each props is and type it for themself.
- It's harder to use it in nested components (need to pass props deeper)
- It's not scalable. In future, we'll just add new props to the root, which could affect on extra re-render or causing dirty code on user's side.

It's also redundant to use render props + context as for me, since we don't have much props yet (only `componentId` is useful).

So we've just added util helper `getEditorParams` that will help users to get a value. It's typed and has extra check to prevent use it in viewer environment.

Settings.ts

```
componentDidMount() {
  const { componentId } = getEditorParams();
  // Do smth with `componentId`
}
```

WDYT?

maoryosef2:
Looks really good

PR #2256, opened 2020-04-19
Bundle editor-elements-library similar to thudnerbolt-elements

NitayRabi:
### 🔦 Summary

- editor-elements-registry is an NPM module and should not be bundled
- editor-elements-library should be the GA'able artifact and be bundled the same way as thunderbolt-elements

CC: @tuchk4


NitayRabi:
Related to https://github.com/wix-private/editor-elements/issues/336

PR #2257, opened 2020-04-19
Add an option to use `@wix/wix-bootstrap-renderer`

### LINKED ISSUE #2255: Yoshi Server: add an option to use `@wix/wix-bootstrap-renderer`
We want our users to be able to render their `ejs` file without our `render` method, which is currently coupled to the `htmlWebpackPlugin`.

yanivefraim:
Fixes #2255 (see more details inside the issue)


packages/yoshi-server/src/server.ts
```diff
@@ -76,6 +76,9 @@ export default class Server {
           await new Promise(resolve => requireHttps(req, res, resolve));
           await new Promise(resolve => cookieParser()(req, res, resolve));
           await new Promise(resolve => wixExpressCsrf()(req, res, resolve));
+          await new Promise(resolve =>
```
yanivefraim, line 79: I think we should refactor those promises, maybe in a later PR

PR #2266, opened 2020-04-19
Change exports to export s to make the search work on API page

ranyitz:
### 🔦 Summary
For some time now, our search doesn't work when trying to search something from the API tab. For some really strange reason it seems that the word `exports` as an h2 tag, cause this to happen.

This is a horrible workaround, but I couldn't think on anything better.

I opened an issue to docusaurus and hopefully it'll get fixed and we'll be able to clean this.

https://github.com/facebook/docusaurus/issues/2629


amitdahan:
Weird, why did the deploy failed?

ranyitz:
@amitdahan I don't know, it's flaky and fails sometimes.

PR #2267, opened 2020-04-20
Add a progress bar for all build & start commands

### LINKED ISSUE #1987: Show a progress bar for webpack's compilation process
# 🙋 Feature Request

### 😯 Current Behavior

When building there is no feedback to the user, this process can take around 5 minutes without providing a feedback to the user.

### 🤔 Expected Behavior

Yoshi will show some indication to progress of the bundling process.

### 💁 Possible Solution

Show a progress bar!

Here is some plugin - https://github.com/nuxt/webpackbar
Here is another - https://webpack.js.org/plugins/progress-plugin/

### 🔦 Context

It can happen (even as a bug) that a build takes a lot of time, while we thought about creating a time based estimation (#1832 ) for our commands, this is much easier to do.

ranyitz:
### 🔦 Summary
This PR fixes #1987 

It adds a progress bar that shows the progress of every compiler Yoshi use in its Webpack compilation.

This PR will affect all of the build & start commands of all flows.

* In monorepos this would also add the package name.
* There are distinctive colors for each compilation type.
* It does not run on CI, so we won't see progress there. That could be another feature, but it's not in the scope of this one.
* Do not use the progress bar on `legacy flow` and `library flow` where the progress bar doesn't integrate well with the current output. I don't think that we'll add new features to the legacy flow. Library flow needs to be verified before we can enable it.

### Fullstack (Build)
fullstack-progress

### Out Of Iframe (Build)
ooi-progress

### Thunderbolt (monorepo build)
thunderbolt-progress

### Out Of Iframe (Start)
ooi-start-progress


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -71,6 +72,47 @@ function prependNameWith(filename: string, prefix: string) {
...
+      return { name: `Server [production]`, color: 'orange' };
+    }
+  }
+
+  const obj = getObject();
+
+  obj.name = obj.name.padEnd(longestNameLength);
+
+  if (isMonorepo) {
+    obj.name = `${stripOrganization(packageName)}\n  ` + obj.name;
```
amitdahan, line 110: Might as well 
```suggestion
    obj.name = `${stripOrganization(packageName)}\n  ${obj.name}`;
```

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -670,6 +714,14 @@ export function createBaseWebpackConfig({
             ]
           : []
         : []),
+
+      ...(isProduction && !inTeamCity
+        ? [
+            new WebpackBar({
+              ...getProgressBarInfo(configName, isDev, isMonorepo, name),
+            }),
```
amitdahan, line 722: Should work the same, no?

```suggestion
            new WebpackBar(
              getProgressBarInfo(configName, isDev, isMonorepo, name),
            ),
```

PR #2272, opened 2020-04-21
Generate automatically bundle analyzer port

yavorsky:
### 🔦 Summary
Currently we have issues related to multiple bundle analyzer processes. If you want to compare different projects, different versions of the project, branches, etc you can't start 2 bundle analyzer processes on the same time:
Screen Shot 2020-04-21 at 11 55 11 AM
image

Moreover if you have some complex projects including 2-3 webpack processes (worker, client, server) only first built project will generated for analyzing. Other ones will just throw the same error.

Fortunately webpack-bundle-analyzer [allows auto generated ports via `auto` value](https://github.com/webpack-contrib/webpack-bundle-analyzer/blob/acf2d1cfa86139689fa6e936acb7ce71c110503e/src/BundleAnalyzerPlugin.js#L25)

This PR supposes no one (except our tests) actually rely on 8888 port for bundle analyzer and doesn't care what port it will be if we auto-open the browser page.

We can use here more complex migration strategy: allow `--analyze auto` and `--analyze port`, use `--analyze` as a current `8888` value. But I really don't see advantages of it based on my assumption that no one really care about `8888` port. 
That's why I moved port number configuration to be env variable instead of CLI value.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -384,6 +388,11 @@ export function createBaseWebpackConfig({
     ...nodeExternalsWhitelist,
   ];
 
+  const inBrowser = process.env.BROWSER !== 'none';
```
ranyitz, line 391: I think it's `isBrowser`, right?

packages/yoshi-flow-legacy/test/build.spec.js
```diff
@@ -895,7 +895,9 @@ describe('Aggregator: Build', () => {
           'src/client.js': '',
           'package.json': fx.packageJson(),
         })
-        .spawn('build', ['--analyze']);
+        .spawn('build', ['--analyze'], {
+          ANALYZE_PORT: '8888',
```
ranyitz, line 899:
```suggestion
          ANALYZE_PORT: analyzerServerPort,
```

PR #2274, opened 2020-04-21
Minimize node server code for Tb-Elements

ronami:
### 🔦 Summary

Currently, TB downloads the SSR code for TB Elements. Since this file starts getting larger, this is an easy way of making it significantly slower, even though it may get larger in the future.

I don't think that this is something other Node servers should use since most of them are very small (`node_modules` are external) and it may make debugging (or understanding stack traces) harder.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -501,7 +503,7 @@ export function createBaseWebpackConfig({
             // Do not modify/set the value of `process.env.NODE_ENV`
             nodeEnv: false,
             // Faster build time and possibly easier debugging
```
ranyitz, line 505: It's not critical, but the comment above now makes less sense when this is a variable and not a `false` value.

```js
// Faster build time and possibly easier debugging
```

I think that `forceMinimizeServer` would be a bit better here in communicating that (talking about variable name) and that the comment should be updated a bit.

Issue #705, opened 2018-10-31
Notes regarding improving Guides in yoshi documentation


ranyitz:
### Improve Guides in Yoshi
> Write your name if you're taking the task.
- [ ] Remove AB translate. (Maybe move to fed handbook)
- [ ] Change assets guide (public ? Remove completely) - A really good reference - https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#using-the-public-folder
- [ ] Change disable-CSS-modules to working with CSS and talk about all the gotchas
- [ ] Enzyme support - we should change the section to running tests and enzyme should be a section inside of component testing using jsdom.  (See https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#running-tests) ( @yurynix )
- [ ] export ES modules along with commonjS - Work on the phrasing
- [ ] Moment.js optimization (maybe put In troubleshooting section, as it’s not important to understand unless you have a problem) - https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#momentjs-locales-are-missing 
- [ ] Split chunks - We should talk about code splitting in general, speak about how to do dynamic import and split chunks.
- [ ] How to use SVG - adding SVGs
- [ ] Add Displaying Lint Output in the Editor section (https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#displaying-lint-output-in-the-editor )
- [ ] Adding-images-fonts-and-files -  we should write about that. https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-images-fonts-and-files
- [ ] Move all of jest-preset-yoshi readme into the website. ( @yurynix )

PR #2276, opened 2020-04-22
Brand new website

ranyitz:
### 🔦 Summary
Our website is old and does not reflect Yoshi as a product anymore. This PR is a new face lift to the site itself which includes rearrange and addition of relevant content.

### New Look
1. Migration to docusaurus 2 (alpha, but works well!)
2. Include dark mode
3. Verify all current features are working (versioning etc)

### Re-Arranged to Communicate what we do
1. Documentation is now separated by flows
2. Added a Welcome Page that explains about Yoshi in general

### Want to look at it before merge?
https://wix-my-website-pr-2276.surge.sh/

### Degradation
1. Search is not working for v3.x (Until docusaurus v2 will add this feature)

### Notes
1. I thought about moving each flow to it's own section with a dedicated sidebar. While this is possible, I felt it's not the best thing to do because:
a. We don't have a search per flow, the next best thing is to use `ctrl+f` in a flow page.
b. Each flow is currently consist of a single page, and I don't see a good reason to solve this now.
2. Links to the flows are to the "next" version. I don't think this matters a lot, but decided to start like that as version 4 does not exist yet for those changes.
3. The search will be updated only after we'll merge this to master.

### What's need to be done
1. All "guides" moved to be named "legacy guides", we should go over them and try to separate the "generic" parts from the legacy flow parts. All the generic parts should be moved to a different sections later on. There are many missing parts there which were written around in this issue - #705 
2. Add the docs for editor flow, bm flow and monorepo flow.
3. Go over app flow and Create adjust it for today (configuration and CLI reference)


website/blog/2019-03-12-yoshi-4.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Yoshi 4: Slick and polished
+title: Yoshi 4 - Slick and polished
```
ranyitz, line 2: This is not a valid yaml...

website/docs/welcome.md
```diff
@@ -0,0 +1,45 @@
...
+### Business Manager Flow
+This flow is specific to building Business-Manager-Modules.
+
+### Monorepo Flow
+An experimental flow to support monorepos, currently used by the big platforms: Editor-X, Thunderbolt & Editor Elements.
+
+### [Legacy Flow](./legacy-flow)
+> Also known as `yoshi`
+
+This is how we refer to the old flowless yoshi nowadays, it can be configured in many ways and it's generally better to use a different, specific flow.
```
ranyitz, line 45: It is flawless 😉

website/docusaurus.config.js
```diff
@@ -0,0 +1,153 @@
...
+
+  themeConfig: {
+    algolia: {
+      apiKey: '5807169f7e8a322a659ac4145a3e5d8a',
+      indexName: 'wix_yoshi',
+      algoliaOptions: {
+        // The search currently work for version 4 only
+        // Once this is fixed in docusaurus we can use the dynamic
+        // version of the page to filter the results
+        facetFilters: [`version:${versions[0]}`],
```
ranyitz, line 27: This is not very critical as we don't have many users on version 3. In any case, once it will be solved we should update it as well. Especially important for version 5

PR #2279, opened 2020-04-22
Customize module topology through `.module.json`

amitdahan:
### 🔦 Summary
Users rarely only need their own artifact's url.
This lets them customize the topology injected through module-params, via `.module.json` (that eventually affects the generated `app-config-templates/module_X.json` file).

This also removes the unused `TranslationProvider` & `useTranslate` duo, until we properly Do I18n.


packages/yoshi-flow-bm/src/config.ts
```diff
@@ -8,15 +8,26 @@ import { MODULE_CONFIG_PATH } from './constants';
 export interface FlowBMConfig {
   moduleId: string;
   routeNamespace: string;
+  topology: Record<string, { artifactId: string }>;
 }
 
 export function readModuleConfig(cwd: string): FlowBMConfig {
-  const config = readJsonSilent(path.resolve(cwd, MODULE_CONFIG_PATH));
+  const {
+    moduleId = getProjectArtifactId(cwd)!,
```
ranyitz, line 16: It looks like we're calling this function twice, would be better to assign it to a variable.
e.g.
```ts
const artifactId = getProjectArtifactId(cwd);
```

packages/yoshi-flow-bm/src/config.ts
```diff
@@ -8,15 +8,26 @@ import { MODULE_CONFIG_PATH } from './constants';
...
+    routeNamespace = '',
+    topology = {
+      staticsUrl: {
+        artifactId: getProjectArtifactId(cwd)!,
+      },
+    },
+  }: Partial<FlowBMConfig> = readJsonSilent(
+    path.resolve(cwd, MODULE_CONFIG_PATH),
```
ranyitz, line 24: I prefer using `path.join` as `path.resolve()` is actually already a join with `process.cwd()` which can be a bit confusing.

amitdahan:
@ranyitz wdyt of the test + the config setup?

packages/yoshi-flow-bm/src/config/normalize.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  return defaultsDeep(initialConfig, moduleConfigDefaults) as ModuleConfig;
+};
+
+export const normalizePageConfig = (
+  initialConfig: InitialPageConfig,
+  pagePath: string,
+  { moduleId }: ModuleConfig,
+): PageConfig => {
+  const componentId = `${moduleId}.pages.${path.parse(pagePath).name}`;
```
ranyitz, line 39: You can pass the pageName from the loadPageConfig function where it is already parsed. Not critical of course, just a note.

amitdahan: Fixed!

PR #2283, opened 2020-04-23
Yoshi Server: Support BM routing

yanivefraim:
The convention in Business Manager is that calls to `/_api/projectName` will be routed to `/api` (both in production, in Fryinfpan, and locally, in BM testkit).
For example (locally)

`http://localhost:5000/business-manager/3fdba72b-c9e7-4529-9219-807ad4b36d91/projectName` -> `http://localhost:3000/api`, where port 5000 represents BM server and port 3000 represents app server.

Those are some options that we can have in order to fix this (wanted to have a discussion)

TODO: 
- Add tests
- Decide on a solution


packages/yoshi-server/src/server.ts
```diff
@@ -68,6 +68,7 @@ export default class Server {
 
   public handle: RequestHandler = async (req, res, next): Promise<void> => {
     try {
+      // req.url = req.url.replace('/api/_api_', '/_api_');
```
ronami, line 71: can probably be removed

PR #2287, opened 2020-04-23
Add flow editor to docs

yavorsky:
### 🔦 Summary
This PR moves docs related to editor flow from internal directory to the new amazing hipster docusaurus website.

I've updated a bit structure, so editor flow has own page with own sidebar 😎

Next steps I want to do:
- I think we it's a bit confusing having `4x` vs `next` version. Maybe we should just have 4x as a next since it's latest one? So we don't need to maintain same stuff in 2 different directories? cc @ranyitz 
- Change the name on the main page from Yoshi to Yoshi Universe and add planets for each big project (yoshi-server, library flow, editor flow, etc) 🌍🌎🌏
- Remove flows from the main sidebar. (currently I've added a redirect to separate page for editor flow for consistency)


ranyitz:
> I think we it's a bit confusing having 4x vs next version. Maybe we should just have 4x as a next since it's latest one? So we don't need to maintain same stuff in 2 different directories? cc @ranyitz
Change the name on the main page from Yoshi to Yoshi Universe and add planets for each big project (yoshi-server, library flow, editor flow, etc) 🌍🌎🌏
Remove flows from the main sidebar. (currently I've added a redirect to separate page for editor flow for consistency)

Wow! All of them sound to me like great ideas, feel free to open either a PR or a dedicated issue to discuss website improvements 💯

yavorsky:
@ranyitz Agree, overview looks bad in general. We've added it 2 months ago and it's a bit outdated. Since we'll have good relevant docs, I'll try to include documentation update with any breaking changes / new features

PR #2288, opened 2020-04-24
Brand new main page based on yoshi flows universe idea 🚀

yavorsky:
### 🔦 Summary
This PR updates main page to preview 'Yoshi Universe' idea.
Planets are representing each big project like flows and yoshi-server.
We can enhance it in future. I also tried to make it good for both light and dark themes, but for me dark feels a bit better. Maybe we should use it by default?

DEMO: https://wix-yoshi-website-pr-2288.surge.sh


website/src/pages/Header/header.module.css
```diff
@@ -0,0 +1,50 @@
...
+  justify-content: center;
+  align-items: center;
+  flex-wrap: wrap;
+}
+
+@media screen and (max-width: 966px) {
+  .planetsWrapper {
+    display: none;
+  }
+}
```
ranyitz, line 50: ❤️ 📱

PR #2289, opened 2020-04-24
Change tagline and improve planets logic

ranyitz:
### 🔦 Summary
After seeing @yavorsky's #2288 I realized that the tagline doesn't make sense anymore.

So This PR changes it.

There are also small design improvements.

1. There were cases when 3 planets had the same color/size, creating a suboptimal galaxy.
2. The features' background did not blend nicely with the universe color in the dark mode, so this PR fixes it.
3. This is maybe minor, but when searched something, the planets were assigned with different values (another render), which created a glich when all of the planets changed for a split second. This PR assigns the planets values only once per session, meaning that only a refresh will assign with new ones.


ranyitz:
Should be available here - https://wix-yoshi-website-pr-2289.surge.sh/

website/src/pages/index.js
```diff
@@ -2,7 +2,7 @@
 import React from 'react';
 import classnames from 'classnames';
 import Layout from '@theme/Layout';
-import Header from './Header';
+import Universe from './Universe';
```
yavorsky, line 5: 😏

PR #2301, opened 2020-04-27
Improve Elements' server bundle size

ronami:
### 🔦 Summary

This PR improves the bundle size and the memory foot-print of Elements' server bundle. Lodash started being used and it makes the bundle significantly larger, this alone took it from 6XXkb to 2.1mb (before we started minifying the server bundle and with full inline source maps).

To improve, I externalized Lodash so it's expected to be available in TB (through `package.json`). I noticed we externalized React and Svelte for all node servers and externalizing Lodash for everyone may be a breaking change (if it happens to be in devDependencies instead of dependencies). I changed this to externalize those only for Elements.

I also changed the (inline) source maps for the server code to be cheap source maps. This will make stack traces less accurate but reduces the size of the bundle significantly (along with better defaults for minifying the server bundle) down to a reasonable 303kb.


gonengar:
Following

PR #2302, opened 2020-04-27
Decrease the size of Elements/TB stats file

ronami:
### 🔦 Summary

See title. This should make it easier to download and load the stats file. I noticed recent changes now require only the `namedChunkGroups ` field from the stats file which means that we can reduce its size significantly.

This will mean that future changes to the code that use the stats file will have to be in sync with Yoshi. If we see that this slows us down we can consider opening the stats data and/or transformation to a configurable value in build time.


ranyitz:
@ronami I'm interested, what are the numbers? How much KB is this removing from the elements/TB?

ronami:
> @ronami I'm interested, what are the numbers? How much KB is this removing from the elements/TB?

9.6KB 😄

gonengar:
Following

PR #2313, opened 2020-04-29
Reduce amount of data saved in stats

Gongreg:
### 🔦 Summary
We don't need all of the stats to support analyzer, I've disabled some of them. This way we avoid writing crazy amount of data to webpack-stats.json . The stats disabled are equivalent of settings stats to `detailed`: https://webpack.js.org/configuration/stats/#stats-presets

### 🗺️ Test Plan
Tested on WSR => 930MB to 24MB and analyzer still runs.


ranyitz:
Thanks @Gongreg.

I'm not an expert in webpack's stats, but I thought that [`stats.sources`](https://webpack.js.org/configuration/stats/#statssource) is what decides whether the source code will be included or not.

I don't know what are the usages of the stats in Wix, but completely removing `chunkModules` may be a breaking change.

It's cool though that the bundle analyzer can run without it. I'm not sure what's the importance of `chunkModules` or how low would we get by just eliminating the `sources`.

@ronami what do you say?

Gongreg:
Disabling those two options is equivalent to settings stats: 'detailed'. And it removes the heavy part of the stats.

PR #2316, opened 2020-04-30
yoshi serve support for monorepo

ranyitz:
### 🔦 Summary
Yoshi serve currently doesn't support monorepo flow, this PR adds this capability.

It also adds the app name in order to distinguish between the apps in the monorepo:

**BEFORE**
image

**AFTER**
image

### What were the needed modifications?
1. A lot of functions needed to accept `cwd` as parameter instead of `process.cwd()`.
2. We have to know the port on which each app server needs to be served on. (This currently handled via `PORT` environment parameter which is configured on the main `package.json` of the monorepo) I added a configuration option to yoshi config: `servers.app.port` in order to get this information statically for each app.
3. Added a `useAppName` for the serverProcess so it would know to use it instead of the regular `SERVER`, currently only used by the monorepo serve function.
4. Thunderbolt uses `process.env.NODE_ENV=test` in oppose to `development` which is being used by our default serve mechanism. I added an `env` parameter to let them override/add any environment parameter.

All of this functionality was also added to every other app/monorepo that may need it.

### What needs to be done:
* [x] Check that everything works as expected with thunderbolt, see PR https://github.com/wix-private/thunderbolt/pull/5423 with the modifications

* [x] Decide if we want to expose this as another command or as a node API. Letting `yoshi-common` to handle monorepo required me to either use monorepo logic from `yoshi-common` or instead just write the logic there. Both didn't feel right. I rather implement a monorepo serve in `yoshi-flow-monorepo` package. We can think on the concern of other tools later on, but we can create a `yoshi-serve` package that routes between all of the possibilities including old yoshi versions (Like @Schniz wisly suggested)

* [x] Add documentation - there is modification needed for the regular yoshi serve with the additional arguments that it now accepts.

* [x] Add tests - A happy flow that serve can run 2 apps at the same time.

______________

Regarding how to expose this. We'll start with adding the capability under `yoshi-flow-monorepo` and we'll enhance/change it in the future if we'll need.


packages/yoshi-flow-library/src/commands/test.ts
```diff
@@ -69,7 +69,11 @@ const test: cliCommand = async function(argv, config) {
 
     if (config.bundleConfig) {
       closeCdn = (
-        await startCDN({ ssl: false, port: config.bundleConfig.port })
+        await startCDN({
+          ssl: false,
+          port: config.bundleConfig.port,
+          cwd: process.cwd(),
```
ronami, line 75: super tiny, isn't it the default already?

ranyitz: It actually used to be. I've changed that to use `path.join` with and argument of `cwd` https://github.com/wix/yoshi/pull/2316/files/fd5041e71b67b434f9d672a25bfaee0bbb54c6aa#diff-8f604946314b354dbf0e99035065c183R19

I feel that having it as default can be confusing so instead I rather provide it explicitly to the function from the places that start the CDN.

ranyitz:
**Update.**

All left to be done is to add another file at the root for nicer api (`yoshi-flow-monorepo/serve`) and document the API.

PR #2333, opened 2020-05-05
Use `data-hook` instead of `data-testid` in templates

sidoruk-sv:
### 🔦 Summary

First I thought that using out of the box `data-testid` would be better. But when realized that a developer will ever need to select something by data-hook of the WSR component, he will need to create some hacks for this special case.

So as for me, using `data-hook` (default at Wix for testing hooks) everywhere is a better strategy, WDYT?


amitdahan:
I'm all-in for consistency, so I'm leaning towards this regardless.
But I'm interested to know what you mean by hacks for WSR?
You mean in cases where their testkits aren't enough? That you have to manually query for their `data-hook`s?

sidoruk-sv:
> You mean in cases where their testkits aren't enough? That you have to manually query for their data-hooks?

Yes, sometimes WSR testkit is not enough.
Also, testing-library is not everywhere. All of WIX projects use data-hook for tests, regardless of UI and testing frameworks.

amitdahan:
Throwing an idea - I wonder if our jest preset could do this automatically, if user has testing-library installed, in the spirit of providing good defaults.

This would keep us that much closer to zero-config, since this forces all `testing-library`-using templates having a `spec-setup` file.

This'd be a breaking change for users who rely on the current defaults.

@ranyitz wdyt?

ranyitz:
@amitdahan I agree. We either do this in the next major version or attempt to do this as a non breaking change.

We can start with a configuration in the `jest-yoshi.config.js` which will be turned on in the templates, that would become the default on the next major version.

PR #2335, opened 2020-05-06
Allow configuring serverExternals

yavorsky:
### 🔦 Summary
This should help with:
https://jira.wixpress.com/projects/FEDINF/issues/FEDINF-1094

Like any other externals (webworkerExternals/externals) we will let the user configure it because they are changing according to the app/platform.

Thunderbolt's server (which downloads thunderbolt-elements) needs to verify those externals are located in the node_modules directory.

We are currently not supporting different types of externals in array, but it's pretty ok for start and not sure a lot of user's using it like this.

Will open a follow-up PRs to `thunderbolt-elements`, `editor-elements-library` and `thunderbolt-app`

### DIFF
--- a/packages/yoshi-common/src/webpack.config.ts
+++ b/packages/yoshi-common/src/webpack.config.ts
@@ -1113,8 +1113,8 @@ export function createBaseWebpackConfig({
               // relevant for Thunderbolt's use-case.
               if (serverExternals) {
                 if (typeof serverExternals === 'function') {
-                  return serverExternals(context, res, (ctx, result) =>
-                    callback(ctx, result ? `commonjs ${request}` : undefined),
+                  return serverExternals(context, res, (error, result) =>
+                    callback(error, result ? `commonjs ${request}` : undefined),
                   );
                 } else if (isPlainObject(serverExternals)) {
                   const configuredExternal = findExternalFromObject(
--- a/packages/yoshi-common/src/webpack.config.ts
+++ b/packages/yoshi-common/src/webpack.config.ts
@@ -1,7 +1,11 @@
 import path from 'path';
 import os from 'os';
 import fs from 'fs-extra';
-import webpack from 'webpack';
+import webpack, {
+  ExternalsElement,
+  ExternalsObjectElement,
+  ExternalsFunctionElement,
+} from 'webpack';
 import {
   SRC_DIR,
   STATICS_DIR,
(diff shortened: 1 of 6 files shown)


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -1091,7 +1115,29 @@ export function createBaseWebpackConfig({
               // Svelte and React should always be external on the server side. Only relevant when more than one
```
ranyitz, line 1115: This comment belongs to the "forceSpecificNodeExternals" line.

I would add a comment that explains that we let the user to configure the externals for the server.

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -200,6 +200,8 @@ export function createServerWebpackConfig(
     nodeExternalsWhitelist: libs.map(pkg => new RegExp(pkg.name)),
     useAssetRelocator: pkg.config.experimentalUseAssetRelocator,
     forceMinimizeServer: isThunderboltElementModule(pkg),
+    serverExternals: pkg.config.serverExternals,
+    // Remove next line when serverExternals will be implemented on thunderbolt app and element modules
```
ranyitz, line 204: Actually, `thunderbolt-app` is a regular server that doesn't need externals, you probably mean `thunderbolt-elements`.

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -333,7 +331,7 @@ export function createSiteAssetsWebpackConfig(
     cwd: pkg.location,
     configName: 'site-assets',
     target: 'node',
-    useNodeExternals: false,
+    serverExternals: {},
```
ranyitz, line 334: It worth adding a comment here, that this means we don't have any externals (even for node bundle).

NitayRabi:
If I understand correctly - what we need to make sure that `react` (svelte is no longer relevant) should **NOT** be bundled into the server bundle, and that's supposed to be handled in package json configuration?

yavorsky:
@NitayRabi yeah, in `yoshi.config.js` 🙂
Like here: https://github.com/wix-private/editor-elements/pull/614

NitayRabi:
LGTM (from editor-elements point of view)

PR #2341, opened 2020-05-06
run webpackbar in basic mode in every ci environment

ranyitz:
### 🔦 Summary
Yoshi should not use the `fancy` mode of webpackbar when used on CI.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -126,9 +127,9 @@ function getProgressBarInfo(
   if (isMonorepo) {
     obj.name = `${stripOrganization(packageName)}\n  ${obj.name}`;
   }
-
+
```
ranyitz, line 130:
```suggestion

```

PR #2346, opened 2020-05-06
Add lodash to create yoshi app deps

yavorsky:
### 🔦 Summary
We are using it in runtime while generating dynamic templates.

It's sometimes easy to miss stuff like this since you already has it in monorepo's root deps and everything works.


yavorsky:
@ranyitz Yeah, I thought about the same. I'll add an issue for this.

PR #2348, opened 2020-05-06
Prevent asking language for projects if only one is supported

yavorsky:
### 🔦 Summary
We have a bug for experimental flows, that don't support JS.
ezgif com-video-to-gif
So I've updated our general solution to use extended-promts and added last question to be dynamic. As a result if we support only one language, we will select it automatically.

Other big change - rename Babel to Javascript.

I think this fix will be interested for other experimental flow maintainers cc @ranyitz @amitdahan


packages/create-yoshi-app/src/getQuestions.ts
```diff
@@ -1,6 +1,7 @@
 import getGitConfig from 'parse-git-config';
-import { PromptObject } from 'prompts';
+import { capitalize } from 'lodash';
```
yavorsky, line 2: 🤞

packages/create-yoshi-app/src/createApp.ts
```diff
@@ -37,13 +36,16 @@ export default async ({
 
     // If we don't have template model injected, ask the user
     // to answer the questions and generate one for us
-    templateModel = await runPrompt(workingDir);
+    const runPrompt = require('./runPrompt').default;
```
ranyitz, line 39: Why we've moved it here? (instead of `import`)

packages/create-yoshi-app/src/getQuestions.ts
```diff
@@ -1,6 +1,7 @@
...
 import templates from './templates';
+import { ExtendedPromptObject } from './dev-center-registration/extended-prompts';
```
ranyitz, line 4: I think it's time to move `extended-prompts` from `dev-center-registration`

packages/create-yoshi-app/src/createApp.ts
yavorsky: b/c we want to import it only if needed. It's happening in condition, and to be consistent with `runPrompts` of devCenter optimization, we have cases when we already passed templateData and don't need this module.

yavorsky: And the second reason - we can correctly mock it in tests 🙃

packages/create-yoshi-app/src/getQuestions.ts
yavorsky: Agree! Will do in separate PR.

PR #2357, opened 2020-05-08
Add platform app studio component to be generated via wizard and generate editor builder automatically

yavorsky:
### First change is a new Dev Center Wizard for platform apps
Follow-up for https://github.com/wix/yoshi/pull/2318

Add dev center wizard we've implemented for OOI apps to support app studio builder apps.

It's working almost the same, but
- Ask only widget, no pages here
- Use different model for widgets (not component URL, etc)

### Second change - generate editorScriptBuilder according to the file structure
Currently in editor.app.ts we should do:

*editor.app.ts*
```ts
import speakersWidget
import speakerWidget

const editorReady = () => ...
const eventHandler = () => ...

export default editorScriptBuilder()
	.withEditorReady(editorReady)
	.withAppManifest(appManifest)
	.withEventHandler(eventHandler)
	.withWidget(speakersWidget)
	.withWidget(speakerWidget)
	.build();
```

but since we know all the structure, we can allow users just export needed handlers in `editor.app.ts` and generate builder under the hood:

*editor.app.ts*
```ts
export const editorReady = () => ...
export const eventHandler = () => ...
```

It will be much cleaner and we can easily add new methods we support to our wrapper.

I've included this change to wizard PR, since our editor.app.ts should inport all editor controllers, which becoming flexible for now. So to not make double-work, we can just hide it to the build step instead of create-yoshi-app generation 

Todo:
- [ ] Create a URL page and set it as a start url in template. This page will have some guide how to finish set up. The same we already have OOI - https://www.editor-flow-dev.com (cc @maoryosef2 I think I need your help here 🙂)


packages/create-yoshi-app/src/dev-center-registration/getQuestions.ts
```diff
@@ -205,20 +251,37 @@ export default (): Array<ExtendedPromptObject<string>> => {
...
               before(answers) {
                 if (!answers.components) {
                   answers.components = [];
                 }
                 return answers;
               },
+              async getDynamicChoices(answers, context) {
+                if (
+                  context.templateDefinition.name ===
+                  'flow-editor - Out of iFrame'
```
ranyitz, line 263: This constant should be written in some variable.

Something like:
```ts
const OOI_TEMPLATE_NAME = 'flow-editor - Out of iFrame';
```

packages/create-yoshi-app/src/templates.ts
```diff
@@ -44,15 +44,19 @@ if (process.env.EXPERIMENTAL_FLOW_BM === 'true') {
     language: ['typescript'],
   });
 }
+
 if (process.env.EXPERIMENTAL_FLOW_EDITOR === 'true') {
-  const usePlatformTemplate =
-    process.env.EXPERIMENTAL_PLATFORM_TEMPLATE === 'true';
+  templates.push({
+    name: 'flow-editor - Out of iFrame',
+    path: toTemplatePath('flow-editor'),
```
ranyitz, line 51: Not for this PR, but the template path probably needs to be change to `flow-editor-ooi` in the future.

packages/create-yoshi-app/src/dev-center-registration/getDefaultAnswers.ts
```diff
@@ -0,0 +1,28 @@
...
+    appId: '945c0d1a-5d9d-422d-adc0-da8fb3fe8c8f',
+    appName: 'Editor flow test app',
+    components: [
+      {
+        id: '945c0d1a-5d9d-422d-adc0-da8fb3fe8c8f-yik50',
+        type: 'STUDIO_COMPONENT',
+        name: 'firstWidget',
+      },
+    ],
+  };
```
maoryosef2, line 27: creating the widgets themselves though the CLI is a bit problematic, since they need to be linked to their stucture, 
this is a step that still needs to be done through app studio

packages/create-yoshi-app/src/dev-center-registration/getQuestions.ts
yavorsky: Agree 🙂

packages/create-yoshi-app/src/templates.ts
yavorsky: yeah :+1:

yavorsky:
@maoryosef2 Updated. So for now we just asking to fill app name and then registering an app with `firstWidget`.
Would be good to set `startUrl` as a app builder URL and open page with it after `yoshi start`.

Or even create some website with **Next steps** like we are doing for OOI: https://www.editor-flow-dev.com/?appName=NewApp

packages/create-yoshi-app/src/dev-center-registration/getQuestions.ts
```diff
@@ -53,27 +60,66 @@ const wasSelected = (ids: Array<string>) => (component: DevCenterComponent) =>
...
+  ) {
+    return generateOOIComponentData(projectName, componentName, componentType);
+  }
+  return generateStudioComponentData();
+};
+
+const generatePlatformComponentData = (projectName: string) => {
   const baseUrl = getBaseUrl(projectName);
   return {
     viewerScriptUrl: `${baseUrl}/viewerScript.bundle.min.js`,
```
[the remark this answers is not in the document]
yavorsky: @maoryosef2 we are not generating platform component for app builder apps, so it's only for OOI.
Do you think we need?

packages/create-yoshi-app/templates/flow-editor-platform/typescript/src/components/{each%flowData.components:name%each}/viewer.controller.e2e.ts
```diff
@@ -0,0 +1,34 @@
...
+        targetText,
+      );
+    } catch (e) {
+      const actualWidgetText = await page.$eval('h2', node =>
+        node?.textContent?.toLowerCase(),
+      );
+      throw new Error(
+        `The text "${targetText}" was not found on the page. "${actualWidgetText}" found instead`,
+      );
+    }
```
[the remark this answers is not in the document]
yavorsky: sounds good

yavorsky:
I'm merging it. We can update in future viewerUrl in platform template and create a platform component with viewer and editor scripts in dev center

PR #2359, opened 2020-05-08
[templates] use locals for ejs variables

sidoruk-sv:
### 🔦 Summary

As @OriMarron found (https://wix.slack.com/archives/G6WS71Z2B/p1588865483004800?thread_ts=1588864942.003900&cid=G6WS71Z2B), using `variables` inside `.ejs` with `locals.variableName` (https://ejs.co/#docs) is more bulletproof solution then without it.

We sometimes get in NewRelic:
```
variableName is not defined
```

For us, this appears after GA:
- fresh GAed `index.ejs` from hadron;
- old `server` file;
- new or removed variable leads to the 500 error page showing to users for a short period of time (tiny incident).

The problem here is that if something will be forgotten it would be `undefined` in ejs after GA (so e2e tests would not fail for `server.js` file)
Maybe need some more checks and fallbacks, but to start a discussion.
WDYT?


ranyitz:
Thanks @sidoruk-sv. I can't see the slack thread for some reason. I understand that there is a tradeoff between the way failures are seen in production to build time validation. @yanivefraim What do you think?

sidoruk-sv:
> all variables in ejs are available under locals for example <%- locals.cookieConsentPolicy %> is the same as <%- cookieConsentPolicy %>, but it will not crash, just return undefined
> it’s here, but not well documented https://ejs.co/ search for locals``

ranyitz:
@sidoruk-sv I want the **"Why"** to be clear here. (`use locals for ejs variable` is the how/what).

Do you want new relic's logs to be clearer?

sidoruk-sv:
@ranyitz not only NewRelic, when these parameters is unsync there is 500 Error page on production, even if not vital variable failed

ranyitz:
@sidoruk-sv I guess that there are some variable which are vital and some which are not. `baseStaticsUrl` is critical while `title` is a bit less. Though I have to say that in case of a small bug I would probably want to know about it, even if it's a small one. (Instead of showing something in production which I don't know what it is, or how bad it is)

In this case, I would love to get to the root of the problem, do you know why is there a mismatch between the server and the `ejs` that it serves in the first place?

sidoruk-sv:
@ranyitz because that how hadron works, I think


https://wix.slack.com/archives/C0NJPS8N5/p1586798191147200?thread_ts=1586788431.144000&channel=C0NJPS8N5&message_ts=1586798191.147200

ranyitz:
@sidoruk-sv Thanks! From reading this thread I understand that you should use a test-bed and do it in 2 steps.

> So yes, if you remove something from the render-model & ejs and you want to testbed, you need to do it in two steps.

As far as a concern this a sync problem.

I think we should strive to make hadron take the local `ejs` version for production cases (I may be missing something, but it sounds logical to me)

I'm not sure what people do today to avoid 500 after a change in the render model and a deployment. Maybe @saarkuriel & @AdirAmsalem knows.

saarkuriel:
> @sidoruk-sv Thanks! From reading this thread I understand that you should use a test-bed and do it in 2 steps.
> 
> > So yes, if you remove something from the render-model & ejs and you want to testbed, you need to do it in two steps.
> 
> As far as a concern this a sync problem.
> 
> I think we should strive to make hadron take the local `ejs` version for production cases (I may be missing something, but it sounds logical to me)
> 
> I'm not sure what people do today to avoid 500 after a change in the render model and a deployment. Maybe @saarkuriel & @AdirAmsalem knows.

This is what we do, testbed in 2 steps...

AdirAmsalem:
Hi all

@sidoruk-sv, I disagree regarding `locals` being "more bulletproof solution than without it". 
Using `locals` will indeed allow the ejs to render when a variable is missing, so it won't crash - but that means it'll also allow having typos in variable names, mistakenly remove used variables, etc. 

For the underlying issue, as @ranyitz mentioned that's assumed to be a sync issue. On version change (GA,Testbed,RB, etc) - statics versions are updated faster than the server code in the k8s pods (they are updated on the "old" servers), which makes the server render the updated-version EJS that needs values the existing-version server doesn't pass, so it crashes.

Although it sucks, and you're right that it needs to be solved - those are two different issues.

Using `locals` hides the issue, it doesn't solve it. Once there will be a real issue - you won't notice it until it'll break production. In my opinion, we should fail fast in this case. If a variable doesn't exist the rendering should fail, fast and clear, no room for doubt.

If you're asking me, the solution is to make them update together, fixing the issue at its core. @ranyitz mentioned one way to do this (taking the local ejs). I'm not familiar with all the ways and all the tradeoffs, this is something that should be thought about - but that's the "better" solution in my opinion.

ranyitz:
Thanks @saarkuriel And @AdirAmsalem! I'm with @AdirAmsalem on this one, I don't think that silently ignoring the error is a good solution here and it actually causing problems which are harder to understand for other cases.

In the [thread](https://wix.slack.com/archives/C0NJPS8N5/p1587021337155000?thread_ts=1586788431.144000&cid=C0NJPS8N5), I've seen that @hugebdu tagged @berman7 & @yuval1488 as the contact guys for hadron/testbed. This looks like a sync problem that we would love to know if you're aware of or/and have ideas for solution.

sidoruk-sv:
Agreed, so closing this PR for now.
But still, need to somehow manage the problem with Hadron.

PR #2363, opened 2020-05-10
Migrate BM templates to use `wix-i18n-config`

amitdahan:
### 🔦 Summary
~~Blocked by https://github.com/wix-private/fed-infra/pull/1903.
Waiting for it to merge and for `@wix/wix-i18n-config` to be released as `5.0.0`.~~

~~This should still pass on `^5.0.0`, but no polyfills are applied, so it's only green on supported browsers.~~

`@wix/wix-i18n-config@5.0.0` is released! This migrates the BM template to use that instead of manually using `i18next` & `react-i18next`.


packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
```diff
@@ -1,35 +1,48 @@
...
-      <I18nextProvider i18n={i18n(locale, baseUrl)}>
-        <App />
-      </I18nextProvider>
-    );
-  }
-}
+const AppContainer = ({ locale }) => {
+  useOnce(() => notifyViewStartLoading(COMPONENT_NAME));
+
+  const i18n = useMemo(
```
ronenst, line 26: cool, this solves a serious issue we've been having with apps re-downloading translations on every mount.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
```diff
@@ -1,35 +1,48 @@
...
-    const baseUrl = config.topology.staticsUrl;
-    return (
-      <I18nextProvider i18n={i18n(locale, baseUrl)}>
-        <App />
-      </I18nextProvider>
-    );
-  }
-}
+const AppContainer = ({ locale }) => {
+  useOnce(() => notifyViewStartLoading(COMPONENT_NAME));
```
ronenst, line 24: Why once? It should be on every mount in theory?

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
```diff
@@ -1,35 +1,48 @@
...
+
+  const i18n = useMemo(
+    () =>
+      initI18n({
+        locale,
+        asyncMessagesLoader: locale2 =>
+          import(`./assets/locale/messages_${locale2}.json`),
```
ronenst, line 31: This is interesting, in theory you can now remove this:
https://github.com/wix/yoshi/blob/create-yoshi-app/bm-i18n-config/packages/create-yoshi-app/templates/business-manager-module/javascript/app-config-templates/module_%7B%25PROJECT_NAME%25%7D.json#L14

Since it's now unused. On the other hand, this PR Will make it more difficult to transition to independent deployment of translations, but I'm not sure that's an issue.

packages/create-yoshi-app/templates/business-manager-module/typescript/src/client.tsx
```diff
@@ -1,29 +1,42 @@
...
+};
+
+const AppContainer: FC<IBMModuleParams> = ({ locale }) => {
+  useOnce(() => notifyViewStartLoading(COMPONENT_NAME));
+
+  const i18n = useMemo(
+    () =>
+      initI18n({
+        locale,
+        asyncMessagesLoader: locale2 =>
```
ronenst, line 29: is locale2 really required? Can't is just be:
```suggestion
        asyncMessagesLoader: () => import(`./assets/locale/messages_${locale}.json`),
```

vybu:
Looks good!

I thinks it's worth changing messages to use an interpolation just so a correct syntax is demoed.

```
{
  "app.title": "Hallo Wereld",
  "app.intro": "Get begonnen hier: {introUrl}"
}
```

packages/create-yoshi-app/templates/business-manager-module/typescript/src/client.tsx
amitdahan: Yeah this made me think, if we have to explicitly pass `locale` on each "init" call, why must the loader function accept a locale?
Is there a different use case where we don't pass `locale`?
I'll fix our usage regardless :)

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
amitdahan: Yep, I wondered if it was worth "killing" the `staticsUrl` "default" topology "convention" (?), or leave it cause it may be inevitably added on most projects?

Do you know of any other use cases for having your own statics base-url injected to your app other than translations (deployed separately / just using the `xhr-backend` thingy)?

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
amitdahan: Previously this was called in the constructor, I tried imitating that the best I could.
This will happen on each mount (cause the underlying ref used will only be reset on a fresh mount), but earlier than a `didMount` (`useEffect` with no deps) would.
We usually have the notify-view-finished in a `didMount` somewhere down the tree, so I thought putting this in the mount phase would make this almost instant, which I assume is the original intention?

packages/create-yoshi-app/templates/business-manager-module/typescript/src/client.tsx
vybu: In other cases i18n is initialized at file/module level, rather than component. So if you change locale after app has been initialized, you'll need a loader function to know what locale to load. That's why it accepts argument.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
ronenst: I consider the need for `topology` an edge case most apps don't need and the current usage apps have a bit of an anti-pattern (a module shouldn't depend on a different module to provide it, it already know it...)

Might be worth leaving an empty `topology: {}` maybe but otherwise leaving the default IMO can just cause confusion.

Of course, it's not really critical either way.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
ronenst: I see thanks, only issue is `useOnce` seems a "boilerplatey" but I don't think I have any alternative.

Maybe in the future we can consider if this method should be at the module file level.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
amitdahan: I'm all for zero-config! 
If you consider it an edge-case, I'd rather remove the `config` (and `topology`) altogether.
Wdyt?

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
amitdahan: Hmmm I tried thinking of alternatives, since it really wasn't pleasant adding this much code to the generators just for the sake of a "proper refactor".

Wdyt of this?
```typescript
notifyViewStartLoading(COMPONENT_NAME);

const AppContainer = ({locale}) => {
  // ...
};

export default AppContainer;
```

This relies on the assumption that the file is never accidentally _eager_.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
ronenst: While I believe `config` is supposed to be optional, I'm not sure we actually have a case like that so it might be the first time BM sees a module like that.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
ronenst: I considered it, but is it good practice? Also I'm not sure if it will work for remounts (or it might, but just as a side-effect of it being re-attached to the dom?)

ronenst: Of course there's the option of leaving it a class component 🙀

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
ronenst: Our test module has none so it's probably fine although it isn't exactly a production module:
https://github.com/wix-private/business-manager/blob/master/service-discovery/service-discovery-server/app-config-templates/module_TEST_MODULE.json

amitdahan: Is it safe to assume that if the local testkit works - it's safe for prod?

ronenst: I think it should be fine but we have a lot of cases. In any case I think if someone tries to integrate and fails on that we should fix it and the workaround will be to add it until it's fixed.

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
amitdahan: Yeah it shouldn't work for page remounts :(

Regarding class components -
The reason I felt I was only "improving" the template by changing it to a function is - (some) users want to use hooks, and they simply can't with a class component, this isn't true for the opposite direction.
But yeah - the only exception (right now) is the ctor.

Wdyt about moving it to a `useEffect` instead? 
It's still 2 different components (`client` & `App`), I assume this only _really_ matters for async stuff and not for waiting for react, right?

amitdahan:
This fails the template tests because of an issue with a dependency (`@wix/react-test-wrappers`) that BM's testkit brings.

When [this](https://github.com/wix-private/fed-infra/pull/1924) (or a different solution) gets merged - it'll go green ✅!

packages/create-yoshi-app/templates/business-manager-module/javascript/src/client.js
ronenst: I think `useEffect` should be good enough until proven differently.

This is just boiler plate code that should be doing most of the stuff in `App` anyway, does having a class anywhere in the tree negate the ability use hooks?

amitdahan: Nope, doesn't affect the rest of the tree, just prevents using hooks inside the specific render function.

PR #2370, opened 2020-05-11
better error handling for webpack errors during build

ranyitz:
### 🔦 Summary

When running `yoshi build` sometimes you can only see parts of the error, which are not enough to find where is the problem.

Take this example:
image

It didn't showed me important parts of information regarding the problem. This what you will see after the fix:

image

__________

This PR also logs all of the errors in case there are multiple errors that Webpack caught. After debugging some problems with @giladsegal from Editor Elements, we've realize it would be much nicer to see all of the errors in the context of running `yoshi build` as it would not require the developer to fix a single problem and run the build again. So while showing only the first error may be true for `yoshi start` it can slow us down during build.


packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -226,13 +226,11 @@ async function runWebpack(
     const messages = formatWebpackMessages(webpackStats.toJson({}, true));
 
     if (messages.errors.length) {
-      // Only keep the first error. Others are often indicative
-      // of the same problem, but confuse the reader with noise.
-      if (messages.errors.length > 1) {
-        messages.errors.length = 1;
-      }
+      // Error handled by webpack
+      console.log(chalk.red('Failed to compile.\n'));
```
ranyitz, line 230: We introduce a bit of duplication here, but I think it's currently a good tradeoff, as the change to the code is quite small.

PR #2373, opened 2020-05-11
Yoshi Server: add a Business Manager default baseUrl

yanivefraim:
We want the default baseUrl to be `/_api/<packageName>`, so it will fit the business manager convention.

### DIFF
--- a/packages/yoshi-common/src/webpack.config.ts
+++ b/packages/yoshi-common/src/webpack.config.ts
@@ -702,7 +702,7 @@ export function createBaseWebpackConfig({
               'process.env.ARTIFACT_ID': JSON.stringify(getProjectArtifactId()),
             }
           : {}),
-        'process.env.packageName': JSON.stringify(stripOrganization(name)),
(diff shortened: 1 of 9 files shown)


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -702,6 +702,7 @@ export function createBaseWebpackConfig({
               'process.env.ARTIFACT_ID': JSON.stringify(getProjectArtifactId()),
             }
           : {}),
+        'process.env.packageName': JSON.stringify(stripOrganization(name)),
```
ronami, line 705: Really small but I think we should have all caps to match the convention of env vars.

packages/yoshi-server-client/src/index.ts
```diff
@@ -10,7 +10,7 @@ import { createHeaders } from '@wix/headers';
 import { joinUrls } from './utils';
 
 type Options = {
-  baseUrl?: string;
+  baseUrl: string;
```
ronami, line 13: I think this change will tell TypeScript that `baseUrl` is mandatory so the default won't work

PR #2379, opened 2020-05-12
fix 'a different process is already running on port 3200' error when root folder has whitespace

Akrabut:
### The issue

Followed the instructions in the [FED onboarding program](https://github.com/Akrabut/fed-crash-course/tree/master/exercises/client-tdd) and `npx jest --watch` failed to run with error:

```
 ● Test suite failed to run

    A different process (/Users/orh/Desktop/FED) is already running on port '3200', aborting.
```

Folder name is 'FED Training' and everything past the first whitespace was ignored.
This happens because `awk '{print $9}'` treats everything after said whitespace as 10th column.

### Reproduce

1. `create-yoshi-app` in folder with whitespaces in name.
2. `npm start`.
3. `npx jest --watch` with any puppeteer tests.

### Solution
`awk '{print substr($0, index($0,$9))}'` takes all text from 9th column to end of the line ($0).

### DIFF
--- a/packages/yoshi-helpers/src/utils.ts
+++ b/packages/yoshi-helpers/src/utils.ts
@@ -108,7 +108,7 @@ export const getProcessIdOnPort = (port: number) => {
 
 function getDirectoryOfProcessById(pid: number) {
   return childProcess
-    .execSync(`lsof -p ${pid} | grep cwd | awk '{print $9}'`, {
+    .execSync(`lsof -p ${pid} | grep cwd | awk '{print substr($0, index($0,$9))}'`, {
       encoding: 'utf-8',
     })
     .toString()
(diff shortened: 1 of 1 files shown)


ranyitz:
@Akrabut It seems that lint is failing, it's just prettier formatting. Can you push a fix?
image

Akrabut:
I'm trying to run `npm run lint` after having installed but get the following error:
image

ranyitz:
@Akrabut have you installed using `yarn`? Usually we just run `yarn` in the root and everything works fine.

Akrabut:
Ah I forgot about it and used NPM.

Any reason for `getDirectoryOfProcessById` to be the only function in the file that isn't declared as arrow function, by the way? thought about changing it for consistency.

ranyitz:
@Akrabut There is not reason that I can think of. Feel free to change it if you want.

PR #2384, opened 2020-05-12
Integrate sled to yoshi flow editor ooi template

yavorsky:
### 🔦 Summary
Generate basic sled tests for each component.
We are using drivers approach, creating a common class and trying to extend from it in each widget.

Initially after bootstrapping the app, we'll go to https://www.editor-flow-dev.com and verify component with specific hook was rendered.

For the https://www.editor-flow-dev.com we are generating this hook based on query params.
This approach will be useful after people updated viewerUrl in `sites.ts` to own website with installed app and then don't need to change tests a lot, since selector will work for local app!

Also `sled` script was added and we are running in by default.


packages/create-yoshi-app/templates/flow-editor/typescript/sled/ssr/{each%flowData.components:name%each}/{%name%}.e2e.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+import { appName } from '../../../.application.json';
+import { ViewerE2EDriver } from './{%name%}.e2e.driver';
+
+describe('{%name%}', () => {
+  let driver: ViewerE2EDriver;
+
+  beforeEach(async () => {
+    driver = new ViewerE2EDriver();
+    // You can remove this condition after viewerUrl will be updated to real viewer website.
+    const homeUrl = HOME_URL.includes('editor-flow-dev.com')
```
ranyitz, line 12: That's a cool trick right there! 🥇

packages/create-yoshi-app/templates/flow-editor/typescript/{%packagejson%}.json
```diff
@@ -61,7 +62,7 @@
     "uuid": "^3.3.2",
     "wix-ui-tpa": "^1.0.123",
     "typescript": "~3.6.0",
-    "yoshi-flow-editor-runtime": "*"
+    "yoshi-flow-editor-runtime": "^4.67.0"
```
ranyitz, line 65: I'm glad we've remembered to do this 😄

yavorsky: Yeah, would also cool to somehow update it with releasing a new version programmatically

packages/create-yoshi-app/scripts/e2e.js
```diff
@@ -141,7 +142,9 @@ const testTemplate = mockedAnswers => {
     it('step 3: should run npm test', async () => {
       console.log('running npm test...');
 
-      const testResult = await exec('npm test');
+      const testResult = await exec('npm test', {
+        VCS_BRANCH_NAME: undefined,
```
yavorsky, line 146: We need to prevent any test script thinking that we are currently in some yoshi branch

yavorsky:
@ranyitz For sure! But I like this boilerplate mostly is dynamic and being generated to for each component. And most of it is not somehow related to yoshi, but just a driver code, which would be written by users finally.

PR #2386, opened 2020-05-12
Fix website publishing on master using travis

ranyitz:
Recreate `yarn.lock` file with public registry to enable yarn installation on travis.

See the problem:
image


amitdahan:
So that's what it was!
How did you see this log? Previously I didn't see any build triggered on Travis.

ranyitz:
@amitdahan I just connected to travis and searched for "yoshi". In any case, this is a temporary solution.

PR #2390, opened 2020-05-13
Yoshi generators: fullstack use wix-i18n-config

vybu:
Use wix-i18n-config on full stack ts template.


packages/create-yoshi-app/templates/fullstack/typescript/src/client.tsx
```diff
@@ -1,26 +1,29 @@
...
 const baseURL = window.__BASEURL__;
-const locale = window.__LOCALE__;
 
 wixAxiosConfig(axios, { baseURL });
 
 const fedopsLogger = createFedopsLogger('{%projectName%}');
 
+const i18n = initI18n({
+  locale: window.__LOCALE__,
+  messages: JSON.parse(window.__MESSAGES__),
```
ranyitz, line 17: Just an idea, but it would be cool if `initI18n` could also accept a string as `messages`.

In case it identifies the `string` it can `JSON.parse` it. It should remove this small boilerplate from all future usages.

So it would result as:

```ts
const i18n = initI18n({
  locale: window.__LOCALE__,
  messages: window.__MESSAGES__,
});
```

packages/create-yoshi-app/templates/fullstack/typescript/src/server.ts
```diff
@@ -1,7 +1,13 @@
+import * as path from 'path';
 import { Router, Request } from 'express';
 import { hot } from 'bootstrap-hot-loader';
 import wixExpressCsrf from '@wix/wix-express-csrf';
 import wixExpressRequireHttps from '@wix/wix-express-require-https';
+import * as WixNodeI18nCache from 'wix-node-i18n-cache';
+
+const localI18NCache = new WixNodeI18nCache({
```
ranyitz, line 8: We try to explain things in those files using comments. In this instance, I would explain why we need this and link to the original library for further reading.

https://github.com/wix-private/wix-node-i18n-cache

packages/create-yoshi-app/templates/fullstack/typescript/src/client.tsx
ranyitz: Just to be clear, if you think that's a good idea, I would love to contribute it to `wix-i18n-config` and we surely can merge this PR and add this as a feature later on.

packages/create-yoshi-app/templates/fullstack/typescript/src/server.ts
ranyitz:
```suggestion
// caches translation files and serves them per request
// https://github.com/wix-private/wix-node-i18n-cache
const localI18NCache = new WixNodeI18nCache({
```

That's a suggestion

packages/create-yoshi-app/templates/fullstack/typescript/src/client.tsx
vybu: I think I found a better way, I changed how values are set in the template, so that there is no more need to parse JSON. As messages is evaluated as JS object directly in at document.script level.

vybu: I see you already merged, I'll add the change together with PR for js template

ranyitz: I mean, this pr already improved the template. We can surely add this in another PR

PR #2392, opened 2020-05-13
Port js universal template to use wix-i18n-config

vybu:
Part 2, to port js universal template. Also few adjustments to ts one


vybu:
What I find worrying that I had multiple mistakes/bugs ran tests and it all passed nically :confused: 

Although now to verify that all's good I just tried running generated projects myself and all seem fine.

ranyitz:
> What I find worrying that I had multiple mistakes/bugs ran tests and it all passed nically 😕

What do you mean?

We have a template test that installs the template, run lint + build + test commands.

vybu:
> > What I find worrying that I had multiple mistakes/bugs ran tests and it all passed nically confused
> 
> What do you mean?
> 
> We have a template test that installs the template, run lint + build + test commands.

I ran those tests and they passed although I had some bugs. I guess what I am saying it's still best to just check by hand - generate project and see if it actually works :)

packages/create-yoshi-app/templates/fullstack/typescript/src/components/App/App.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { withTranslation, WithTranslation } from 'react-i18next';
+import { withTranslation, WithTranslation } from '@wix/wix-i18n-config';
```
vybu, line 2: fix for issue in TS template

packages/create-yoshi-app/templates/fullstack/typescript/src/external-types.d.ts
```diff
@@ -1,7 +1,7 @@
 interface Window {
   __LOCALE__: string;
   __BASEURL__: string;
-  __MESSAGES__: string;
+  __MESSAGES__: Messages;
```
amitdahan, line 4: Where do we get this type from?

vybu: from the @wix/wix-i18n-config

amitdahan: Hmm but it seems to not be imported from anywhere, is it exposed as a global type?

vybu: https://github.com/wix-private/fed-infra/blob/1941a8556b487a563c47a6050de9a79726e42627/wix-i18n-config/src/index.d.ts#L11

PR #2403, opened 2020-05-14
Allow to set `controllerType` for component configuration

maoryosef2:
### 🔦 Summary
For platform apps, the id that you get in the `controllerConfigs` map is actually the `controllerType` which is defined on the `AppController` / `AppWidget` data.
this change will enable you to also keep the id that the widget gets from dev center, and also define the controller type that should be used in order to load the script.

This is useful for BoB widgets, since we need the dev center id to install the widget on stage in the editor script like here:
https://github.com/wix-private/bob-bootstrap/blob/master/src/editor/appInstaller.ts#L30
But the id needed in order for the correct script to load is a different id
https://github.com/wix-private/bob-bootstrap/blob/master/src/components/speakers/.component.json#L2

With this change, I could keep the id and the controllerType in `.component.json`
and this will give a better dev experience for when I want to use the dev center id.

### DIFF
--- a/packages/yoshi-flow-editor/src/model.ts
+++ b/packages/yoshi-flow-editor/src/model.ts
@@ -42,7 +42,7 @@ export interface ComponentModel {
   editorControllerFileName: string | null;
   settingsFileName: string | null;
   id: string | null;
-  controllerType?: string;
+  controllerId?: string;
 }
 
 export interface AppConfig {
@@ -52,7 +52,7 @@ export interface AppConfig {
 }
 export interface ComponentConfig {
   id: string;
-  controllerType?: string;
+  controllerId?: string;
   type?: WidgetType;
 }
 
@@ -212,7 +212,7 @@ For more info, visit http://tiny.cc/dev-center-registration`);
         editorControllerFileName,
         settingsFileName,
         id: componentConfig.id,
-        controllerType: componentConfig.controllerType,
+        controllerId: componentConfig.controllerId,
       };
       return processedModels.concat(componentModel);
     },
--- a/packages/yoshi-flow-editor/src/model.ts
+++ b/packages/yoshi-flow-editor/src/model.ts
@@ -42,6 +42,7 @@ export interface ComponentModel {
   editorControllerFileName: string | null;
   settingsFileName: string | null;
   id: string | null;
+  controllerType?: string;
 }
 
 export interface AppConfig {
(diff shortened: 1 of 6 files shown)


packages/yoshi-flow-editor/src/model.ts
```diff
@@ -38,13 +38,15 @@ export interface ComponentModel {
   editorControllerFileName: string | null;
   settingsFileName: string | null;
   id: string | null;
+  controllerType?: string;
```
yavorsky, line 41: What's the difference between controllerType and id?  I think we just converting type to id, so for users it's more clear?

yavorsky, line 41: What's the difference between controllerType and id? I think we just converting type to id, so for users it's more clear?

packages/yoshi-flow-editor/src/wrappers/templates/CommonViewerScriptEntry.ts
```diff
@@ -29,6 +30,12 @@ const importsForControllers = t<{
   }}
 `;
 
+const getControllerScriptId = (controller: TemplateControllerConfig) => {
+  const controllerScriptId = controller.controllerType || controller.id;
```
yavorsky, line 34: we can ask users to use id instead of controllerType ?

packages/yoshi-flow-editor/src/model.ts
maoryosef2: so basically the id that you are getting in the controllersConfig is the `controllerType`, so that would be the correct terminoligy (however, we don't have to stick with it in this context)

maoryosef2: meaning I am fine with `controllerTypeId` if it makes more sense to you

yavorsky: :+1: yeah, it will be much more clear as for me.

maoryosef2:
for OOI, the controllerType === devCenterId, but for platform apps its not the case, since you wont neceserraly have a devCetnerId (for platform apps its not a must)

in BoB case, we need the devCenter id for installing the widget, but not for running the code.

Changing bob widgets controllerType to match the devCenter Id will have more implications, since they are created in separate times, and we want to keep the flow as similiar as possible to app builder widgets (which are the external users version of BoB widgets)

packages/yoshi-flow-editor/src/model.ts
maoryosef2: eventually changed to `controllerId`
`controllerTypeId` make less sense

PR #2406, opened 2020-05-15
Add example BILogger to Editor flow template

yavorsky:
### 🔦 Summary
Here we are adding example BI logger to Editor Flow.

BI logger is tricky, since teams should provide own schema, which will be used for `web` and `settings` bi factories.
So here, we are going to create example logger and use it in settings. When user want to add own one, they will replace our schema initializer with their one.

We are also providing `BILoggerProvider` and `BILogger` render prop to use it across the app.
I will add docs here in a separate PR.


yavorsky:
1 hour - implementing the issue.. 5 hours - trying to deal with our CI

PR #2408, opened 2020-05-16
Add fedops logger to editor-flow template

bordi:
Add FedOps to Editor flow.

This PR inclues:
- Add `fedopsLogger`, which will be initialized udner the hood and we'll just pass `fedopsLogger` to the user's controller, which will remove some boilerplate. (TODO: ADD IT TO DOCS!)
- Add basic logger usage to the template. We will log loadStart and SSR loadEnd events.
- Add isSSR helper to the template
- generate `fedops.json` based on the project structure we have.
- implement new feature to our template:

```
{forEach%someData.someArray%forEach}doSomething in element's scope {%id%} {/forEach}
```
allowing to generate some stuff for each item in array. In our case it's each component's configuration for `fedops.json`. It's working recursively, so we can use even loops inside loops.

And welcome the new contributor @bordi which created almost everything here 👏👏👏
image

updated by @yavorsky


packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -78,11 +78,18 @@ function ooiControllerWrapper(
     setState,
   };
 
+  const { appDefinitionId } = appParams;
+  const fedopsLogger = platformAPIs.fedOpsLoggerFactory?.getLoggerForWidget({
```
yavorsky, line 82: do we have it transpiled in our bundles? 👀

bordi: Just checked. All good - consumer's build passed

yavorsky: 👏

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
```diff
@@ -28,6 +30,10 @@ function isMobile({ wixCodeApi }: IWidgetControllerConfig) {
   return wixCodeApi.window.formFactor === 'Mobile';
 }
 
+function isSSR({ wixCodeApi }: IWidgetControllerConfig): boolean {
```
ranyitz, line 33: (Not as part of this PR)

We should expose `isSSR`/`isMobile` and other utilities on our runtime so it won't be repeated in every project.

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
ranyitz: Is there a case when this method doesn't exist? (`getLoggerForWidget`) If not, we should probably update the types.

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
```diff
@@ -7,12 +7,14 @@ import {
...
 import { EXPERIMENTS_SCOPE } from '../../config/constants';
 import { getSiteTranslations } from '../../config/i18n';
+import { id as widgetId } from './.component.json';
 
 export interface ControllerContext {
   frameworkData?: any;
   appData?: IAppData;
   widgetConfig?: IWidgetConfig;
   controllerConfig: IWidgetControllerConfig;
+  fedopsLogger?: any;
```
ranyitz, line 17: I think that we can have the types - https://github.com/wix-private/fed-infra/blob/37bebf7d136cba9d752fd939612a7f98feea52ae/fedops/fedops-logger/types.d.ts#L83

I would not want everyone to `npm install` fedops and bring the types from there.

We can re-export the types so they would be available from `yoshi-flow-editor` types package (which doesn't exist yet...)

I mean, all of the `ControllerContext` should probably be available from the flow. This would eliminate the need to specify each property type on consumers projects.

(This is a general note and can be done a different PR as well)

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
yavorsky: yeah, I thought about it also.. but would be good to find more helpers and pass it to controller

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
yavorsky: I'll polish it in a separate PR

PR #2418, opened 2020-05-17
Open production BM on `yoshi-bm start`

amitdahan:
# 🔦 Summary
[Today](#2322) we open `http://localhost:5000/business-manager/{random-uuid}/{first-route}` on `yoshi-bm start`.
We use a random port since the local BM tesktit can handle non-existing `metaSiteId`s.

In the effort of working with production as much as we can, we would like to open production BM with local overrides pre-configured.

### User-defined `metaSiteId`
In production, instead of using a random uuid as the `metaSiteId`, we need the user to choose his `metaSiteId`.

I added a prompt for the user's `metaSiteId` on `yoshi-bm start`:
image

This should differ from developer to developer.
In order to persist this and remember a user's choice we chose storing this `metaSiteId` in a fs cache.
The cache should be compatible with linux, macos and Windows, by writing to the platform's AppData directory.

For example, in `multilingual`'s case, after running `yoshi-bm start` for the first time, and after responding to the prompt, the following cache will be created on macOS:

`~/Library/Caches/yoshi-flow-bm/multilingual-v2/cache.json`

```json
{
  "metaSiteId": "0fc0420c-dc6a-433b-96c3-ccc866fa84ce"
}
```

To make yoshi "un-remember", a user can run `yoshi-bm start --no-cache` to clear his cache directory.

### Localhost overrides
This also uses [Elgar](https://github.com/wix-private/elgar)'s `/redirect` endpoint to set the `statics-url` cookie to `http://localhost:3200` (hardcoded for now) and redirects back to the proper production BM url.

For example, `yoshi-bm start` would open the following URL:

```
https://apps.wix.com/elgar-server/redirect
  ?redirectToUrl=<REDIRECT_URL>
  &staticsVersions={"<ARTIFACT_ID>":"http://localhost:3200"}
```

Where `<REDIRECT_URL>` would be:

```
https://www.wix.com/dashboard/<METASITE_ID>/<FIRST_PAGE_ROUTE>
```


amitdahan:
### Next steps:
Make the [Statics Override extension](https://github.com/wix-private/chrome-statics-override) parse existing cookies (set by Elgar) properly.

This will enable a "way out" of having production BM point to your `localhost`.

packages/yoshi-flow-bm/src/cache.ts
```diff
@@ -0,0 +1,63 @@
...
+  return cache[key];
+};
+
+export const writeCache = (key: string, value: any) => {
+  const projectName = getProjectArtifactId()!;
+
+  const cachePath = path.join(
+    getCachePath('yoshi-flow-bm', projectName),
+    'cache.json',
+  );
```
ranyitz, line 48: This whole thing is duplicated with lines 26-31

packages/yoshi-flow-bm/src/scripts/start.ts
```diff
@@ -27,6 +28,7 @@ const start: CliCommand = async function(argv, config) {
       '--https': Boolean,
       '--debug': Boolean,
       '--debug-brk': Boolean,
+      '--no-cache': Boolean,
```
ranyitz, line 31: I think that is not clear that `no-cache` also invalidates the cache.

I would go with `--clear-cache` or `--cache-clear`, and also put it on the root. (So it won't be specific to the `start` command)

e.g.

```
yoshi-bm --cache-clear
```

This is a good example - https://jestjs.io/docs/en/cli#--clearcache

packages/yoshi-flow-bm/src/start-url.ts
```diff
@@ -0,0 +1,41 @@
+import prompts from 'prompts';
+import { getProjectArtifactId } from 'yoshi-helpers/build/utils';
+import { readCache, writeCache } from './cache';
+import { FlowBMModel } from './model';
+
+const readMetaSiteId = async () => {
+  const { metaSiteId } = await prompts({
+    type: 'text',
+    name: 'metaSiteId',
+    message: 'What is your preferred metaSiteId (for development)?',
```
ranyitz, line 10: It's not straight away clear what the flow is going to do with the `metaSiteId`. It worth explaining it a bit more in depth.

packages/yoshi-flow-bm/src/start-url.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  return metaSiteId as string;
+};
+
+const getMetaSiteId = async (): Promise<string> =>
+  readCache('metaSiteId') ?? readMetaSiteId();
```
ranyitz, line 17: It's a bit better to put those keys in a Constants. As we're using them in some places in the code

PR #2424, opened 2020-05-17
Update experiments usage for editor flow.

yavorsky:
### 🔦 Summary
This PR updates usage of experiments using editor flow

### Move experiments configuration to the `.application.json`
It's a good place to configure stuff. We already have `sentry` config here, `appDefinitionId`, `appName`. So `experiments: { scope: 'someScope' }` will trigger our flow to load experiments for controller and include is a part of `frameworkData`. It's not in runtime `constants.ts` anymore

### Extend `frameworkData` passed to createController to have `experiments` as a pending promise
It will be useful b/c users want to control should it or not be a blocking operation
For example, my experiment is only needed to change behavior of nested component and it's ok to render something initially w/o it, but then enable if needed.

Moreover, if user have another important promise that could be resolved in parallel with experiments promise and translations promise, he can do it with this approach istead of wating `Promise.all(experiments, somethingElseHandledByUs)` and only then resolve user's promise. 

*I'm going to extend framework data with loaded translations the same way in a separate PR*

### Update templates, so both Widget and Settings will use our `Provider`.
We'll show user's how to use `experiments.enables` in templates

### Update template tests to mock experiments data
Some simplification for experimentals testing showing how we should mock it with testkit.

Other important question here: Should we or not ask experiments scope on project generation the same we are doing for sentry.
I would add it, but I can suggest that sentry is needed for everyone and should be part of prod app from the start, but experiments could be needed only after some experimental feature introduction. Maybe we need to create own CLI command - wrapper for `petri` one, that will also update `.application.json` after addin an experiment.


packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/Settings/Settings.tsx
```diff
@@ -6,11 +6,17 @@ import {
   InjectedTranslateProps,
 } from 'react-i18next';
 import { iframeAppBiLoggerFactory } from '@wix/iframe-app-bi-logger';
+import {
+  ExperimentsProvider,
+  withExperiments,
+  InjectedExperimentsProps,
```
ranyitz, line 12: Those imports are not used
```suggestion
```

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/Widget/Widget.tsx
```diff
@@ -51,16 +51,20 @@ export default class extends React.Component<WidgetWrapperProps> {
...
-    return (
-      <div {...styles('root', {}, rest)} data-hook="CleanList-wrapper">
-        <div className={styles.header}>
-          <h2 data-hook="app-title">
-            {t('app.widget.welcome')} {appName}!
-          </h2>
+  withExperiments<WidgetProps>(
+    ({ appName, t, experiments, ...rest }: WidgetProps) => {
+      return (
+        <div {...styles('root', {}, rest)} data-hook="{%name%}-wrapper">
```
ranyitz, line 57: 🌟

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/Widget/Widget.tsx
```diff
@@ -51,16 +51,20 @@ export default class extends React.Component<WidgetWrapperProps> {
...
+          <div className={styles.header}>
+            <h2 data-hook="app-title">
+              {t('app.widget.welcome')} {appName}!
+            </h2>
+          </div>
+          {experiments.enabled('specs.test.ShouldShowButton') ? (
```
ranyitz, line 63: Maybe we should add a comment that this is a dummy experiment?

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
```diff
@@ -1,16 +1,15 @@
...
   IWidgetConfig,
   IAppData,
 } from '@wix/native-components-infra/dist/src/types/types';
+import { FrameworkData } from 'yoshi-flow-editor-runtime/build/types';
 import { appName } from '../../../.application.json';
-import { EXPERIMENTS_SCOPE } from '../../config/constants';
 import { getSiteTranslations } from '../../config/i18n';
 import { id as widgetId } from './.component.json';
 
 export interface ControllerContext {
```
ranyitz, line 11: We can provide all of the controller context type from our runtime.

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
```diff
@@ -34,23 +33,16 @@ function isSSR({ wixCodeApi }: IWidgetControllerConfig): boolean {
...
   controllerConfig,
   fedopsLogger,
+  frameworkData,
 }: ControllerContext) {
   const { appParams, setProps } = controllerConfig;
   const language = getSiteLanguage(controllerConfig);
   const mobile = isMobile(controllerConfig);
   const [experiments, translations] = await Promise.all([
-    getExperimentsByScope(EXPERIMENTS_SCOPE),
+    frameworkData.experiments,
```
ranyitz, line 45: That's quite confusing as an API. Maybe we should call it `experimentsPromise` or something.

Also, if we don't pass the experiments ourselves in the controller (in userland) would they still be available on the context?

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
```diff
@@ -51,13 +53,22 @@ const createEditorAppForWixSDK = ({
     isEditor: true,
   });
 
+  console.log(
+    experimentsConfig,
+    'experimentsConfigexperimentsConfigexperimentsConfig',
+  );
+
```
ranyitz, line 60:
```suggestion
```

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -97,12 +93,12 @@ function ooiControllerWrapper(
...
         ...userController,
         pageReady: async (...args: Array<any>) => {
-          const awaitedFrameworkData = await objectPromiseAll(frameworkData);
+          // const awaitedFrameworkData = await objectPromiseAll(frameworkData);
 
           // TODO: export by property (methods, state) so we won't have conflicting props
           setProps({
             __publicData__: controllerConfig.config.publicData,
-            ...awaitedFrameworkData,
+            // ...awaitedFrameworkData,
```
ranyitz, line 101: This used to make sure you get the experiments on the context (to your widget/react code) even if the user doesn't need them in the controller.

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
yavorsky: > That's quite confusing as an API. Maybe we should call it experimentsPromise or something.

I think it's typed and should be warned by ts that's it is `Promise<ExperimentsBag>`. I don't know why, but somethingPromise is sounds a bit weird to me 🙂 We don't use this naming much in the project when working with promises.
But we can quickly update it. 

> Also, if we don't pass the experiments ourselves in the controller (in userland) would they still be available on the context?

No, it's different runtime scopes and bundles, so the only pass something to widget is to use `setProps`. Since we controll it, we can pass `setProps`, but it could be unexpected for users who want to control performance and prevent additional `setProps`.

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
yavorsky: This is used only for experiments now.
The basic idea from what I understood is to await all set of promises before calling createControllers.
But sometimes we don't want to block it until experiments will be loaded, since users should control what they need to block widget rendering and passing props here.
For example:
-> we are waiting for all frameworkData to be loaded
-> user is awaiting more stuff (fetch widget data, fetch additional data (translations for now))
-> render

User might understand that experiments could be fetched in parallel with translations and widget data, so can control should be under `Promise.all` or not:
-> Promise.all([await frameworkData.experiments, await frameworkData.somethingElse, loadSomething()]) in `createController`
-> render Widget

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/Widget/Widget.tsx
yavorsky: 👍

yavorsky:
> Good work, I added a few inline comments. Regarding frameworkData I think that by now we can call it flowData as devs relates us as the flow more than the framework.

Great idea!

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
yavorsky: good catch

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
```diff
@@ -1,21 +1,9 @@
-import {
-  IWidgetControllerConfig,
-  IWidgetConfig,
-  IAppData,
-} from '@wix/native-components-infra/dist/src/types/types';
-import { FlowData } from 'yoshi-flow-editor-runtime/build/types';
+import { IWidgetControllerConfig } from '@wix/native-components-infra/dist/src/types/types';
+import { ControllerContext } from 'yoshi-flow-editor-runtime/build/types';
```
ranyitz, line 2: ❤️ 

Can't we just expose the types under `types` property? (in `yoshi-flow-editor-runtime`)

So we'll have:

```
import { ControllerContext } from 'yoshi-flow-editor-runtime';
```

Again, this can be part of another PR, just an idea

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
```diff
@@ -42,7 +30,7 @@ async function createController({
   const language = getSiteLanguage(controllerConfig);
   const mobile = isMobile(controllerConfig);
   const [experiments, translations] = await Promise.all([
-    flowData.experiments,
+    flowData.getExperiments(),
```
ranyitz, line 33: cool cool cool

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -108,7 +107,7 @@ function ooiControllerWrapper(
 
           // Optional `pageReady`
           if (userController.pageReady) {
-            return userController.pageReady(setProps, ...args);
+            return userController.pageReady(...args);
```
ranyitz, line 110: I think that this is a breaking API change. Is this intentional?

packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/controller.ts
yavorsky: Yeah, I wanted to polish it and refactor in a separate PR related to moving all reusable styles to this place

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
yavorsky: Actually this change reverts breaking API change. I don't think it's useful and we shouldn't add custom first prop by getting off platform API

PR #2425, opened 2020-05-17
Create temporary pkg.json for sentry configuration

yavorsky:
### 🔦 Summary
Hotfix for sentry generation. Currently it's failing. Works fine for `create-yoshi-app:dev`, since `cwd` for this process is `yoshi` one.

In this file https://github.com/wix-private/fed-infra/blob/master/add-sentry/lib/project-utils.js#L8 this trying to load package.json in module runtime outside of tasks. This process is happening before we generated a project, so it's failed.

Proposed hotfix is to generate temporary `package.json` and remove it immediately after `add-sentry` modules will be initialized.


ranyitz:
@yavorsky I think it worth opening an issue to https://github.com/wix-private/fed-infra/blob/master/add-sentry/lib/project-utils.js#L8 

Doing things like that during the evaluation of a file is a bad thing in general. They would surly accept a Fix. And we add a comment with a link to the issue until in our codebase.

yavorsky:
@ranyitz totally agree!

yavorsky:
https://github.com/wix-private/fed-infra/issues/1937

PR #2434, opened 2020-05-18
allow to always override SURICATE_TUNNEL_ID

cowchimp:
### 🔦 Summary
If we're using Yoshi with Suricate in the CI the created endpoint will have a generic route:
e.g. `https://bo.wix.com/suricate/tunnel/buildserver.thunderbolt-elements-dev-server/`
This means that two devs or two CI builds running at the same time can use the same route.
To avoid this it should always be possible to reserve a private route for yourself using the `SURICATE_TUNNEL_ID` environment.
Right now Yoshi exposes overriding the tunnelId using this environment variable but because of the order of `if`s it can be ignored.
This PR gives the environment variable the highest priority.

### 🗺️ Test Plan
...


cowchimp:
thanks @ranyitz

PR #2440, opened 2020-05-20
Replace ooi template with flow editor

yavorsky:
### 🔦 Summary
Replace ooi template with editor flow - Out of iframe.
Updated fallback timeout to sign in from 20s to 60.


package.json
```diff
@@ -50,7 +50,6 @@
     "test:templates:library": "npm run library --prefix ./packages/create-yoshi-app",
     "test:templates:flow-library": "npm run flow-library --prefix ./packages/create-yoshi-app",
     "test:templates:bm": "npm run bm --prefix ./packages/create-yoshi-app",
-    "test:templates:ooi": "npm run ooi --prefix ./packages/create-yoshi-app",
```
ranyitz, line 53: That's not critical for now, but we may want to rearrange the CI so we'll have separate tests for platform-app template and OOI template

packages/create-yoshi-app/src/templates.ts
```diff
@@ -46,14 +41,12 @@ if (process.env.EXPERIMENTAL_FLOW_BM === 'true') {
...
-    path: toTemplatePath('flow-editor'),
-    language: ['typescript'],
-  });
-}
+templates.push({
+  name: OOI_TEMPLATE_NAME,
+  title: 'flow-editor - Out of iFrame',
+  path: toTemplatePath('flow-editor'),
+  language: ['typescript'],
+});
```
ranyitz, line 49: Not critical, but we can move it to be part of templates array.

package.json
yavorsky: 👍

PR #2443, opened 2020-05-20
Type `serve.d.ts` correctly

Schniz:
### 🔦 Summary

Typing `module.exports` as `default` is wrong

Usage in Thunderbolt with `@ts-check` leads to an error:

image

### 🗺️ Test Plan

I changed it locally and it works:

image

### DIFF
--- a/packages/yoshi-flow-monorepo/serve.d.ts
+++ b/packages/yoshi-flow-monorepo/serve.d.ts
@@ -1,2 +1,3 @@
 import serve from './build/serve.d';
+
 export = serve;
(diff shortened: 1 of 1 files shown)


amitdahan:
> Typing module.exports as default is wrong

Why? Because it then depends on whether the consuming project has `esModuleInterop`? While this is agnostic? (I may be off)

Schniz:
> > Typing module.exports as default is wrong
> 
> Why? Because it then depends on whether the consuming project has `esModuleInterop`? While this is agnostic? (I may be off)

AFAIK `esModuleInterop` works for `import` statements, but in Thunderbolt we use `require` which won't do the interop

Exporting `default` requires doing something like:

```js
const { default: serve } = require('yoshi-monorepo-flow/serve');
```

but as we both know, it ain't gonna work!

Schniz:
@ranyitz lets merge :shipit: 🙏

PR #2445, opened 2020-05-20
Yoshi Server: error toast message

yanivefraim:
We want to display an informative error modal in case of a server function error:

Kapture 2020-05-20 at 17 16 34


packages/yoshi-server-client/src/index.ts
```diff
@@ -56,9 +56,11 @@ export default class implements HttpClient {
         if (res.headers.get('content-type')?.includes('application/json')) {
           const error = await res.json();
           if (process.env.NODE_ENV !== 'production') {
+            if (process.env.BROWSER !== 'none') {
```
ronami, line 59: I think it's process.env.browser (lowercase): https://github.com/wix/yoshi/blob/master/packages/yoshi-common/src/webpack.config.ts#L735.

packages/yoshi-server-client/src/utils.ts
```diff
@@ -1,3 +1,77 @@
...
   return baseUrl.replace(/\/+$/, '') + '/' + relativeUrl.replace(/^\/+/, '');
 }
+
+export function showToast<
+  Result extends FunctionResult,
+  Args extends FunctionArgs
+>(error: any, method: DSL<Result, Args>, args: any, baseUrl: string) {
+  const Toastify = require('toastify-js');
+  const css = createCss();
+  document.body.appendChild(css);
```
ronami, line 16: can we append css to the document just once? I think it's the same CSS for all errors/toasts.

packages/yoshi-server-client/src/index.ts
yanivefraim: Thanks, good catch

packages/yoshi-server-client/src/utils.ts
yanivefraim: sure, fixed

packages/yoshi-server-client/src/utils.ts
```diff
@@ -1,3 +1,81 @@
...
+          credentials: 'same-origin',
+          method: 'GET',
+        },
+      );
+    },
+  }).showToast();
+}
+
+function createCss() {
+  const toastCloseClass = '.toast-close';
```
ronami, line 46: I think this is a mistake

yanivefraim: fixed

PR #2446, opened 2020-05-20
fix umdNamedDefine config property

tuchk4:
Seems like property exists at docs and at the [noramlize](https://github.com/wix/yoshi/blob/master/packages/yoshi-config/src/normalize.ts) but was not passed to the webpack config


packages/yoshi-flow-bm/src/webpack.config.ts
```diff
@@ -28,6 +28,7 @@ const createDefaultOptions = (config: Config) => {
     useAngular: config.isAngularProject,
     devServerUrl: config.servers.cdn.url,
     separateCss,
+    umdNamedDefine: config.umdNamedDefine,
```
ranyitz, line 31:
```suggestion
```

packages/yoshi-flow-editor/src/webpack.config.ts
```diff
@@ -18,6 +18,7 @@ const createDefaultOptions = (config: Config) => {
     separateStylableCss: true,
     cssModules: true,
     enhancedTpaStyle: true,
+    umdNamedDefine: config.umdNamedDefine,
```
ranyitz, line 21:
```suggestion
```

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -346,6 +346,7 @@ export function createBaseWebpackConfig({
   forceMinimizeServer = false,
   useAbsoluteUrlsForCssAssets = false,
   serverExternals,
+  umdNamedDefine = true,
```
ranyitz, line 349: I guess that despite [what's the docs says](https://wix.github.io/yoshi/docs/yoshi-config#umdnameddefine), it was configured to `false` for a long time now. I prefer to leave it as is (`false`) and change the docs instead of changing the default.

website/versioned_docs/version-3.x/api/configuration.md
```diff
@@ -258,7 +258,7 @@ Next time you'll run `yoshi lint`, this command will execute and only then the l
 `Boolean`
 
 If option is `true` AMD modules of the UMD build will have names. Otherwise an anonymous define is used(the same as in [webpack](https://webpack.js.org/configuration/output/#output-umdnameddefine)).
-By default it is `true`.
+By default it is `false`.
```
tuchk4: I'll revert, I think we should change only actual docs.

website/versioned_docs/version-4.x/yoshi-config.md
```diff
@@ -290,4 +290,4 @@ Next time you'll run `yoshi lint`, this command will execute and only then the l
 `Boolean`
 
 If option is `true` AMD modules of the UMD build will have names. Otherwise an anonymous define is used(the same as in [webpack](https://webpack.js.org/configuration/output/#output-umdnameddefine)).
-By default it is `true`.
+By default it is `false`.
```
tuchk4: same here

PR #2449, opened 2020-05-20
Pass `dataCallback` to editor sentry

yavorsky:
### 🔦 Summary
We need data to be returned and they are making it with data callback


yavorsky:
It's kinda urgent to me, I'm going to merge it. We can discuss here and update tomorrow if it isn't looking good.

PR #2450, opened 2020-05-21
Yoshi Server: fix request/response types

yanivefraim:
Currently, `this.req`/`this.res` points to Express' `Request/Response`.
We want TS users to be able to do:
`this.req.aspects`. 

(currently, they have TS error for that)


packages/yoshi-server/src/types.ts
```diff
@@ -19,7 +20,7 @@ export type FunctionResult = OptionalPromise<any>;
 
 // Server function types
 export type FunctionContext = {
-  req: Request;
+  req: Request & WithAspects;
   res: Response;
   context: BootstrapContext;
```
ronami, line 25: Super tiny, maybe later on we can extract common types between function/route to another type

PR #2453, opened 2020-05-21
[docs] adding docs for assets.md

sidoruk-sv:
### 🔦 Summary

Based on https://wix.github.io/yoshi/docs/legacy-guides/assets/

Additions motivated and authored by `Ran Yitzhaki` in the thread:
https://wix.slack.com/archives/CAL591CDV/p1582101124185300


website/docs/styles-and-assets/assets.md
```diff
@@ -0,0 +1,14 @@
+---
+id: assets
+title: Assets
+sidebar_label: Assets
+---
+
+There are few ways of using asset files inside client application:
+
+- Install through npm and import to your bundle;
+- Create `src/assets` or `app/assets` folder. Every assets inside the folder will be copied to the `dist` folder and can then being served from the CDN;
```
ranyitz, line 10:
```suggestion
```

The first part is not adding important information here.
The second part only works in the legacy flow, and already [added as such](https://wix-yoshi-website-pr-2453.surge.sh/docs/legacy-flow#assets-folder)

website/docs/styles-and-assets/assets.md
```diff
@@ -0,0 +1,14 @@
...
+- import files (e.g.: `.png`, `.jpg`) from .ts/.js files. It should go through one of the loaders. Therefore it will result as either base64 DataURL, or a file controlled by webpack:
```
ranyitz, line 11:
```suggestion
- When importing assets (e.g. `.png`/`.jpg`) from `.ts`/`.js` files they go through one of the loaders. Therefore it will result as either base64 DataURL, or a file controlled by Webpack:
```

website/docs/styles-and-assets/assets.md
```diff
@@ -0,0 +1,14 @@
+---
+id: assets
+title: Assets
+sidebar_label: Assets
+---
+
+There are few ways of using asset files inside client application:
```
ranyitz, line 7: I would change that a bit if we're going to remove some of the bullets below

website/docs/styles-and-assets/assets.md
ranyitz: A code example can improve this article as well

sidoruk-sv:
@ranyitz am I correct that CRA docs could be reused here (almost without changes)?
Removed paragraphs about customizing the limits and all other info is relevant for both tools.

website/docs/styles-and-assets/assets.md
```diff
@@ -0,0 +1,39 @@
...
+
+To reduce the number of requests to the server, importing images that are less than 10,000 bytes returns a [data URI](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) instead of a path. This applies to the following file extensions: bmp, gif, jpg, jpeg, and png.
+
+Here is an example:
+
+```js
+import React from "react";
+import logo from "./logo.png"; // Tell webpack this JS file uses this image
+
+console.log(logo); // /logo.84287d09.png
```
ranyitz, line 19:
```suggestion
console.log(logo); // /media/logo.ea74b57b.png
```

It would result in the `media` directory.

website/versioned_docs/version-4.x/styles-and-assets/assets.md
```diff
@@ -0,0 +1,39 @@
...
+
+This works in CSS too:
+
+```css
+.Logo {
+  background-image: url(./logo.png);
+}
+```
+
+webpack finds all relative module references in CSS (they start with `./`) and replaces them with the final paths from the compiled bundle. If you make a typo or accidentally delete an important file, you will see a compilation error, like when you import a non-existent JavaScript module. The final filenames in the compiled bundle are generated by webpack from content hashes. If the file content changes in the future, webpack will give it a different name in production so you don’t need to worry about long-term caching of assets.
```
ranyitz, line 39: This is not entirely true as normally we are publishing each version under a different base url (according to the version). We do have a way to solve this which we want to promote, but this is not the case for most projects.

```suggestion
webpack finds all relative module references in CSS (they start with `./`) and replaces them with the final paths from the compiled bundle. If you make a typo or accidentally delete an important file, you will see a compilation error, like when you import a non-existent JavaScript module. The final filenames in the compiled bundle are generated by webpack from content hashes.
```

ranyitz:
@sidoruk-sv Note that you don't need to modify the "version-4.x" parts as they would be auto generated during next version release. Also, regarding the side-bar, please modify the regular sidebar and not the version 4 sidebar.

sidoruk-sv:
Thanks, didn't mention that there were `sidebars.js` at the root

Need to add some notes about contributing to documentation to [CONTRIBUTING.md](https://github.com/wix/yoshi/blob/master/CONTRIBUTING.md)))

ranyitz:
@sidoruk-sv You right, the contributing should be updated.

PR #2454, opened 2020-05-22
Fix editor script for bob apps.

yavorsky:
### 🔦 Summary
Fix an editor script for platform apps, when user's `editor.app.ts` returns a function.

Currently, we are trying to override editorReady, but in fact for platform apps it's removing other fields since it's a `Function`.


packages/yoshi-flow-editor/src/wrappers/templates/CommonEditorScriptEntry.ts
```diff
@@ -46,6 +46,7 @@ export default t<Opts>`
     )
     .join('\n  ')}
   editorScriptEntry = builder.build();
+  module.exports = editorScriptEntry.default || editorScriptEntry;
```
maoryosef2, line 49: this check is redundant, since `.build()` will never return an object with `default`

What we can do however, is if the original editorScriptEntry (before you reassigend it with the builder.build result)
then return `default` if it exists

```suggestion
if (editorScriptEntry.default) {
  module.exports = editorScriptEntry.default;
} else {
  builder = editorScriptBuilder()
  //...
    
  module.exports = builder.build()
}
```

yavorsky: Got it :+1:

PR #2458, opened 2020-05-24
Start BM loader during first render

amitdahan:
### 🔦 Summary
After #2363, we converted the root component to a function component, but we didn't have a 1:1 replacement for invoking the "show loader" side-effect in the constructor so instead we used `useEffect`.

This made the child component finish mounting before the parent, and thus switching the order of the loader calls (finished -> started, instead of correctly, started -> finished).

This fixes that by using `useMemo` instead (similar to my `useOnce` implementation except it's built-in).


ranyitz:
Also, not super critical, but it deserves a comment above.

amitdahan:
> Maybe I'm missing something, but why we even need the `notifyViewStartLoading` function?

Docs say so:
https://github.com/wix-private/business-manager/blob/master/docs/pageComponents.md#notify-viewstartloading-and-viewfinishedloading

In BM Flow we call it for you.

PR #2463, opened 2020-05-24
Allow overriding the module bundle name

amitdahan:
### 🔦 Summary
We output `module.bundle.js` by default, and there's no way to change that.
Since I rather not touch users' BM integration (which is the point where BM refers to said `module.bundle.js`) during migrations - I think it's best to allow overriding this.

Added a new config field to `.module.json` named `moduleEntry` which will default to `module` to not break existing users (singular actually).


amitdahan:
@ranyitz Sounds good!

PR #2465, opened 2020-05-24
Yoshi Server: copy Dockerfile

yanivefraim:
On CI, we want to create the Dockerfile in case of a Yoshi Server.


packages/yoshi-common/src/copy-docker.ts
```diff
@@ -0,0 +1,14 @@
+import path from 'path';
+import fs from 'fs-extra';
+import { Config } from 'yoshi-config/build/config';
+
+export = async function async(config: Config, cwd = process.cwd()) {
```
ronami, line 5: why not use `export default`?

packages/yoshi-common/src/copy-docker.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (!config.yoshiServer) {
+    return;
+  }
+  const projectDockerExists = fs.existsSync(path.resolve(cwd, 'Dockerfile'));
```
ronami, line 9: I think we can add `Dockerfile` to https://github.com/wix/yoshi/blob/master/packages/yoshi-config/src/paths.ts.

packages/yoshi-common/src/templates/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM docker-repo.wixpress.com/wix-bootstrap-onbuild:stable
+MAINTAINER {%authorName%} <{%authorEmail%}>
```
ronami, line 2: I think we'll need to fill these as we copy it to a project, normally it happens when the developer generates a project with CYA.

packages/yoshi-flow-app/src/scripts/build.ts
```diff
@@ -79,6 +80,7 @@ const build: cliCommand = async function(argv, config) {
         clientProjectName: config.clientProjectName,
         staticsDir: config.clientFilesPath,
       }),
+      copyDocker.default(config),
```
ronami, line 83: we can use `export default` and then we won't have to use `.default`.

packages/yoshi-common/src/copy-docker.ts
yanivefraim: fixed, thanks

packages/yoshi-common/src/templates/Dockerfile
yanivefraim: nice catch, done. 10x!

PR #2466, opened 2020-05-25
Add correct types version to flow-bm's comsmiconfig

amitdahan:
### 🔦 Summary
This explicitly pins `@types/cosmiconfig` which yoshi-bm needs.


yarn.lock
```diff
@@ -4899,7 +4906,7 @@
     "@wix/wix-testkit-base" latest
     cors "^2.8.5"
 
-"@wix/wix-express-aspects@~1.0.843":
+"@wix/wix-express-aspects@^1.0.0", "@wix/wix-express-aspects@~1.0.843":
```
amitdahan, line 4909: This was a result of an un-saved `yarn.lock` change, which maybe happened due to a difference in yarn versions (our CI, Renovate's, and local versions?)

PR #2467, opened 2020-05-25
Emit metadata file with data on how to load each entry separately

ronami:
### Summary

This PR adds a Webpack plugin that emits a JSON file with data on which files need to be loaded to render an entry. This is going to be used initially by TB to load different Webpack entries in different parts of the HTML.

The metadata file looks like the following:

```json
{
  "entries": {
    "main": [
      "webpack-runtime.bundle.js",
      "common.chunk.js",
      "main.chunk.css",
      "main.chunk.js"
    ],
    "other.inline": [
      "webpack-runtime.bundle.js",
      "common.inline.chunk.js",
      "bi.chunk.js"
    ]
  },
  "publicPath": "https://wix.com/"
}
``` 

Every key (entry) lists all of the files that needs to be loaded to run that entry. Common files will be listed in every entry and it's about the runtime code that uses it to make sure files don't get written more than once.

I also included the `publicPath` so it's easier for the server that renders it to prepend it if files needed to be fetched from the CDN rather than inlined.


packages/yoshi-common/src/html-rendering-data-plugin.ts
```diff
@@ -0,0 +1,33 @@
...
+
+            return {
+              ...acc,
+              [entryName]: filesNames,
+            };
+          }, {}),
+          publicPath: compiler.options.output?.publicPath,
+        });
+
+        compilation.assets['html-rendering-data.json'] = {
```
ranyitz, line 26: We have more than one web build (debug/production) will this be created only for production?

ronami: Good catch, thanks! 👍

PR #2472, opened 2020-05-25
Site Assets web compilation for running site-assets code in the browser

ranyitz:
### 🔦 Summary
This PR adds another site-assets compilation which we transpile for target `web` and can run in the browser.

This is a PR that adds this compilation to the thunderbolt build. I needed to modify a few things and take some decisions.

One of the complex parts here was to tranpile carmi's output. The result solution was to transpile it using `babel-loader` specifically as this is not raising the overall build time by too much. Carmi files are usually "ts" files and therefore went to `ts-loader` by default which took a lot of time to transpile those files.

1. The progress bars now looks like that:
> The important thing is that we distinguish between `web` and `node` compilations for site-assets.

image

2. The build message in the end now looks like that:

> There are added headers for both **Site Assets (web)** and **Site Assets (node)** so it will be clear to understand which artifacts belongs to which compilations.

image

3. By the suggestion of @ronami I added the `noParse` option for carmi modules as they return a blob of code without `require`/`import` calls. I couldn't see a major change in build time, but I checked on my machine when the build times varies. I'm sure that this it generally a good change.

4. I deleted some possible configurations that wasn't used. We'll add it if needed.

____________

This compilation was added to start as well

image

It adds around 10seconds give or take to the start compilation.

____________________

We also make sure that we transpile Carmi's output only on site-assets web compilation and that we don't do that during `yoshi start` when working locally on latest chrome. This should make sure the build time is kept to its minimum.


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -103,6 +103,7 @@ function getProgressBarInfo(
   function getObject() {
     switch (configName) {
       case 'client':
+      case 'site-assets':
```
ranyitz, line 106: We should print something specific for site-assets server build which is not directed for the browser/client and can be confusing.

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -987,12 +986,27 @@ export function createBaseWebpackConfig({
...
+          test: /\.carmi.(js|ts)$/,
+          include: includeInTranspilation,
+          use: [
+            {
+              loader: 'babel-loader',
+              options: {
+                ...babelConfig,
+                // Carmi files are quite big and this option
+                // prevents from a warning to be showing during the build
+                // https://babeljs.io/docs/en/options#compact
```
ranyitz, line 999: If you're interested, this it the error - https://stackoverflow.com/questions/29576341/what-does-the-code-generator-has-deoptimised-the-styling-of-some-file-as-it-e

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -987,12 +986,27 @@ export function createBaseWebpackConfig({
...
+              },
+            },
+          ],
+        },
         {
           test: reScript,
           // Don't transpile the output of Carmi with Babel/TypeScript
           // https://github.com/wix/yoshi/pull/2227
-          exclude: /\.carmi.(js|ts)$/,
+          exclude: /\.carmi.js$/,
```
ranyitz, line 1009: the reScript includes only `js?` files which doesn't include `ts`

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -335,10 +337,7 @@ export function createSiteAssetsWebpackConfig(
   const config = createBaseWebpackConfig({
     cwd: pkg.location,
     configName: 'site-assets',
-    target: 'node',
-    // We don't have any server externals for `site assets` bundle
-    // So with empty object, we'll be sure that no default externals value will be applied
-    serverExternals: {},
+    target,
```
ranyitz, line 340: we probably need to add `serverExternals: {}` for node. Or at least check that it still works as expected

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -353,18 +352,6 @@ export function createSiteAssetsWebpackConfig(
   });
 
   config.entry = isSingleEntry(entry) ? { app: entry as string } : entry;
```
ranyitz, line 354: I don't think the following are being used so I removed them

packages/yoshi-flow-monorepo/src/scripts/build.ts
```diff
@@ -257,21 +272,13 @@ const build: cliCommand = async function(argv, rootConfig, { apps, libs }) {
...
-      const [serverStats, siteAssetsStats] = getAppData(pkg.name).stats;
+    const [clientStats, clientOptimizedStats] = getAppData(pkg.name).stats;
 
-      printBuildResult({
-        webpackStats: [siteAssetsStats, serverStats],
-        cwd: pkg.location,
-      });
-    } else {
-      const [, clientOptimizedStats, serverStats] = getAppData(pkg.name).stats;
+    printBuildResult({
```
ranyitz, line 277: We need to support the regular case as well

packages/yoshi-common/src/webpack.config.ts
ranyitz: done

packages/yoshi-flow-monorepo/src/scripts/build.ts
ranyitz: done

packages/yoshi-flow-monorepo/src/webpack.config.ts
ranyitz: done

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -395,10 +395,12 @@ export default class DevEnvironment {
     yoshiServer = false,
   }: {
     webpackConfigs: [
-      webpack.Configuration?,
-      webpack.Configuration?,
-      webpack.Configuration?,
-      webpack.Configuration?,
+      webpack.Configuration?, // Main client config
```
ranyitz, line 398: I added some comments here.

We should start thinking on the next phase of this abstraction.

each compilations here plays a specific role (with or without hot updates and so on)

I think that it's currently not clear enough. I added those comments to explain what is the role of each configuration

packages/yoshi-flow-monorepo/src/scripts/start.ts
```diff
@@ -163,10 +171,12 @@ const start: cliCommand = async function(argv, rootConfig, { apps, libs }) {
 
   const devEnvironment = await DevEnvironment.create({
     webpackConfigs: [
-      clientOrSiteAssetsConfig,
+      clientConfig,
```
ranyitz, line 174: The client config is different than the site-assets. For example you would not like `hmr` update to be on `siteAssetsConfigNode`.

This is a change with the current behavior, but it makes sense to me.

PR #2474, opened 2020-05-26
Remove yoshi config from the editor flow templates

yavorsky:
### 🔦 Summary
Currently we are forcing people to use `yoshi.config.js` by having it in template.

Current yoshi config solution is bad b/c mix a lot of stuff we don't need at all.
`https://wix.github.io/yoshi/docs/yoshi-config`

We just using few of these options. If we need more, we can add it to `.application.json` or even editor-flow.config.js or something like this with own schema, validation rules and template.


packages/yoshi-flow-editor/src/model.ts
```diff
@@ -81,10 +88,14 @@ function resolveFileNamesFromDirectory(dir: string, fileNames: Array<string>) {
...
-  path: string,
+function getLocalConfig<C extends AppConfig | ComponentConfig | URLsConfig>(
+  configPath: string,
 ): C {
-  return JSON.parse(fs.readFileSync(path, 'utf8'));
+  const extension = path.extname(configPath);
+  if (extension === '.js') {
+    return require(configPath);
+  }
+  return JSON.parse(fs.readFileSync(configPath, 'utf8'));
```
ranyitz, line 98: This is not urgent, but it would be good if we could handle the config loading with validations and normalization.

See the bm configs which I think constructed in a good way - https://github.com/wix/yoshi/blob/master/packages/yoshi-flow-bm/src/config/index.ts

packages/yoshi-flow-editor/src/utils.ts
```diff
@@ -1,21 +1,22 @@
...
 
 export const joinDirs = (...dirs: Array<string>) =>
   path.join(process.cwd(), ...dirs);
 
-export const normalizeStartUrlOption = (urls: StartUrl): Array<string> => {
-  if (Array.isArray(urls)) {
-    return urls;
-  } else if (typeof urls === 'string') {
-    return [urls];
+export const normalizeStartUrlOption = (urls: URLsConfig): Array<string> => {
```
ranyitz, line 11: This can be written as a "private" function of a normalize function for a config.

packages/yoshi-flow-editor/src/cli.ts
```diff
@@ -91,11 +91,6 @@ Promise.resolve().then(async () => {
   }
 
   const config = loadConfig();
-  // This line is because the default ssl config is false,
-  // and since we use yoshi-flow-legacy test command,
-  // we need to configure ssl to true becase we did implement build and start commands
-  // with ssl true as default (this should be removed when test cmd is implemented)
-  config.servers.cdn.ssl = true;
```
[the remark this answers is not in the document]
yavorsky: we'll have to override the same thing in 3 different places

packages/yoshi-flow-editor/src/utils.ts
yavorsky: I think it's not related to current changes, but we can create keep it in mind

yavorsky:
I think in general using `loadConfig()` in the global scope is not a good practice.
Currently, if we want to get rid of config, we need to find all places where we are using both approaches:
`loadConfig()`and `require('yoshi-config')`, which is calling it under the hood.
and in this places we need to directly inject ssl configuration we want.

For example, jest-yoshi-preset, which can't be configured inside of the flow, so every flow will go here and update something.

Even after updating it, we are still have certificate error on editor side, so I predict here we have more places where config is being just loaded internally, so we'll need to create more overrides in the place.

packages/yoshi-config/src/normalize.ts
```diff
@@ -2,17 +2,24 @@ import { PackageJson } from 'read-pkg';
 import { Config, InitialConfig } from './config';
 import { multipleModules, singleModule } from './globs';
 
+const FLOWS_WITH_SSL = ['yoshi-flow-editor'];
```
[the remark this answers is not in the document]
yavorsky: > I don't want to configure all of the flows from yoshi config, I prefer instead to be able and configure it from outside.

Since we are using `loadConfig` from some internal modules, overriding the result could be uncontrolled and will be spread for the whole monorepo

yavorsky:
After diving to deep, I decided to simplify approach: https://github.com/wix/yoshi/pull/2487

PR #2478, opened 2020-05-26
Update module_{%PROJECT_NAME%}.json

naamagr:
Added appDefId to module config file as it is now required by business-manager.

### DIFF
--- a/packages/create-yoshi-app/templates/business-manager-module/javascript/app-config-templates/module_{%PROJECT_NAME%}.json
+++ b/packages/create-yoshi-app/templates/business-manager-module/javascript/app-config-templates/module_{%PROJECT_NAME%}.json
@@ -1,6 +1,7 @@
 {
   "moduleId": "{%PROJECT_NAME%}",
   "mainPageComponentId": "{%projectName%}",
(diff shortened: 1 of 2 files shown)


amitdahan:
Hey hey! 

2 questions:

1. What's BM gonna do with a non-existing `appDefId` like this one?
2. Is it possible to have 2 configs with the same `appDefId`?

naamagr:
Hey :)
1. Not sure what you're referring to when you say "this one", but in a few days we're going to ask every module to add an appDefId if it doesn't have it already.
2. No it's not possible, should be unique per module.

amitdahan:
This one = `00000000-0000-0000-0000-000000000000`

amitdahan:
So it'll only work because in the BM testkit environment, that's the "only  module" with this `appDefId`?

naamagr:
Right, and because we don't have that validation just yet 😉 
However, I'm not sure about this template change. It might cause more confusion than help... 
So I'll discuss it with my team and close the PR for now.
Sorry for the hassle...

amitdahan:
NP!

PR #2480, opened 2020-05-26
Remove sentry from dev/sites

yavorsky:
### 🔦 Summary
This PR just moves sentry url from sites.js to .application.json, so we can use it in future for logging


packages/create-yoshi-app/templates/flow-editor/typescript/.application.json
```diff
@@ -5,7 +5,8 @@
     "DSN": "{%sentryData.DSN%}",
     "id": "{%sentryData.id%}",
     "projectName": "{%sentryData.projectName%}",
-    "teamName": "{%sentryData.teamName%}"
+    "teamName": "{%sentryData.teamName%}",
+    "webPage": "https://sentry.wixpress.com/sentry/{%sentryData.projectName%}"
```
amitdahan, line 9: What happens if the user changes the `webPage` or `teamName` fields?

yavorsky: We'll pass updated config to the platform's sentry monitoring.
And it won't affect much, since from what I understand only DSN is important for runtime.

amitdahan: So why even keep it in the config?
Why not just have the DSN?

yavorsky: We need it to display during the build

PR #2482, opened 2020-05-27
Use default config for editor flow jest. Remove ssr files

yavorsky:
### 🔦 Summary
This PR removes ssr files and jest configuration files from the projects. We are going to handle it by default.

SSR testing should take place on sled side.
We still need to test settings via e2e, so we will use default config handled by `defaults` directory. The plan is to remove it if we can test it on real sites via editor part.


yavorsky:
Cool. Will open a PR for DEVTOOLS env variable with docs update

PR #2484, opened 2020-05-27
fix a bug with two manifests with the same name for site-assets

ranyitz:
### 🔦 Summary
Site assets now have 2 minified compilations and therefore they will have 2 `manifest.min.json` in the same location.

This PR will create `manifest.min.json` file for the node build (for backwards compatibility reasons) and `manifest-web.min.json` for the new web build.

__________

This also fixed a bug with creating a dev build for site-assets node which eliminated the content hashes.


ronami:
LGTM

PR #2487, opened 2020-05-28
Remove yoshi config for editor flow

yavorsky:
### 🔦 Summary
Another attempt to remove it w/o refactoring a lot of packages.


packages/yoshi-flow-editor/src/cli.ts
```diff
@@ -1,3 +1,6 @@
+// We want to default yoshi.config.js. It will be removed when yoshi will be refactored to not use loadConfig from
+// multiple places outside of the methods.
```
ranyitz, line 2:
```suggestion
// We want to always load the same yoshi.config.js
```

I would add the "TODO" on yoshi config itself

packages/yoshi-config/src/loadConfig.ts
```diff
@@ -11,7 +11,7 @@ const explorer = cosmiconfig('yoshi', {
 });
 
 export default ({ validate = true, cwd = process.cwd() } = {}): Config => {
-  const result = explorer.searchSync(cwd);
+  const result = explorer.searchSync(process.env.YOSHI_CONFIG_DIR || cwd);
```
ranyitz, line 14:
```suggestion
  // YOSHI_CONFIG_DIR is a hack for letting different flows to load a fixed
  // config, this is currently being used from both `jest-yoshi-preset` and the flow's cli
  const result = explorer.searchSync(process.env.YOSHI_CONFIG_DIR || cwd);
```

yavorsky:
To make workaround which will cover all cases we should have almost the same checks for jest-yoshi-config, legacy-flow (since we are still using it for unhandled scripts. It will be updated, but later) and some specific flow itself.

I agree we should find a better solution, but this one seems to be the most stable and prevents us from duplicating the code in different parts. I'm not sure we are not using `import yoshi-config` in some other internal modules, which will be broken if we'll forget to handle some other specific case.

This is the simplest solution, which will be easy to revert / update when we'll handle more cases in flow's part. Updating it in 3 places with the same logic will cause a lot of work in future to get rid of it.

PR #2494, opened 2020-06-01
petri-specs in a monorepo

ranyitz:
### 🔦 Summary
Our experimental monorepo flow which is being used in thunderbolt/editor-elements/editor-x

They all would like to have their `petri-specs` directory in the root.

This PR changes the `petri-specs` build task to happen once in the root in a monorepo (instead of in each app)

___________

1. This PR is needed for Thunderbolt - https://github.com/wix-private/thunderbolt/pull/7800 (@Alonkad)
2. Editor-x will get automatic updates for specs, because they are already using [this convention](https://github.com/wix-private/responsive-editor-packages/tree/master/petri-specs) (@amiryonatan)
3. Editor Elements will be able to start using experiments when needed (@giladsegal)


packages/yoshi-flow-monorepo/src/scripts/build.ts
```diff
@@ -123,14 +123,13 @@ const build: cliCommand = async function(argv, rootConfig, { apps, libs }) {
     const wixMavenStatics = await import('yoshi-common/build/maven-statics');
     const copyDocker = await import('yoshi-common/build/copy-docker');
 
+    // Run petri-specs once in the root of the monorepo
+    await petriSpecs.default();
```
amitdahan, line 127: Where'd the config-passing go?

```suggestion
    await petriSpecs.default({
      config: app.config.petriSpecsConfig,
    });
```

packages/yoshi-common/src/sync-petri-specs.ts
```diff
@@ -4,12 +4,12 @@ import petriSpecs from 'petri-specs';
...
 export = async function({
-  config,
+  config = {},
   cwd = process.cwd(),
 }: {
-  config: any;
+  config?: any;
   cwd?: string;
-}) {
+} = {}) {
```
amitdahan, line 12: image

packages/yoshi-flow-monorepo/src/scripts/build.ts
ranyitz: The config is loaded for each `app`. Petri-specs doesn't run for each app but once in the root. So if we want to add it we should probably use `rootConfig.petriSpecsConfig` instead.

I couldn't find any occurrences of petri-specs and this configuration option in our docs, so I assumed that it's not really important. Actually, i'm not sure what it does. So I thought that it would be good to supply this functionality without the config and add it when there is a need.

ranyitz: @amitdahan What do you say? Do you know what is it used for?

amitdahan: Found these references to this config object:
1. https://github.com/wix-private/fed-infra/blob/master/petri-specs/lib/petri-specs-build.js
1. https://github.com/wix-private/fed-infra/blob/master/ab-translate/src/abTranslate.js

Having said that, I agree about keeping it out of monorepo unless it's actually being used, which seems it's not :)

PR #2498, opened 2020-06-02
Configure `auto-release` by default

yavorsky:
### 🔦 Summary
Currently each new generated project should configure auto-release manually. Users should go to https://bo.wix.com/_serverless/app-service-autorelease/manage and configure it manually.
We want to remove this step at all and do it during new app generation.

We'll send a request to config generation service with `artifactId` and `appDefId` and will update the schema with the second request using received value.

So with just creating an app and pushing new code to github you'll have always fresh version for development mode. For production one, you have to submit each version in the dev center, which is expected.


packages/create-yoshi-app/scripts/dev.js
```diff
@@ -184,7 +184,10 @@ async function init() {
     });
   } else {
     // first time generation
-    workingDir = path.join(tempy.directory(), 'generated');
+    workingDir = path.join(
+      tempy.directory(),
+      process.env.DEV_ARTIFACT_ID || 'generated',
```
yavorsky, line 189: It's different addition, which helps to have custom artifact id for development mode

packages/create-yoshi-app/src/sentry-registration/addSentry.ts
```diff
@@ -24,14 +28,16 @@ export const addSentry = ({
   teamName: string;
   projectName: string;
 }) => {
-  addPackageJSONMock();
+  const packageMockWasAdded = addPackageJSONMock();
```
yavorsky, line 31: It's a bug, for `yarn create-yoshi-app:dev`, when it's removing `create-yoshi-app/package.json`.

yavorsky:
@amitdahan Hey. Good questions!
I agree that when if we'll generate it during the project generating, we can miss something for later development. For example, if user update the structure of the app and add new component via Dev Center manually, we will miss the step to regenerate both fedops and aut-release.

So in this case we should maintain it on CI, but we have to create some persistent storage where we'll compare new structure with the old one and provide some updates needed. 

But on the other hand, we want to provide configuration asap and allow users working with the app even if it wasn't published using local url overrides. So we are registering app in dev center via `create-yoshi-app`.

I'm working on the CLI extension for the current flow, so we can add new widgets / pages via it.
Something like `yoshi-flow-editor add` will open a wizard and ask what we are going to add, will register fedops and re-generate auto-release for it. This solution requires just reusing some tasks from CYA, which is pretty easy for current flow.

In case of moving the whole approach to CI process, it has own props, but requires some refactoring. I would like to continue with immediate-generation flow and see how it works and then move to CI-post-build generation if we understand it's better

PR #2502, opened 2020-06-03
Fix nested route inference in flow-bm

amitdahan:
### 🔦 Summary
There was a bug preventing inference of deeper-than-1 routes, for example:

`src/pages/a/b/c/index.tsx` would emit as `/` (or `/<routeNamespace>`).

This fixes that.
(I used `path.delimiter` instead of `path.sep` 🗡️)
(UPDATE: I used `path.sep` instead of `/`)


amitdahan:
Oh, sounds like I just need `/` then, I don't want different behavior in Windows.
Whoops!

ranyitz:
Yeah, It's for the url and not file system path

PR #2504, opened 2020-06-03
kill child servers on serve process exit

ranyitz:
### 🔦 Summary
When running the serve command there are left hanging processes. We do run the user server on a different process, we should verify it's killed when the parent process is killed.

For example, when used in Thunderbolt we see the sled script [exiting the process](https://github.com/wix-private/thunderbolt/blob/e99c891100911ec7c5cb6f8e424db3f9f67b39db/scripts/run-sled.js#L131-L134).

_____________

This PR make use of execa to verify that the child processes are being killed while the parent process is killed.


ranyitz:
The solution here is a naive solution as it doesn't wait for the servers to be killed. once `process.exit()` was called by a user we can't perform async operations and therefore it would be better to just send a message to the child process and let him kill itself. This is handled here - https://github.com/sindresorhus/execa#cleanup And I think should be used in this case or even a similar approach.

While this PR solves the problem described [here](https://wix.slack.com/archives/CL5TBPB8S/p1591084892221000) I do think we should improve this for other edge cases, to not leave hanging processes.

packages/yoshi-common/src/commands/serve.ts
```diff
@@ -67,5 +67,7 @@ export const serveApp = async ({
     }),
   ]);
 
-  return () => Promise.all([serverProcess.close(), cdn.close()]);
+  const closeServers = () => Promise.all([serverProcess.close(), cdn.close()]);
```
ranyitz, line 70: Just a refactor, nothing critical

packages/yoshi-common/src/server-process.ts
```diff
@@ -80,9 +81,10 @@ export class ServerProcess {
       './server-process-worker-with-transpilation.js',
     );
 
-    this.child = child_process.fork(serverProcessWorker, [], {
+    this.child = execa.node(serverProcessWorker, [], {
       stdio: 'pipe',
-      execArgv: [inspectArg]
+      // execArgv
```
ranyitz, line 86: https://github.com/sindresorhus/execa#nodeoptions-for-node-only

PR #2511, opened 2020-06-07
EP-1009|added support for appBuilderUrl

tzachiazo1:
### 🔦 Summary
...
Add an optional url to FlowEditorModel - appBuilderUrl.
In `bob-bootstrap` we want to open the appBuilder of the app the user currently working on with the local platformViewerScript.

Because we don't want to override other scripts like editorPlatformScript we added a map
to tell which overrides we want to use for each origin, (see urlOriginToSupportedOverridesMap in utils.ts)

### 🗺️ Test Plan
...


packages/yoshi-flow-editor/src/utils.ts
```diff
@@ -3,19 +3,43 @@ import { URL } from 'url';
...
 
 export const joinDirs = (...dirs: Array<string>) =>
   path.join(process.cwd(), ...dirs);
 
-export const normalizeStartUrlOption = (urls: URLsConfig): Array<string> => {
+const urlOriginToSupportedOverridesMap: Record<string, Array<string>> = {
+  viewerUrl: [
+    'tpaWidgetUrlOverride',
+    'widgetsUrlOverride',
+    'tpaSettingsUrlOverride',
```
yavorsky, line 15: `tpaSettingsUrlOverride` and `tpaWidgetUrlOverride` needed only for editor part, so we can remove it from  `viewerUrl `

packages/yoshi-flow-editor/src/utils.ts
```diff
@@ -57,45 +81,68 @@ const withComponents = (components: Array<ComponentModel>) => {
...
+  );
+};
+
 export const overrideQueryParamsWithModel = (
   model: FlowEditorModel,
   { cdnUrl, serverUrl }: { cdnUrl: string; serverUrl: string },
-) => (url: string): string => {
+) => (url: string | null | undefined, origin: string): string | undefined => {
+  if (!url) {
+    return undefined;
```
yavorsky, line 99: Can you please explain why it's becoming optional?

tzachiazo1: because urls in model.urls can be null | undefined,
I can check outside the function if the url is undefined and call it only if url is defined

WDYT?

packages/yoshi-flow-editor/src/utils.ts
tzachiazo1: 👌

packages/yoshi-flow-editor/src/utils.ts
yavorsky: Oh, I just wanted to clarify why it was changed
I think current solutions is good 🙂

PR #2512, opened 2020-06-07
in case there is an error during prompts it will now throw properly

ranyitz:
### 🔦 Summary
When there was an error thrown from the questions part of CYA the error was swallowed. This PR makes it so that it will be thrown.

While this is not a user facing error, it makes it easier to debug in case there is an error.


packages/create-yoshi-app/src/runPrompt.ts
```diff
@@ -17,6 +17,8 @@ export default async (workingDir = process.cwd()) => {
       console.log('Aborting ...');
       process.exit(0);
     }
+
+    throw e;
```
yavorsky, line 21: LGTM!

Issue #2515, opened 2020-06-07
Switch to prettier 2.x (to support typescript 3.8+)


liorcode:
# 🙋 Feature Request
### 😯 Current Behavior

Prettier 1.9 is used, which does not support typescript 3.8 syntax and breaks (e.g.: having `import type...` in a file, would break prettier)

### 🤔 Expected Behavior

Use prettier 2+
### 💁 Possible Solution 

For now, i'm overriding the default prettier version using a yarn resolution:
```
  "resolutions": {
    "**/prettier": "2.0.5"
  }
```

### 🔦 Context 

This is the PR where prettier added support for TS 3.8: 
https://github.com/prettier/prettier/pull/7631
(they only merged it to prettier 2)

Note that prettier 2 does come with some breaking rule changes:
https://prettier.io/blog/2020/03/21/2.0.0.html#breaking-changes
The biggest one is "Change default value for `arrowParens` to `always`". Can be overridden of course, or otherwise it would generate warnings for anyone relaying on no parenthesis.

ranyitz:
@liorcode Thanks for opening that! We should definitely upgrade.

We usually upgrade major versions in a yoshi major version, but in fact, prettier is configured to only add warnings and therefore should not be a breaking change.

Of course that there are users that configure their CI to break on warnings, but that's an edge case and it's usually the case in highly maintained codebase. I think that we should upgrade prettier and require them to upgrade their codebase accordingly.

The only problem that I'm afraid from is that they [dropped node@8 support](https://prettier.io/blog/2020/03/21/2.0.0.html#drop-support-for-node-versions-older-than-10-6908httpsgithubcomprettierprettierpull6908-by-fiskerhttpsgithubcomfisker). And that there are some projects that still depend on it. The node platform [requires node@12](https://github.com/wix-platform/wix-node-platform/pull/2755) While [create-yoshi-app requires node 12 as well](https://github.com/wix/yoshi/pull/1614).

In this case, it only means it will may break a project that uses yoshi and node 8.

____________

I'm willing to accept a PR that upgrades prettier, we'll adjust if needed.

PRs welcome.

ranyitz:
Fixed on #2516

PR #2516, opened 2020-06-07
chore(lint): upgrade prettier to v2

liorcode:
# 🔦 Summary
Upgrade prettier 1.x to 2 
See #2515 

Note that since prettier 2 introduced some breaking changes, this PR also effects the yoshi monorepo itself.
To align all the yoshi projects with the new version, `yarn lint --fix` should run on the entire repo.
As this would reduce in a large PR, I didn't want to do it here.


liorcode:
done 👍

ranyitz:
@liorcode The templates tests are flaky and not a result of this PR.
The unit tests are failing for a reason. (We test our eslint config and it fails because of prettier instead of the real reason)

I don't want to waste your time, so let me know if you can take this. In case your stuck I can make a PR to this PR.

Thanks

liorcode:
it appears the tests are failing because the tests needs to be updated to match the new prettier rules.
I'd be happy if you could PR the PR :)

ranyitz:
@liorcode here is the PR https://github.com/liorcode/yoshi/pull/1

PR #2521, opened 2020-06-08
Init flowAPI even if no viewer.app.ts was created

yavorsky:
### 🔦 Summary
There is a bug introduced after https://github.com/wix/yoshi/pull/2507 merge

We are initializing `flowAPI` for controllers in initAppForPage wrapper. And in case user doesn't have it - we are passing this process, so no flowAPI is appearing in controller.


packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -176,7 +176,7 @@ export const createControllersWithDescriptors = (
 };
 
 export const initAppForPageWrapper = (
-  initAppForPage: InitAppForPageFn,
+  initAppForPage: InitAppForPageFn | undefined,
```
yavorsky, line 179: For now it's optional

packages/yoshi-flow-editor/src/utils.ts
```diff
@@ -84,13 +84,10 @@ export const overrideQueryParamsWithModel = (
...
   );
 
-  // Adding editorScript override url only if editor.app.ts entry file is present in project
-  if (model.editorEntryFileName) {
-    urlWithParams.searchParams.set(
-      'editorScriptUrlOverride',
-      editorScriptUrlFormatter(model, cdnUrl),
-    );
-  }
+  urlWithParams.searchParams.set(
```
yavorsky, line 87: We also will override editorScript url to prevent cases, when user removed it locally, it's fetching from the CDN w/o override and then some bug appear in future when new version of editorScript will be appear.
We want to be consistent locally and override to localhost in any case.

packages/yoshi-flow-editor/src/wrappers/templates/CommonEditorScriptEntry.ts
```diff
@@ -21,7 +21,7 @@ export default t<Opts>`
     return `var editorScriptEntry = ${
       editorEntryFileName
         ? `require('${editorEntryFileName}');`
-        : `{ editorReady: function {} };`
+        : `{ editorReady: function () {} };`
```
yavorsky, line 24: 🥴

packages/yoshi-flow-editor/src/wrappers/templates/CommonViewerScriptEntry.ts
```diff
@@ -91,10 +91,7 @@ export default t<Opts>`
   }`
       : 'null'};
 
-  ${({ viewerEntryFileName }) =>
```
yavorsky, line 94: We want to add it in any case, since in wrapper we are initializing the flowAPI for viewer script and using it for controller.

PR #2522, opened 2020-06-08
Add warning about global installation

### LINKED ISSUE #2481: Update create-yoshi-app to download latest version under the hood
# 🙋 Feature Request

### 😯 Current Behavior
Currently a lot of people are installing `create-yoshi-app` as a global package and use it for months. It's bad b/c we are updating it a lot, so users should receive latest ones.

We are asking to use `npx create-yoshi-app@latest` in docs, but anyway it's being ignored by 
users.

### 🤔 Expected Behavior
The create-yoshi-app should always use latest version.

### 💁 Possible Solution 
Would be good if we'll have create-yoshi-app as a wrapper with latest version loader inside.
It will download latest version of our runner with latest templates and scripts.

Users can still use global version, so a lot of issues of using outdated version will be resolved.

`create-yoshi-app my-app` -> GET create-yoshi-app@latest -> execute it

daliusd:
Closes #2481


amitdahan:
Thanks @daliusd 🙏!

PR #2523, opened 2020-06-08
Fix a bug with server process and `--debug` argv

ranyitz:
### 🔦 Summary

This PR move the parsing of `process.argv` (CLI arguments) to the CLI, out of the internal server process.

_________________________

Up until now, the server process used to look into `process.argv` and pick up `--debug` and `--debug-brk` parameters.

This worked fine for simple cases, but I believe that it's generally good to parse CLI arguments on the CLI level and pass them to the "cli agnostic" server process as a simple flag.

This is adding more boilerplate, and is also means that this feature won't work OOTB for other flows, but I'm totally ok with that. Other flows may want to use a different API or just add this line when they'll encounter a usecase.

### Context
There was an edge case when sled tests ran in Thunderbolt and added a way to debug them:

```yarn run-sled --debug```

This command ran yoshi's `serve` using the node API and in turn also ran the server process.

This caused multiple debuggers to be attached into those servers while this was not the intent of the user.

image

This bug was reported by @eyal-er, after it's merged we'll open another PR to upgrade the version of yoshi in Thunderbolt.


eyal-er:
🥇

PR #2525, opened 2020-06-08
Do not show progress bar if `DEBUG=true`

### LINKED ISSUE #2350: Turn off webpackbar (progress bar) during DEBUG mode
# 🙋 Feature Request

After #2142 we can disable the UI of the progress bar when using `PROGRESS_BAR=false` this feature suggest that we'll remove it in case `DEBUG` environment parameter is `true`.

It's important to say that `PROGRESS_BAR=false` doesn't remove it completely but instead it changes the UI to be the "basic" UI which leaves much less clutter on the screen. While this is not bad, I would prefer to not have that as well during debugging.

You can see that we have a [`useProgressBar`](https://github.com/wix/yoshi/blob/7f020e7218312f1cf9eeb80b68ea9d27504b797d/packages/yoshi-common/src/webpack.config.ts#L335) option that can be easily disabled in case `process.env.DEBUG==='true'`

daliusd:
Closes #2350.


daliusd:
It was not 100% sure if PROGRESS_BAR functionality must be removed or it must be kept in place so I have not touched it.

PR #2526, opened 2020-06-08
Support configuring webworker compilation's resolve.alias

ronami:
### Summary

Pretty straight-forward, it's currently needed specifically in Thunderbolt and is already supported for client compilation.

I only exposed this configuration in monorepo and app flows.


yotambarzilay:
@ronami 🙏

packages/yoshi-config/src/validConfig.ts
```diff
@@ -76,6 +76,7 @@ const validConfig: RequiredRecursively<InitialConfig> = {
         app: 'index.js',
       },
     ),
+    resolveAlias: {},
```
yotambarzilay, line 79: @ronami does this require some values in order to pass validation? I'm getting validation warnings:
```
Unknown option "webWorker.resolveAlias.lodash$" with value "/Users/yotamb/Projects/thunderbolt/packages/thunderbolt-app/src/lodashWorker.ts" was found.
  This is probably a typing mistake. Fixing it will remove this message.
```

ranyitz: I didn't dive into that, but I see that we're passing an object (non empty) to the client's compilation resolve alias - https://github.com/wix/yoshi/blob/808412a78f6fc24674e8b50d25459f62fa50f1c8/packages/yoshi-config/src/validConfig.ts#L37-L39

PR #2528, opened 2020-06-09
Remove editor.app.ts and viewer.app.ts from the template. Update docs

yavorsky:
### 🔦 Summary
This change provides 2 things:

- Removing viewer and editor script entries from the create-yoshi-app template. We decided that it's additional boilerplate and better to just get rid of it + provide guide in the docs how to init it. We are also going to allow adding these files via editor flow CLI tool.
- Merging app configuration level in the docs with application level. Technically it's located in different directories (root and src), but having separate article related to app configuration on the same level as app level or component level is confusing. User's usually doesn't touch it and it's much better to prioritize something more useful.
- Updated a bit editor and viewer script entries documentation related to initialization part and general idea behind it.


website/docs/editor-flow/structure-api/application.md
```diff
@@ -4,8 +4,40 @@ title: Application Level
 sidebar_label: Application Level
 ---
 
+# App configuration
```
ranyitz, line 7:
```suggestion
```
It looks a bit weird to have 2 h1 headers on after the other:

image

https://wix-yoshi-website-pr-2528.surge.sh/docs/next/editor-flow/structure-api/application/

yavorsky: wow, that's weird.. locally it looks like
Screen Shot 2020-06-09 at 4 18 04 PM

PR #2531, opened 2020-06-09
Write full webpack errors log to file.

### LINKED ISSUE #2378: When Yoshi crashes, throw a very verbose information to a file
# 🐛 Bug Report

### 😯 Current Behavior
In case Yoshi's crash this is most of the time a result of webpack compilation failure. When this happens there are many cases where the output is not detailed enough. It depends on what is the error of course and who throw it. But I've seen that there are many cases when there are error messages in the stats file that could reduce the amount of time needed for debugging a lot.

While we use [format-webpack-messages](https://github.com/wix/yoshi/blob/05558a2503dc198abf146371657e5fbcb7417211/packages/yoshi-common/src/webpack-utils.ts#L226) to get nicer errors for the CLI it does makes sense to be able and see the original error somewhere.

The workaround is to save the stats to a file (modifying the builded `node_modules`) something that i'm not expecting users to do.

### 🤔 Expected Behavior

The full errors array would be available somewhere.

### 🔦 Context
I debug a lot of bugs with Yoshi and this problem slows me down.

### 💁 Possible Solution

Generate a `yoshi-error.log` file with the output. It can be on the root on the monorepo or even `target/yoshi-error.log` if we want to make this a subtle change. (As directory is already git ignored for most of the projects @ Wix)

### 💰 Additional value
We can also add the "info" information to this file and tell developers to send it to a slack channel in case there is a bug. It should contain a lot of valuable information for anyone who wants to understand what is the problem.

daliusd:
Closes #2378


daliusd:
I have not added information that this file can be sent to slack channel as project is public but channel is private.

packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -233,6 +233,9 @@ async function runWebpack(
       console.log(chalk.red('Failed to compile.\n'));
       console.error(messages.errors.join('\n\n'));
 
+      const fullLog = webpackStats.stats.map(s => s.toString()).join('\n\n');
```
ranyitz, line 236: Is this the whole stats, or we're filtering the errors?

daliusd: Here is everything. I have not filtered error messages out in case other messages could help with context.

daliusd:
> It worth logging that a full error log will be located on `target/yoshi-error.log`.
> 
> So the flow would be, seeing the message and opening this file to further research.

Done.

packages/yoshi-common/src/webpack-utils.ts
ranyitz: The stats object can be very big, this means that even just writing them to a file can take time. I don't want to just write the stats (as this functionality is available using `--stats`).

I do think that we can start out with filtering of the `errors` and maybe also the `warnings`

Please see how [this object looks](https://webpack.js.org/api/stats/#structure)

ranyitz: We have to remember that if the log would be too long, no one will read it as well. Here we have the pareto (Just to log the errors). And after we can continue and refine this file

daliusd: OK. So I'm writing only stats that hasErrors() now. E.g. I have one error and file has 105 lines.

ranyitz: I would like to verify that we're only outputting relevant things. Can you send an example to how this file output will looks like?

(Create a problem output a file, and show its contents, right now, i'm not sure what it would look like)

PR #2537, opened 2020-06-09
fix create yoshi app library flow creation bug

ranyitz:
### 🔦 Summary
This problem was reported in this [slack thread](https://wix.slack.com/archives/CAL591CDV/p1591706345362200?thread_ts=1585234306.031600&cid=CAL591CDV)

#2514 Added this degradation, this is a fix.

_________

We should add an e2e tests for this template


ranyitz:
Fixed the tests, will merge when it's green

PR #2539, opened 2020-06-09
Send telemetry when start building

ranyitz:
### 🔦 Summary

This PR is basically the same as @ronami's https://github.com/wix/yoshi/pull/1722 with small adjustments.

After #2529 we can integrate wix's bi-loggers without breaking users.

_________

This will send the [buildStart](https://bo.wix.com/bi-catalog-webapp#/sources/11/events/91400?artifactId=com.wixpress.yoshi) event to the BI.

We will also know the flow used by the user and whether the command ran from the CI or locally.


packages/yoshi-flow-bm/src/scripts/build.ts
```diff
@@ -20,6 +21,8 @@ import renderModuleConfig from '../renderModuleConfig';
 const join = (...dirs: Array<string>) => path.join(process.cwd(), ...dirs);
 
 const build: CliCommand = async function (argv, config) {
+  telemetry.buildStart('Editor', config.name);
```
ranyitz, line 24:
```suggestion
  telemetry.buildStart('BM', config.name);
```

packages/yoshi-flow-library/src/commands/build.ts
```diff
@@ -11,13 +11,16 @@ import fs from 'fs-extra';
...
+import * as telemetry from 'yoshi-common/build/telemetry';
 import runBabel from 'yoshi-common/build/typescript/run-babel';
 import { createClientWebpackConfig } from '../webpack.config';
 import { cliCommand } from '../cli';
 
 const join = (...dirs: Array<string>) => path.join(process.cwd(), ...dirs);
 
-const build: cliCommand = async function (argv, config) {
+const build: cliCommand = async function (argv, config, pkgJson) {
+  telemetry.buildStart('App', pkgJson.name!);
```
ranyitz, line 22:
```suggestion
  telemetry.buildStart('Library', pkgJson.name!);
```

packages/yoshi-common/src/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+// Create BI factory
+const biLoggerFactory = biLoggerClient.factory() as BiLoggerFactory<
+  ReturnType<typeof getLoggerConf>
+>;
+
+// Register a custom publisher that uses Node's HTTPS API
+biLoggerFactory.addPublisher(async (eventParams: any, context: any) => {
+  // Don't collect events from yoshi's own tests
+  if (process.env.NPM_PACKAGE === 'yoshi-monorepo') {
+    return;
```
ranyitz, line 17: Instead of sending BI for our tests with a property, we'll just won't send it at all.

packages/yoshi-common/src/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+    // Swallow errors
+  }
+});
+
+// Create logger
+const biLogger = initSchemaLogger(biLoggerFactory)();
+
+const { version: yoshiVersion } = require('../package.json');
+
+biLogger.util.updateDefaults({
```
ranyitz, line 32: The next PR would be for`startStart` 🤣  But really, I feel that those are not specific to an event and we would like to always have them.

PR #2540, opened 2020-06-09
make shouldDeployToCDN using ciBuild info

justame:
### 🔦 Summary

* [ ] shouldDelpoyToCDN
* [ ] anotherFunction

### 🗺️ Test Plan
...

### DIFF
--- a/packages/yoshi-helpers/src/queries.ts
+++ b/packages/yoshi-helpers/src/queries.ts
@@ -6,6 +6,9 @@ import * as globs from 'yoshi-config/build/globs';
 import { POM_FILE } from 'yoshi-config/build/paths';
 import isCi from 'is-ci';
 import { defaultEntry } from './constants';
+import { tryRequire } from './utils';
+
+const ciBuildInfo = tryRequire('@wix/ci-build-info');
 
 export const exists = (
   patterns: string | ReadonlyArray<string>,
@@ -76,7 +79,11 @@ export const hasBundleInStaticsDir = (cwd = process.cwd()) => {
   );
 };
 
-export const shouldDeployToCDN = () => {
+export const shouldDeployToCDN = (packageName: string) => {
+  if (ciBuildInfo) {
+    const buildInfo = ciBuildInfo.getBuildInfoByPackageName(packageName);
+    return isCi && buildInfo.artifactVersion;
+  }
   return (
     isCi &&
     (process.env.ARTIFACT_VERSION || process.env.SRC_MD5) &&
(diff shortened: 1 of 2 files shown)


packages/yoshi-helpers/src/queries.ts
```diff
@@ -6,6 +6,9 @@ import * as globs from 'yoshi-config/build/globs';
 import { POM_FILE } from 'yoshi-config/build/paths';
 import isCi from 'is-ci';
 import { defaultEntry } from './constants';
+import { tryRequire } from './utils';
+
+const ciBuildInfo = tryRequire('@wix/ci-build-info');
```
ranyitz, line 11: We don't specify `wix/ci-build-info` in the dependencies. I guess that it will be supplied on our CI environment somehow. I think that it's not very clear and that we should add a comment above explaining it for future developers working on the repo.

ranyitz: Can we also add a link to where this library is managed on GitHub?

packages/yoshi-helpers/src/queries.ts
```diff
@@ -76,7 +79,11 @@ export const hasBundleInStaticsDir = (cwd = process.cwd()) => {
   );
 };
 
-export const shouldDeployToCDN = () => {
+export const shouldDeployToCDN = (packageName: string) => {
+  if (ciBuildInfo) {
+    const buildInfo = ciBuildInfo.getBuildInfoByPackageName(packageName);
```
ranyitz, line 84: How does this function work? We use the `package.json` name to get the artifact version?

packages/yoshi-helpers/src/queries.ts
```diff
@@ -76,7 +79,11 @@ export const hasBundleInStaticsDir = (cwd = process.cwd()) => {
   );
 };
 
-export const shouldDeployToCDN = () => {
+export const shouldDeployToCDN = (packageName: string) => {
```
ranyitz, line 82: If we're not on CI we can just exit early as this condition is duplicated.

e.g.

```ts
if (!isCI) {
  return false;
}

// Here we should have an explanation about the fact that this condition
// will be used only on the CI 2.0 when this module would be available
if (ciBuildInfo) {
  return ciBuildInfo.getBuildInfoByPackageName(packageName).artifactVersion;
}

return (process.env.ARTIFACT_VERSION || process.env.SRC_MD5) && fs.existsSync(POM_FILE);
```

ohana54:
@ranyitz I continued the work in #2699, so this can be closed.

ranyitz:
Thanks @ohana54 closing

PR #2541, opened 2020-06-09
Automatic `fedops` logging for Editor Flow projects

yavorsky:
### 🔦 Summary
It's another enhancement how to reduce boilerplate from user's side and handle fedops interaction almost all apps have automatically.

The idea, is to call `fedops` methods `appLoadStarted` and `appLoaded` by the flow.
We'll also make it by following best practices for widget: both CSR and SSR will be handled.

if user wants to log some interaction - we still providing `fedopsMonitor` via `flowAPI`:
```
export default ({ flowAPI }) => {
  const handleSubmitSomething = () => {
    flowAPI.fedopsLogger.interactionStarted(SUBMIT_SOMETHING_FEDOPS_INTERACTION);
    // process the action...
    flowAPI.fedopsLogger.interactionEnded(SUBMIT_SOMETHING_FEDOPS_INTERACTION);
  };

  return {
    pageReady() {
      setProps({
        handleSubmitSomething,
      });
    }
  };
};
```

To notify user's who didn't got the idea and keep logging `appLoadStarted` / `appLoaded` on the project side, we still can provide these methods, but notify users to remove these calls.

Docs was also updated a bit.


packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -108,16 +116,21 @@ function ooiControllerWrapper(
       throw error;
     })
     .then((userController: any) => {
+      if (flowAPI.isSSR()) {
+        flowAPI.fedopsLogger.appLoaded();
```
bordi, line 120: According to the docs  SSR `appLoaded` should be called at the end of client `pageReady`. In this case it will be called before. Or am I missing something? 
Docs: https://bo.wix.com/wix-docs/client/viewer-platform---ooi/articles/fedops#viewer-platform---ooi_articles_fedops_ssr

yavorsky: good catch, thanks. I missed that we have own pageReady implementation, so it should be called after flow's one.

It will be replaced in near future, so we'll have single pageReady and won't provide an additional call by the flow, but currently it would be good to update it. Thanks!

bordi: 👍 Got it. Thank you, Artem for amazing work!

yavorsky: 😄 Thanks for the perfect catch!

bordi: image

packages/create-yoshi-app/templates/flow-editor/typescript/__tests__/helpers/controllerConfig.mock.ts
```diff
@@ -11,32 +15,10 @@ export default (
       warmupData: jest.fn(),
       $w: jest.fn() as any,
       platformAPIs: {
-        pubSub: {
-          publish: jest.fn(),
-          subscribe: jest.fn(),
-          unsubscribe: jest.fn(),
-        },
+        fedOpsLoggerFactory,
+        pubSub: new PubSubMock(),
```
yavorsky, line 19: Small change not related to PR, but I would include it here instead of separate PR. In future we'll export mock from editor runtime part

PR #2543, opened 2020-06-10
Copy templates in flow-bm

amitdahan:
### 🔦 Summary
Adds missing step for copying templates in BM flow.
(This broke the first merged migration - `price-quotes-web` 😄)


amitdahan:
I'm wondering if app/editor/bm flows need to share some `start`/`build` logic.
They all clean `dist` & `target`, they all need to copy templates, etc, launch a devenv, etc.

I do think it made sense to diverge for easier prototyping of the flows, but I'm starting to think about having some shared "phases" (?).

amitdahan:
@ranyitz Not being called in `start.ts` (app & editor flow), so not adding it here as well.

PR #2555, opened 2020-06-11
Make library flow template deploy a tar.gz file to parastorage

amitdahan:
### 🔦 Summary
Libraries that expose a umd bundle with some dynamic imports, will only work when pulled from good ol' `parastorage`.

This lets the default library template upload statics to `parastorage` (previously only npm worked OOTB).


ohana54:
@amitdahan question - should this be the default?
NPM packages at Wix are not published by default in the public registry, meaning the code stays private. With this change, if I understand correctly - every npm package will be uploaded to parastorage, making the code public. This also affects the load on CI and Media Platform.
WDYT?

cc @liricooli @netanelgilad

netanelgilad:
Maybe the flow should be changed to "web library flow". For other types of libraries (used by backend for example) this is a security hole and probably also just not needed.
My personal suggestion is to change the name of the flow, and maybe add a verification question in the generator aling the lines of "You are about to create a web library, which is public, are you sure?".

Also, I think using unpkg:true for those web libs is better than basic parastorage, as it has semver capabilities.

ohana54:
@ranyitz as Amit is on vacation, we'd love your opinion here.
There's a good chance I misunderstood the implications, so please correct me :)

ranyitz:
First of all, thanks for joining the conversation, I think that those are great inputs!

Just to be clear, this is a **web/browser/client library**, meaning it was designed to run in a web browser (In oppose to a server library/node library). In this case, the code will end up in parastorage (Either by itself or by a different deployable artifact that bundles it).

> With this change, if I understand correctly - every npm package will be uploaded to parastorage, making the code public

This is true, it will also publish it to npm by default. We did that because causing a library to publish its code to parastorage was **quite hard and time consuming task** and we didn't want everyone to face this. I do think that there can be unused publishes (publish to parastorage when no one actually use it from there).

**There are [two main ways to consume a library created by this flow](https://wix.github.io/yoshi/docs/library-flow/#how-can-i-use-a-library-create-by-this-flow):**

**From npm** - Import the library from the application's code and bundle it
**From the cdn** - Consume the already bundled library using a `<script src="library.umd.js">` tag

Usually it's consumed by a single way, but there is also a case when a library will be consumed by both ways and it's up to the consumer to decide what's the best way. It is very nice when a library supports both ways of publishing, the advantages are that less time need to be wasted on configuration down the road. The downside is that there may be unused publishes.

If a user decides to create a bundle, It would make sense to assume that he wants to publish it to parastorage. There is a configuration to [create this bundle](https://wix.github.io/yoshi/docs/library-flow/#configuration) which is turned on by default, and we can argue whether it's the right thing to do. But without it there would not be a `dist/statics` directory and therefore nothing should be deployed to parastorage.

I'm willing to change the template name to `browser-library`. I believe that it will make it clearer.
___________________

Regarding `unpkg` really cool initiative, unfortunately we can't use dynamic imports OOTB because [we don't know what would be the public path](https://github.com/wix/yoshi/pull/455/files#diff-38fe2a94fa5e620d1165ac1611c59438R32-R38) as the version of the package is decided during publish time.

netanelgilad:
Cool. 
+1 on the name change.
unpkg - agree, we'll solve in Falcon CI (the new CI new codename :D )

ranyitz:
+1 on Falcon CI 👍  I'll change the name of the template to `browser-libarary` so it'll be clearer.

I still want to say that we're going to continue to create the `pom.xml` file with the statics configuration but I'm considering to remove the `bundle` option from the initial template, this means that there would be nothing created on `dist/statics` and that the only change you'll need to do is to add `bundle: true` in the `package.json`.

I'm wondering what would happen if the `pom.xml` would be configured to upload `dist/statics` and there would be nothing there.

netanelgilad:
I'm not sure myself, but I think it will scream

ranyitz:
If it will scream we can totally just generate/modify the `pom.xml` ourselves in case the user has configured `bundle: true`

PR #2556, opened 2020-06-12
Screencast scripts and simple screencast created.

### LINKED ISSUE #1333: Auto record installation SVG for readme
# 🙋 Feature Request

### 😯 Current Behavior

The gif of installation using `create-yoshi-app` looks bad and it's outdated.

### 🤔 Expected Behavior

We'll have a script that generates an up to date nice looking SVG of the creation.

### 💁 Possible Solution 

Use https://asciinema.org/ to create a script that generates the SVG for us.
See https://github.com/Schniz/fnm/tree/master/docs & https://github.com/facebook/create-react-app/blob/master/tasks/screencast.js for inspiration.

This script can take ~2 minutes to run, which means we don't really want to run it as part of our release/build process, but it would be nice if we'll have a command that generates it when needed 💯

daliusd:
Closes #1333.


scripts/screencast/screencast.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/zsh
+
+set -e
+
+# Disable npx progress bar
+export TRAVIS=2
+
+GAL_PROMPT_PREFIX='\e[34m✡ \e[0m'
```
Schniz, line 8: 💯

Issue #2520, opened 2020-06-08
[RFC] yoshi-flow-editor CLI extension


yavorsky:
# 🙋 Feature Request

This is a RFC for `yoshi-flow-editor` CLI command tool extension.

Currently we support few basic commands:
- `start`: starts a development server with local overrides
- `build`: builds production bundles
- `test`: start unit and e2e tests with pre-configured jest under the hood
- `lint`:  lint check

All commands are working pretty good on the current stage. But it could be extended with bigger functionality.

### `test`
It starts a unit and e2e tests. e2e tests we have just for Settings, since Widget is being tested via `sled`.
There are few issues:
- Widget for Editor part is not being tested at all. Sled is testing it only in viewer environment.
- Settings is being tested w/o scope also: just serving an app on localhost and providing some e2e checks.
- This command is not starting sled tests. So usually we should have `yoshi-flow-editor test && sled`. It's not good practice in general, since we these commands is not running in parallel and and with `|` operator we will mix/lose some logs.

There're some ideas how we can improve the flow in general
- experiment with running sled against editor environment. We can already override editorScriptUrl: https://github.com/wix-private/sled/pull/541 and since we can already use query params to override everything we need to editor part, maybe few changes to the sled side will help is.
- Proxy sled command from the yoshi CLI and run it when we are running `test`. We can also support commands like `test:unit`, `test:sled`, so users can run some specific tests locally.

### `add`
A new command that should add new things to the current project.

Currently, we are bootstrapping almost everything needed for development and release with `create-yoshi-app` for editor flow. It's Dev Center registration for widgets, pages, platform Components with `viewerScript`, `editorScript` urls, configuring sentry, fedops, basic experiments bootstrap.

But on the same time it's really hard to add new component. You need to provide all the steps manually spending much more time than just generate a new project.

With `yoshi-flow-editor add` command we can open a wizard with questions to:
- add `Page` | `Widget`. In this case we'll provide the whole flow to correctly inject it in our project: Register on Dev Center, update `fedops.json`, create a correct `.component.json` config with component id, create a sled test for it.
- add `editor.app.ts` | `viewer.app.ts` (editorScript and viewerScript entry-points). In this case we'll just bootstrap it with some simple code `initAppForPage` or `editorReady`. It will be really useful if we want to remove this files from the `template`.
- add `sentry` - in case of changing the sentry team or adding a sentry to project that didn't configure a sentry, so correct `sentry` config will be updated in `.application.json`
[...]

PR #2557, opened 2020-06-12
New CLI Command for editor flow: add

yavorsky:
### 🔦 Summary
Implementation of the adding a component part from the RFC: https://github.com/wix/yoshi/issues/2520

Currently, we are bootstrapping almost everything needed for development and release with create-yoshi-app for editor flow:
- Generating a basic Widget, controller and Settings
- Dev Center registration for widgets
- configuring sentry
- sled tests for each component
- fedops configuration
- basic experiments bootstrap
- auto-release for current relevant project structure

And it's working like a magic for structure created via `create-yoshi-app`.
In case of new components added, some features will be missed for it under the hood b/c of lack of configuration. It's happening only from the project bootstrap.

### `yoshi-flow-editor add`
This PR solves it with new `add` command. It will open wizard with 3 options:
`Add a Widget` - enabled one.
`Add a Page` - disabled. When you hover it, we'll warn that 'adding a page to current project can break it in production and we are working on providing good solution on it'. We still preserve Page option, so user won't be confused if we'll just add Widget automatically.
`Cancel` - just close the promt.

After, we'll ask for a component name.

Then we can generate a new component:
- Register it in dev center
- Add to your project's structure (`src/components/:newComponent`)
- Generate new sled tests for it
- Update `fedops.json`
- Update [auto-release](https://bo.wix.com/_serverless/app-service-autorelease/manage) with new structure

ezgif com-video-to-gif (2)

### DIFF
--- a/packages/yoshi-flow-editor/src/commands/add-component/index.ts
+++ b/packages/yoshi-flow-editor/src/commands/add-component/index.ts
@@ -1,0 +1,71 @@
+import arg from 'arg';
+import {
+  addOOIComponentStep,
+  ExtendedPropmtsAnswers,
+  extendedPropmts,
+  getAuthInstance,
+  initAPIService,
+} from 'create-yoshi-app';
+import { cliCommand } from '../../cli';
+
+const add: cliCommand = async function (argv, config, model) {
+  const args = arg(
+    {
+      // Types
+      '--help': Boolean,
+
+      // Aliases
+      '-h': '--help',
+    },
+    { argv },
+  );
+
+  const { '--help': help } = args;
+
+  if (help) {
+    console.log(
+      `
+      Description
+        Adds a component to current project, integrates it with Dev Center, Fedops and sled configuration.
+
+      Usage
+        $ yoshi-flow-editor add
+
+      Options
+        --help, -h      Displays this message
+    `,
+    );
+
+    process.exit(0);
+  }
+
+  let answers: ExtendedPropmtsAnswers<string>;
+
+  const instance = await getAuthInstance();
+  if (instance) {
+    initAPIService(instance);
+  }
+
+  try {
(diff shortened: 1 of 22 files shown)


packages/yoshi-flow-editor/src/commands/add-component/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    process.exit(0);
+  }
+
+  const templateModel = await runPrompt(model);
+  generateTemplate(templateModel, 'src/components');
+  generateTemplate(templateModel, 'sled/ssr');
+  const flowData = templateModel.getFlowData();
+  if (!flowData) {
+    throw new Error("Can't initalize the flow data for a new component");
```
ranyitz, line 63: I'm not sure what is the case for this error, but I would move it to be before we actually generate something, now it seems that we first generate files and throw after.

packages/yoshi-flow-editor/src/utils.ts
```diff
@@ -107,3 +107,7 @@ export const generateSentryScript = (sentry: SentryConfig) => {
   var t=c[p];c[p]=function(a){m({p:a.reason});t&&t.apply(c,arguments)};f||setTimeout(function(){k(h)})})(window,document,"script","onerror","onunhandledrejection","Sentry","${sentry.id}","${BROWSER_LIB_URL}",{"dsn":"${sentry.DSN}"});
   </script>\n`;
 };
+
+export const normalizeProjectName = (projectName: string) => {
```
ranyitz, line 111: Not critical ofcs, but we have a function with similar functionality [here](https://github.com/wix/yoshi/blob/9c2d1376450a4c1b84db94760875084a2778f839/packages/yoshi-helpers/src/utils.ts#L259-L260)

PR #2558, opened 2020-06-12
html-webpack-plugin enabled in library flow.

### LINKED ISSUE #2263: yoshi-library-flow have a basic served html
# 🙋 Feature Request

Create a basic `html` that includes the `js` bundle and the `css` bundle as script and link tags. This html should be served using webpack-dev-server as part of the bundle itself.

Today it's needed to create this with a custom script and also serve that `html` for both testing and working locally. It makes sense to present it as a feature, the user doesn't have to use it, so it should be on by default.

___________

Later on we'll probably need to enable its customization, but for start, just having it would work.

daliusd:
Closes #2263.


packages/yoshi-flow-library/package.json
```diff
@@ -30,6 +30,7 @@
     "execa": "4.0.2",
     "fs-extra": "8.1.0",
     "globby": "11.0.0",
+    "html-webpack-plugin": "4.3.0",
```
amitdahan, line 33: Hey did you run `yarn`? I think there should be a small lock file change, no?

daliusd: I run `yarn`. No changes.

daliusd:
> 1. Where does this point to by default? Does it use a default empty (in-memory) html file that has the generated umd bundle?

If you run `npm run build` it created index.html file:

```
Compiled successfully.

  1.4 KB  (633 B GZIP)  dist/statics/generated.umd.min.js
  235 B  (187 B GZIP)  dist/statics/index.html
```

I have not even assumed that `npm start` might be benefiting from this.

> 2. If we don’t bundle, it simply won’t run? (If so - makes sense)
I guess so.

> 3. Does the browser open the html file by default?
No.

> 4. Do we/should we allow overriding the html template or an empty one should be enough? For example if said umd bundle just exposes something on the window, and I want to call a function on the window-attached namespace? I guess I could call it in console, but I guess different use-cases could bring up different needs. Wdyt? This can also be improved in the future.
I guess empty will do now. I'm not really sure yet what value that creates actually :)

> 5. Should we also document this? It’s possible that if it opens by default and no configuration is needed for most imaginable cases, then documentation is redundant, but maybe we can just explain that `yoshi-library start` will render your umd bundle by default somewhere in library flow docs?

I am not completely sure if what I have done is enough and what we want to achieve. Therefore documenting this is even harder 😄

PR #2565, opened 2020-06-14
Migration guide for Library flow from the Legacy flow

ranyitz:
### 🔦 Summary
This PR adds a migration guide for the Library flow.

Now that we need more than a single page for a library flow it make sense to separate it into multiple pages.

This PR also separates the library flow into 3
* Overview
* Configuration
* Migration Guide

Can be viewed on - https://wix-yoshi-website-pr-2565.surge.sh/docs/next/library-flow/overview/

PR #2566, opened 2020-06-14
[Library Flow]: Add target for case of a node library

ranyitz:
### 🔦 Summary
The library flow was designed with browser libraries in mind, but when thinking about it, the only change needed to properly support node libraries would be to remove the `esm` directory. Everything else can stay the same.

This PR adds the `target` configuration option which is configured to be `web` by default but can be configured to `node`. This would cause the `esm` directory to not be created, leaving only the `cjs` and the `types` directories.

This change is controversial, so i'll try to add some pros and cons:

**disadvantages:**
1. Two different targets can confuse developers. (While we use sensible defaults ["web"] it is still more configuration)
2. Similar behavior can be achieved with running `tsc`/`jest`/`eslint`. (sort of, we still have some features and configurations OOTB which will take some time to achieve, without talking about future enhancements)
3. The `tsconfig.json` looks like it is configured for `esm` when no `esm` directory is created (This can be changed in the future as the tsconfig enforcer can be configurable)
4. The code looks quite ugly - Some areas in the code needs to be refactored, I'm talking about the `tsc-process` and the `dev-environment`. This is not in the scope of this PR but it's definitely doable.

**advantages:**
1. Very small change to catch a wide usecase. I don't know if I would like to maintain another toolkit for it.

______________________

This change needs to be added to the docs, I'm waiting for #2565 to be merged and i'll add docs for this functionality.


ranyitz:
@yavorsky good point. As the user communicates with its target we can do whatever we want. 👍 
supporting `esm` for node could be a nice feature

PR #2567, opened 2020-06-15
Upload source maps to sentry

ronami:
### 🔦 Summary

This PR adds integration with Sentry by uploading [source maps](https://docs.sentry.io/platforms/javascript/sourcemaps) while running in master CI.

This is currently somewhat experimental and enabled specifically for Thunderbolt. I used `process.env.ARTIFACT_VERSION` to tag the release with a unique build number. Sentry injects a global `SENTRY_RELEASE` variable which is checked in [Sentry's client](https://github.com/getsentry/sentry-javascript/blob/d4a9c065c4d8be8107c28e67944addc4386fc81c/packages/browser/src/sdk.ts#L83). This will add the version tag to errors and link them with the matching source maps.

### DIFF
--- a/packages/yoshi-flow-monorepo/src/webpack.config.ts
+++ b/packages/yoshi-flow-monorepo/src/webpack.config.ts
@@ -21,6 +21,7 @@ import { STATICS_DIR, SERVER_ENTRY, SRC_DIR } from 'yoshi-config/build/paths';
 import ManifestPlugin from 'yoshi-common/build/manifest-webpack-plugin';
 import { isObject } from 'lodash';
 import SentryWebpackPlugin from '@sentry/webpack-plugin';
+import { getProjectArtifactVersion } from 'yoshi-helpers/utils';
 import { PackageGraphNode } from './load-package-graph';
 import { isThunderboltElementModule, isThunderboltAppModule } from './utils';
 
@@ -121,7 +122,7 @@ export function createClientWebpackConfig(
       clientConfig.plugins!.push(
         new SentryWebpackPlugin({
           include: path.join(pkg.location, STATICS_DIR),
-          release: process.env.ARTIFACT_VERSION,
+          release: getProjectArtifactVersion(),
         }),
       );
     }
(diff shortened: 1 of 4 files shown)


packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -114,6 +116,18 @@ export function createClientWebpackConfig(
     clientConfig.optimization!.runtimeChunk = false;
   }
 
+  if (isThunderboltAppModule(pkg)) {
+    if (inMasterTeamCity()) {
+      clientConfig.plugins!.push(
+        new SentryCliPlugin({
+          include: pkg.location,
+          ignore: ['node_modules', 'dist'],
+          release: process.env.ARTIFACT_VERSION,
```
ranyitz, line 124: We have a function that strips the `-snapshot` suffix. Also in the future we'll replace this to use fed-ci package and it will be easier to do this from a single place

https://github.com/wix/yoshi/blob/4bc9391c74aacef7936b6cf5f082623880eb7ecd/packages/yoshi-helpers/src/utils.ts#L203-L209

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -114,6 +116,18 @@ export function createClientWebpackConfig(
     clientConfig.optimization!.runtimeChunk = false;
   }
 
+  if (isThunderboltAppModule(pkg)) {
+    if (inMasterTeamCity()) {
+      clientConfig.plugins!.push(
+        new SentryCliPlugin({
+          include: pkg.location,
```
ranyitz: Very small, but can be directed to `dist/statics`

packages/yoshi-flow-monorepo/src/webpack.config.ts
```diff
@@ -114,6 +116,18 @@ export function createClientWebpackConfig(
...
+          ignore: ['node_modules', 'dist'],
```
ranyitz: Actually ignoring `dist` will be problematic as the `map` files are located in `dist`

ronami: yeah, realized it right after we talked 😅

PR #2569, opened 2020-06-15
also minify site-assets node compilation

ranyitz:
### 🔦 Summary

Apply minifying for case of site-assets node compilation.

__________________

```diff
+ minification
- sourcemaps
```
After minifying & [no sourcemaps](https://github.com/wix/yoshi/pull/2568 )

image

```diff
+ minification
+ sourcemaps
```

The actually result of this PR without #2568
image

_________

As a remainder:

```diff
- minification
+ sourcemaps
```
image

```diff
- minification
- sourcemaps
```
image


ranyitz:
@ronami I agree, and in this case we even had a flag for it :)

PR #2571, opened 2020-06-15
Add an Editor flow logger

yavorsky:
### 🔦 Summary
This PR contains logger extension to ensure we are logging everything related to the flow.

Currently, we have a lot of stuff not related to the flow, so the info we are showing for developers during the build in most cases is useless:

- Showing network `ip` is not useful for editor flow apps
- Local url doesn't contain bundle names and just showing a root. We can explain it better
- `To create a production build, use npm run build.` - we don't want people to build it locally at all
- Served apps always has `http` url.

Moreover we can extend it:
- Split logs by components
- Preview start url we are opening with local overrides
- Make these urls short. Have a fallback for terminals which are not supporting hyperlinks
- Made any changes to the logs easily by controlling it on the editor environment

This is a proposed solution:

*Terminal which supports hyperlinks:*
Screen Shot 2020-06-15 at 5 19 53 PM

*Terminal which doesn't*
Screen Shot 2020-06-15 at 5 20 28 PM

In future we can easily update and maybe use https://github.com/vadimdemedes/ink with some loaders, animations, etc.


yavorsky:
Screen Shot 2020-06-16 at 7 21 31 PM
Latest update

PR #2575, opened 2020-06-16
Simplify experiments using `scope` prop in flow-bm

amitdahan:
### 🔦 Summary
Experiments are ready for consumption in pages & exported-components.
This is purely an internal change.

Spoke with @Gongreg about accessing experiments in module-level code (`moduleInit`, or in methods), sounds like it isn't necessarily the desired best-practice, so we'll leave this out for now.

### DIFF
--- a/packages/yoshi-flow-bm-runtime/src/experiments.tsx
+++ b/packages/yoshi-flow-bm-runtime/src/experiments.tsx
@@ -4,10 +4,8 @@ import { ExperimentsProvider } from '@wix/wix-experiments-react';
 export { useExperiments } from '@wix/wix-experiments-react';
 
 export function createExperimentsProvider(experimentsScopes: Array<string>) {
-  const experimentsProp = { scope: experimentsScopes };
-
   const Provider: FC = ({ children }) => (
-    <ExperimentsProvider options={experimentsProp}>
+    <ExperimentsProvider scope={experimentsScopes}>
       {children}
     </ExperimentsProvider>
   );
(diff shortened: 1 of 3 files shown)


packages/yoshi-flow-bm-runtime/src/experiments.tsx
```diff
@@ -4,10 +4,8 @@ import { ExperimentsProvider } from '@wix/wix-experiments-react';
 export { useExperiments } from '@wix/wix-experiments-react';
 
 export function createExperimentsProvider(experimentsScopes: Array<string>) {
-  const experimentsProp = { scope: experimentsScopes };
-
   const Provider: FC = ({ children }) => (
-    <ExperimentsProvider options={experimentsProp}>
+    <ExperimentsProvider scope={experimentsScopes}>
```
yanivefraim, line 8: are you sure this is a valid API?
I cannot find `scope` as prop in the [docs](https://github.com/wix-private/fed-infra/tree/master/experiments/wix-experiments-react#usage) or in the [source](https://github.com/wix-private/fed-infra/blob/master/experiments/wix-experiments-react/src/index.ts#L8)

Gongreg: Hey, they are actually working on a branch: https://github.com/wix-private/fed-infra/tree/wix-react-experiments-16/experiments/wix-experiments-react

amitdahan: What @Gongreg said :D

yanivefraim: Oh, so this one will not work, until we merge, right?

amitdahan: Oh no, what he meant was, `wix-experiments-react` use a secondary branch `wix-react-experiments-16` to release the latest major `^4`.
They use `master` for `^3`.

PR #2577, opened 2020-06-16
Avoid creating changes that are just a reordering of the references

ob1:


packages/yoshi-flow-monorepo/src/verify-typescript-references.ts
```diff
@@ -16,26 +16,34 @@ export default async function verifyTypeScriptReferences({
         .map(([depName]) => {
           return path.relative(pkg.location, graph.get(depName)!.location);
         })
+        .sort()
         .map((relativePath) => ({ path: relativePath }));
 
+      const originalReferencesSorted = [...(tsconfig.references || [])]
+        .map(({ path }) => path)
+        .sort()
+        .map((path) => ({ path }));
```
ranyitz, line 25:
```suggestion
sort((a,b) => a.path > b.path ? 1 : -1)
```

obviously not critical

ob1:
> LGTM
> 
> It just means that when it needs to update, it will also sort it, which is ok in my opinion.

Yeah, the sorting was intended - I figured you'd even want it :)

PR #2582, opened 2020-06-16
Support optional chaining and nullish coalescing in development

ranyitz:
### 🔦 Summary
Currently it's not possible to use `optional-chaning` or `nullish coalescing` because `yoshi start` fails.

The explanation is that during development we specify `last 2 chrome versions, last 2 firefox versions` for the target. Those target already implemented those feature, and because of that it doesn't get transpiled.

The problem here is that webpack@4 is using Acorn@6 which doesn't support this when it parses the code.

This [will be available on webpack@5](https://github.com/webpack/webpack/issues/10227#issuecomment-642734920)

As a workaround we're always adding those plugins, even in development.


ranyitz:
@ronami Very cool! I didn't know that. They didn't have @yavorsky so they actually imported those packages instead of just use the `include` which is nicer :)

PR #2585, opened 2020-06-17
Add `appDefId` config option

amitdahan:
# 🔦 Summary
As suggested by @Gongreg, this makes sense since users will have to react to the recent BM changes, and they won't necessarily understand how to do that using the flow.

This PR also clarifies some config options that only affect the generated `target/module_ID.json` file.


Gongreg:
Awesome!

PR #2586, opened 2020-06-17
Upgrade link to polyfill v3

sidoruk-sv:
### 🔦 Summary

added scopes includes (details at: https://polyfill.io/v3/url-builder/):
* `es2018`
  * Promise.prototype.finally
  * Symbol.asyncIterator
* `es2019`:
  * Array.prototype.flat
  * Array.prototype.flatMap
  * Object.fromEntries
  * String.prototype.trimEnd
  * String.prototype.trimStart

query parameters described: https://polyfill.io/v3/api/

Ready for discussion here.

Related discussion in Slack: https://wix.slack.com/archives/C0NJPS8N5/p1589958546476500

FED handbook already has updated link: https://github.com/wix-private/fed-handbook/blob/master/POLYFILL.md

Also, want to raise a discussion on:
**"If we have polyfill.io service - could we change tsconfig and Babel.js `target` to compile code to es2019"**
(for example: for now, fullstack template has `"target": "es5`)


ranyitz:
Thanks @sidoruk-sv it looks great.

Regarding babel-preset-yoshi. We would like to configure it according the the platform it will be used in.

This is surely a possible optimization, but it is a possible breaking change of production so we need to do this very carfully.

Please open a dedicated issue to discuss it. Hopefully we'll be able to do that for TypeScript as well.

rudnitskih:
@ranyitz @sidoruk-sv What do you think to avoid `if` inside the script for `min` version? I don't any reason to load not a minified version of polyfill. So keep just such URL:

`<script src="https://${clientTopology.staticsDomain}/polyfill/v3/polyfill.min.js?features=default,es6,es7,es2017,es2018,es2019&flags=gated&unknown=polyfill"></script>`

PR #2587, opened 2020-06-17
Allow pass Wix instance for Wix provider to prevent loading original one

yavorsky:
### 🔦 Summary
It's a follow-up for https://github.com/wix-private/tpa-settings/pull/32

We are working on merging tpa-settings into editor flow, but currently we need to have a good way to test it.
Since we want to use Wix SDK testkit to test `tpa-settings` correctly, it would be good to allow use `WixSDKProvider` where we can pass Wix mock with needed configuration for specific test.


packages/yoshi-flow-editor-runtime/src/react/SDK/WixSDKProvider.tsx
```diff
@@ -15,19 +15,26 @@ declare global {
...
 
-export class WixSDKProvider extends React.Component<{}, IState> {
+export class WixSDKProvider extends React.Component<IProps, IState> {
   state: IState = {
     value: defaultWixSDKContext,
   };
 
   componentDidMount() {
-    const editorSdkScriptSrc = DEFAULT_WIX_SDK_SRC;
+    if (this.props.Wix) {
```
ranyitz, line 32: I would add a comment here

PR #2592, opened 2020-06-17
Internal: fix typing problem

yanivefraim:
`getProjectArtifactVersion` used `as string`, when it should be actually returning `string | undefined`; 

- removed the `as string`
- fixed the place where we use it


packages/yoshi-helpers/src/utils.ts
```diff
@@ -226,7 +226,7 @@ export const getProjectCDNBasePath = (useUnversionedBaseUrl: boolean) => {
     // Webpack's `publicUrl` to use the first option.
     artifactPath = 'dist';
   } else {
-    artifactPath = getProjectArtifactVersion();
+    artifactPath = getProjectArtifactVersion() || '';
```
amitdahan, line 229: future-is-now

```suggestion
    artifactPath = getProjectArtifactVersion() ?? '';
```

PR #2595, opened 2020-06-18
Re-export everything from `@wix/wix-experiments-react`

amitdahan:
### 🔦 Summary
This will allow users to drop the `@wix/wix-experiments-react` dependency, and instead import types, providers, hooks and HOCs, straight from `yoshi-flow-bm-runtime`.


packages/yoshi-flow-bm-runtime/src/experiments.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { FC } from 'react';
 import { ExperimentsProvider } from '@wix/wix-experiments-react';
 
-export { useExperiments } from '@wix/wix-experiments-react';
+export * from '@wix/wix-experiments-react';
```
amitdahan, line 4: Explicitly re-export only the things I have a use-case for (even if it's all of them)

amitdahan:
Add docs & template usage

amitdahan:
bm-flow template is failing since `useExperiments` couldn't find an active Context :-(
I'm debugging.

PR #2598, opened 2020-06-18
Implement automatic translations solution for Editor flow

yavorsky:
### 🔦 Summary
This PR will reduce boilerplate from user's side related to translations.
We are going to have a provider on a flow side and user shouldn't configure it at all.

The only configuration we are allowing: default (fallback) language and disable automatic provider injecting, so users can migrate gradually w/o breaking anything. (In fact, everything still works, but 2 requests will be sent)

... More description will be added today

TODO:
- [x] Update docs
- [x] Test on one of real time apps using this branch


yavorsky:
This PR faced 3 random types of test errors during last 3 days being opened.. started from wix-base-ui dependency, finishing with stylable one...

packages/yoshi-flow-editor-runtime/src/SettingsWrapper.tsx
```diff
@@ -18,23 +21,47 @@ declare global {
...
+    <I18nextProvider
+      i18n={i18n({
+        language: getLanguageWithInstance(Wix),
+        waitForReact: true,
+      })}
+    >
+      {children}
+    </I18nextProvider>
+  ) : (
+    (children as React.ReactElement)
```
amitdahan, line 38: Is this necessary? Isn’t this covered by the `FC` type?

packages/yoshi-flow-editor-runtime/src/i18next/index.ts
```diff
@@ -2,16 +2,27 @@ import memoize from 'lodash/memoize';
...
+try {
+  fallbackTranslations = require('assets/locales/messages_en.json');
+} catch (e) {
+  console.error(e);
+}
+
 export function getSiteTranslations(
   language: string,
 ): Promise<Record<string, string>> {
+  console.log('here!');
```
amitdahan, line 23: 😬

website/docs/editor-flow/structure-api/application.md
```diff
@@ -118,3 +124,10 @@ export const getAppManifest = async () => {
 };
 ```
 
+## Assets
+A place for static files like images, fonts, translations.
+
+### Locales
+Directory where all locales are located. Editor flow will expect all locales to be named as `messages_:LANGUAGE`.
+Following the naming convention will help editor flow to locate needed locale files and load it w/o asking user to set up a provider and just directly use
```
amitdahan, line 132: 🙃🙃

package.json
```diff
@@ -151,7 +151,8 @@
     "wix-eventually": "^2.3.0"
   },
   "resolutions": {
-    "@types/node": "^12.0.0"
+    "@types/node": "^12.0.0",
+    "**/@types/i18next": "8.4.5"
```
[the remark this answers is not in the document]
yavorsky: I didn't found it :( 
`@types/i18next-react` have `*` in dependencies, which causes latest version be installed, which contains breaking changes

packages/yoshi-flow-editor-runtime/src/SettingsWrapper.tsx
yavorsky: no :( https://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wFgAoCtCAOwGd4AZFATwgFcYAuRYmAOgBiAYTgBeOAAowOMHQCU4gHxSKcdXBkQ5-NAAtgAGwAmRGgG4K8oA

packages/yoshi-flow-editor-runtime/src/i18next/index.ts
yavorsky: oops thanks!

packages/yoshi-flow-editor-runtime/src/SettingsWrapper.tsx
amitdahan: I fixed it by wrapping with a Fragment: `<>{children}</>` but I'm not sure that's any better 😅

package.json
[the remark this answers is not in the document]
amitdahan: I'm just wondering if this'll eventually clash with the indirect types I'm gonna get by installing `@wix/wix-i18n-config` (since those are using latest).
Can a yarn `resolution` appear in a child `package.json`? That'll work if we encounter that right?

PR #2602, opened 2020-06-20
Clean BM Flow template

amitdahan:
### 🔦 Summary
Following our meeting (@yavorsky & @ranyitz) to make sure our flow's template is "clean", this takes care of most of the simple changes we listed.

### ✅ Checklist
- [x] ~~Remove `PORT=5000`~~ Point `--url` to BM testkit (and leave `PORT=5000` until we remove it).
- [x] Use yoshi-bm instead of yoshi
- [x] Remove unused deps:
  - [x] `wix-axios-config`
  - [x] `wnp-bootstrap-grpc`
  - [x] `wix-experiments`/`wix-experiments-react`
  - [x] `axios`
  - [x] `i18next`/`react-i18next`/`i18next-xhr-backend`
- [x] Move deps to the flow:
  - [x] `yoshi-style-dependencies`
  - [ ] `yoshi/types.d.ts` -> `yoshi-flow-bm/types.d.ts` (#2604)
- [x] Aligning templates & docs:
  - [x] `src/components` -> `src/exported-components`
  - [x] Remove `app-config-templates` dir
- [x] Generalize template:
  - [x] Remove all todo leftovers
  - [x] Remove "content" code that's not demonstrating anything of use (hooks dir)
  - [x] Leave only 1 example method
  - [x] Remove `src/moduleInit.ts`


packages/yoshi-flow-bm/package.json
```diff
@@ -23,9 +23,9 @@
     "webpack": "4.43.0",
     "yoshi-common": "4.103.3",
     "yoshi-config": "4.103.3",
-    "yoshi-flow-bm-runtime": "4.102.0",
```
ranyitz, line 26: Verify that we're not using the runtime (maybe dev dependency)

PR #2604, opened 2020-06-20
Refactor `yoshi/types.d.ts` to be shared across the different flows

amitdahan:
### 🔦 Summary
We expose custom extension types (`.css`, `.jpg` etc) through `yoshi/types.d.ts`.
This previously forced us to copy over its content for the various flows to not depend on `yoshi`.
This moves said extensions to `yoshi-common`, as I feel that's the source-of-truth of our "file extension support".

Since `yoshi-common` is used in all flows, this makes sense.

Additionally, this PR simplifies consumption of environment `.d.ts` files in an overall more succinct (IMO) way, when projects already have an `external-types.d.ts`.
As a side-effect, this has the benefit of relying on node's resolution algorithm for resolving the `.d.ts`'s correct path, in cases where the `.d.ts` file is placed in a different `node_modules` (monorepos?).

In the library-flow template we are only using `jest-yoshi-preset/types.d.ts`, and so creating an `external-types.d.ts` file just to consume that 1 `.d.ts` seems a bit more verbose (@ranyitz wdyt? should I align the library-flow template for consistency?).


amitdahan:
I couldn't find references as to why Next.js (for example) chose this approach 😢.

amitdahan:
> I'm not sure if /// is better than just using imports. Seems like IDE can help with modules that don't exist, but I'm really ok with current solution

I think I'm just gonna blindingly trust that it's preferred for some reason which I cannot find 😄 
Both are better than today, so I'll merge and we'll be on the lookout for disadvantages 😅

PR #2605, opened 2020-06-21
Yoshi serverless: ambassador externals

yanivefraim:
We want:

1. All Yoshi Serverless dependencies to be non externals (we want to bundle everything, so there will be no installation in serverless)

2. Ambassador RPC deps to be external (installed on serverless and consumed from it)


packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -5,10 +5,25 @@ import { Config } from 'yoshi-config/build/config';
 import { getServerlessScope } from 'yoshi-helpers/build/utils';
 
 export default async function publishServerless(config: Config) {
-  console.log('Publishing to Serverless');
   if (!config.yoshiServer || !process.env.EXPERIMENTAL_YOSHI_SERVERLESS) {
     return;
   }
+  console.log('Publishing to Serverless');
+  const packageJson = JSON.parse(
+    fs.readFileSync(path.resolve('package.json'), 'utf-8'),
```
ronami, line 13: can probably also `require()`

packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -19,6 +34,12 @@ export default async function publishServerless(config: Config) {
   );
 
   fs.copySync(path.resolve('serverless'), path.resolve('temp/serverless'));
+  if (dependencies) {
+    fs.writeFileSync(
+      path.resolve(`temp/serverless/${getServerlessScope()}/package.json`),
+      JSON.stringify({ dependencies }),
```
ronami, line 40: I think we can add `JSON.stringify('...', null, 2)` for some extra readability

packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -19,6 +34,12 @@ export default async function publishServerless(config: Config) {
   );
 
   fs.copySync(path.resolve('serverless'), path.resolve('temp/serverless'));
+  if (dependencies) {
```
ronami, line 37: Object.keys().length

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -1224,6 +1224,16 @@ export function createBaseWebpackConfig({
                 return callback();
               }
 
+              // Everything in Yoshi Serverless (but ambassador deps) should be bundled
+              if (useYoshiServer && process.env.EXPERIMENTAL_YOSHI_SERVERLESS) {
```
ronami, line 1228: Since we're checking this combination in a few places, maybe we can add a prop to our config that says `isYoshiServerless` or something

yanivefraim: Cool, I'll do it in another PR

PR #2606, opened 2020-06-21
Add types to `methods` & `moduleInit` in BM flow

amitdahan:
### 🔦 Summary
This adds types to:

- [x] `src/moduleInit.ts`
- [x] `src/methods/**/*.ts`

### 🗺️ Test Plan
Will add temporary usages for methods and moduleInit in template to make sure type-check passes.


amitdahan:
@yanivefraim ping 😅

amitdahan:
@yanivefraim ping #2

packages/yoshi-flow-bm-runtime/src/types.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface ModuleInitOptions {
+  module: BusinessManagerModule;
+  moduleParams: IBMModuleParams;
+}
+
+export type ModuleInitFn = (this: any, options: ModuleInitOptions) => void;
+
+export interface MethodOptions {}
+
+export type MethodFn = (
```
yanivefraim, line 13: the signature is different from the source, is it in purpose? (it was returning any, now it is void)

amitdahan: Cool nice catch! `void` is incorrect, I'll fix back to `any`.

Regarding `options`, yeah it seems I was missing that in the `.bind` call, I only bound it to `this`.

After I fix the runtime code to also bind `{ module, moduleParams }` as the first argument, and also fix the types to be able to return `any`thing, it'll be good, right? wdyt?

PR #2608, opened 2020-06-22
Yoshi Serverless: fix local work

yanivefraim:
This PR enables working locally with Yoshi Serverless. We want to be able to work with production code, overriding our local code. This creates a CORS issue (we have a call from `www.wix.com` to `localhost:port...`. 
Solution: enable CORS for `*`, in dev mode 

* update baseUrl according to local / prod env


packages/yoshi-common/src/templates/serverless/serverless.js
```diff
@@ -1,12 +1,45 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
ronami, line 1: if we have more files like these maybe we can have an override so all files in `templates` have this rule ignored.

packages/yoshi-common/src/templates/serverless/serverless.js
```diff
@@ -1,12 +1,45 @@
...
+const accessControlHeaders = {
+  headers: {
+    'Access-Control-Allow-Origin': '*',
+    'Access-Control-Request-Method': '*',
+    'Access-Control-Allow-Headers':
+      'origin, x-requested-with, content-type, accept, x-wix-scheduler-instance, authorization',
+  },
+};
+
+const isDevelopment = () => process.env.NODE_ENV === 'development';
```
ronami, line 15: I don't think this needs to be a function

packages/yoshi-common/src/templates/serverless/serverless.js
ronami: this file can probably be a ts file with imports since we bundle it

packages/yoshi-common/src/templates/serverless/serverless.js
```diff
@@ -1,12 +1,43 @@
...
+          body: {},
+          ...(process.env.NODE_ENV === 'development'
+            ? accessControlHeaders
+            : {}),
+        });
+      });
+};
+
+module.exports = (functionsBuilder) => {
+  const builder = addOptionsCors(functionsBuilder);
```
ronami, line 30: really small but I think it more clear if we had:

```js
if (development) {
  addOptionsCors(...)
}
```

packages/yoshi-common/src/webpack-utils.ts
```diff
@@ -133,7 +133,9 @@ function createServerEntries(context: string, cwd: string = process.cwd()) {
   entries['routes/_api_'] = yoshiServerEntry;
 
   if (isDevelopment) {
-    entries['routes/_launchEditor_'] = 'yoshi-server/build/routes/launchEditor';
+    entries['routes/_launchEditor_'] = process.env.EXPERIMENTAL_YOSHI_SERVERLESS
```
ronami, line 136: nice catch 👍

PR #2610, opened 2020-06-22
fix a broken checksum in yarn lock

ranyitz:
### 🔦 Summary
We've seen this error both locally and on CI

image

I ran `yarn --update-checksums` and pushed the fixed version here.


ranyitz:
It seems that this break was solved by this PR - https://github.com/wix/yoshi/pull/2607

I guess that there is some problem with node_platform assume that everybody uses `latest` in all of their dependencies while some of their users actually uses lock files.

Closing

PR #2613, opened 2020-06-22
Move TpaComponentsProvider and ExperimentsProvider from user's side

yavorsky:
### 🔦 Summary
This PR removes Providers from Widget part. The only Provider left is BILoggerProvider. It requires additional logic, where user have to configure a schema package, so I would to move it out in separate PR.

So for now, all apps can avoid using and maintaing:
- `ExperimentsProvider`
- `TPAComponentsProvider`
- `i18NextProvider` (added in https://github.com/wix/yoshi/pull/2598)

Moreover, we are going to provide `withExperiments` HOC from editor flow. It will add `experiments` property and include experiments from app's configuration.

As a part of this PR, I have to add `WithProviders` HOC for internal usage, so we can reuse it for Settings also.

I didn't include Settings changes, since we are going to create new PR with tpa-settings migration, which won't contain additional boilerplate. 

Before merge TODO:

- [x] Update docs.


packages/create-yoshi-app/templates/flow-editor/typescript/src/components/{each%flowData.components:name%each}/Widget/Widget.tsx
```diff
@@ -20,27 +15,19 @@ const biLogger = initSchemaLogger(webBiLogger);
...
+    const { appName } = this.props;
 
     return (
       <BILoggerProvider logger={biLogger}>
-        <ExperimentsProvider options={{ experiments }}>
-          <TPAComponentsProvider value={{ mobile }}>
-            <Widget appName={appName} />
-          </TPAComponentsProvider>
-        </ExperimentsProvider>
+        <Widget appName={appName} />
```
amitdahan, line 30: 😍😍 so clean

PR #2618, opened 2020-06-23
monorepoflow – remove stripOrganization from the error logs

tuchk4:
Current error message is:

image

Could not find an app with the name of `editor-elements-library` but it exists at the `Apps found:`

Error message with this PR is:

image

### DIFF
--- a/packages/yoshi-flow-monorepo/src/scripts/start.ts
+++ b/packages/yoshi-flow-monorepo/src/scripts/start.ts
@@ -67,7 +67,7 @@ const start: cliCommand = async function (argv, rootConfig, { apps, libs }) {
     return process.exit(1);
   }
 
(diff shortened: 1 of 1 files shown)


tuchk4:
Also the related issue:
to run package build we use `yoshi build package-name` (dir name)
to run package start we use `yoshi build @wix/package-name`  (pkg name)

let's choice the one style and I will update the PR.

tuchk4:
Updated:

now start and build commands do the same pkg filter.

image

yavorsky:
I would also make `Could not find`... red. It's not really obvious this message is explaining an error.

ronami:
This PR ended up being a breaking change for local development. Existing scripts need to be changed from `yoshi start @wix/app` into `yoshi start app`.

PR #2628, opened 2020-06-25
Add dockerignore and templates

yanivefraim:
We want to add a `.dockerignore` and `templates` folder, both mandatory for fullstack projects


packages/yoshi-common/src/copy-docker.ts
```diff
@@ -38,4 +43,13 @@ export default async function (config: Config, cwd = process.cwd()) {
...
     fs.outputFileSync(dockerDestinationPath, transformedContent);
   }
+
+  const dockerignoreDestinationPath = path.resolve(cwd, DOCKER_IGNORE);
+  const projectDockerignoreExists = fs.existsSync(dockerignoreDestinationPath);
+  if (!projectDockerignoreExists) {
+    const dockerIgnore = ['node_modules', 'target'];
+    fs.writeFileSync(dockerignoreDestinationPath, dockerIgnore.join('\n'));
+  }
+  const templatesFolderPath = path.resolve(cwd, TEMPLATES);
```
ronami, line 53: I think we should add a comment explaining why this is needed.

packages/yoshi-common/src/copy-docker.ts
```diff
@@ -38,4 +43,13 @@ export default async function (config: Config, cwd = process.cwd()) {
     );
     fs.outputFileSync(dockerDestinationPath, transformedContent);
   }
+
+  const dockerignoreDestinationPath = path.resolve(cwd, DOCKER_IGNORE);
+  const projectDockerignoreExists = fs.existsSync(dockerignoreDestinationPath);
```
ronami, line 48: small but we can have a function for `writeIfDoesntExists()`

packages/yoshi-common/src/copy-docker.ts
```diff
@@ -1,22 +1,27 @@
...
   if (!projectDockerExists) {
-    const dockerFileName = path.resolve(
+    const dockerFilePath = path.resolve(
       __dirname,
       `./templates/${DOCKER_FILE}`,
     );
 
-    const dockerDestinationPath = path.resolve(cwd, DOCKER_FILE);
-    fs.copyFileSync(dockerFileName, dockerDestinationPath);
+    fs.copyFileSync(dockerFilePath, dockerDestinationPath);
```
ronami, line 24: Docker has params in it (`MAINTAINER {%authorName%}`), I think we need to fill these, similar to how CYA does it. Maybe we can use the same utility.

packages/yoshi-common/src/copy-docker.ts
```diff
@@ -1,22 +1,27 @@
...
   }
-  const projectDockerExists = fs.existsSync(path.resolve(cwd, DOCKER_FILE));
+
+  const dockerDestinationPath = path.resolve(cwd, DOCKER_FILE);
+  const projectDockerExists = fs.existsSync(dockerDestinationPath);
   if (!projectDockerExists) {
-    const dockerFileName = path.resolve(
+    const dockerFilePath = path.resolve(
       __dirname,
       `./templates/${DOCKER_FILE}`,
```
ronami, line 21: small but we can use `path.resolve(__dirname, 'templates', DOCKER_FILE)`

packages/yoshi-common/src/copy-docker.ts
```diff
@@ -38,4 +43,15 @@ export default async function (config: Config, cwd = process.cwd()) {
...
   }
+
+  const dockerignoreDestinationPath = path.resolve(cwd, DOCKER_IGNORE);
+  const projectDockerignoreExists = fs.existsSync(dockerignoreDestinationPath);
+  if (!projectDockerignoreExists) {
+    const dockerIgnore = ['node_modules', 'target'];
+    fs.writeFileSync(dockerignoreDestinationPath, dockerIgnore.join('\n'));
+  }
+  // a templates folder is mandatory in case of a Docker
+  // (docker task will fail in case there's no such folder)
```
ranyitz, line 54: This is a bit sad :)

Thanks for the comment!

ronami: Yeah, Docker should probably copy it if it exists (not fail otherwise)

PR #2631, opened 2020-06-25
Fix testEnvBuilder naming lint error in server template

sidoruk-sv:
### 🔦 Summary
For now, this code will not pass eslint validation for naming.

image


amitdahan:
It’s interesting that our template tests didn’t catch up on this. 
Any idea how?
We lint (and we `—fix` even) after template generation (in template tests).

Maybe we’re not failing if there are still type errors after the `—fix`?
We’re also running `npm test` which should also run lint, but this assumes we’re running `npm test`, we could be calling `yoshi test` which would fail to find lint errors. 

I’ll check.

amitdahan:
Weird, it seems we're linting in our tests, so it's weird that this happened.
Regardless - let's merge this :)
Thanks!

PR #2635, opened 2020-06-25
Provide translations, experiments automatically.

yavorsky:
### 🔦 Summary

Since we already implemented tpa-settings to our template and reduced big amount of boilerplate from the Widget and controller part, let's pay attention to reducing boilerplate from Settings part.

In this PR we are:
- Removing experiments provider
- Removing translations provider
- Updating tests in template
- Adding translations consumer to the template so it will be consistent with Widget
- Added WixSDK testkit with `configure` prop, so it will be much easier to test Settings part with Wix SDK.


packages/create-yoshi-app/templates/flow-editor/typescript/src/assets/locales/messages_uk.json
```diff
@@ -1,4 +1,6 @@
 {
   "app.widget.welcome": "Вітаю",
-  "app.settings.label": "Налаштуй Віджет"
+  "app.settings.label": "Налаштуй Віджет",
```
bordi, line 3: ❤️

PR #2642, opened 2020-06-28
Jest connection error message

tzachbon:
### 🔦 Summary
Handle warning when user try to create a request to a url and gets 404 or 503.
You can read more about how puppeteer handles those cases [here](https://github.com/puppeteer/puppeteer/blob/v4.0.0/docs/api.md#event-requestfailed)

### 🗺️ Test Plan
Catch the warning log in the console after an error is thrown.


test/typescript/features/jest-preset-errors/e2e.test.ts
```diff
@@ -0,0 +1,23 @@
...
+describe.each(['dev'] as const)('e2e [%s]', (mode) => {
+  it('run tests', async () => {
+    await scripts[mode](async () => {
+      const warnMessage =
+        'Request failed or not found. url: http://localhost:3100/fake-url, status code: 404, method: GET';
+
+      try {
+        await scripts.test(mode);
+      } catch (e) {
+        expect(e.message.includes(warnMessage)).toBe(true);
```
ranyitz, line 19: Using something like `expect(e.message).toMatch(warnMessage)` would provide much better error message in case this will fail. I think it worth changing it.

test/typescript/features/jest-preset-errors/jest-yoshi.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
ranyitz, line 1: I'm not sure this file is necessary as the base one will do (in any case I'm not sure we need the `devtools: true` here)

test/typescript/features/jest-preset-errors/e2e.test.ts
```diff
@@ -0,0 +1,23 @@
+import Scripts from '../../../scripts';
+
+const scripts = Scripts.setupProjectFromTemplate({
+  templateDir: __dirname,
+  projectType: 'typescript',
+});
+
+jest.setTimeout(40000);
+
+describe.each(['dev'] as const)('e2e [%s]', (mode) => {
```
ranyitz, line 10: Can we use a more descriptive name for this test, like `show browser errors in terminal`

tzachbon:
> Hey, I think that @yanivefraim test a similar things. His way to test this was a bit nicer in my opinion, I think we can also add this as a test case to the `request-errors` directory - [#1885 (files)](https://github.com/wix/yoshi/pull/1885/files#diff-def1918137643573b5af34dcf080627c)
> 
> He mocked the global console and asserted on it from the test itself.

I like it too! Updating the tests.

PR #2645, opened 2020-06-29
Yoshi serverless - import fresh with _non_webpack_require

yanivefraim:
`import-fresh` stopped working, because we bundle all dependencies (when using Yoshi Serverless). The reason for that is that webpack doesn't know how to handle [`require(filePath)`](https://github.com/sindresorhus/import-fresh/blob/master/index.js#L31). 
This PR copies `import-fresh` code, using `__non_webpack_require__` instead of regular `require`, which solves the issue


packages/yoshi-serverless/src/non-webpack-require-fresh.ts
```diff
@@ -0,0 +1,34 @@
+const path = require('path');
+const resolveFrom = require('resolve-from');
+const parentModule = require('parent-module');
```
ronami, line 3: can probably be imports be not critical

ronami: also maybe add a comment of `copied from ...` so we can see source if needed

PR #2646, opened 2020-06-29
feat(jest-config): validate when user uses server without port

tzachbon:
### 🔦 Summary
Create new validator for jest config file: `jest-yoshi.config.js`.
When user try to use jest with server command he must add port to prevent unnecessary errors.

#### Expected
Exit the process with a clear error message.
```
        When using server.command in jest config you must specify your port.

        For example:

        module.exports = {
          server: {
            command: 'node dist/server.js',
            port: 3100,
          },
        }
```

#### Actual
The process finishes and the user probably get an error in his tests.

#### Examples

**Valid:** 
```js
{ 
	server: {
		command: 'node dist/server.js',
		port: 3100,
	},
}
```
```js
{ 
	server: {},
}
```

**Invalid:**
```js
{ 
	server: {
		command: 'node dist/server.js',
	},
}
```

### 🗺️ Test Plan
Test for the same output we the process exit when the user does not add port to `jest-yoshi.config.js`.

### DIFF
--- a/packages/yoshi-config/src/jest/validate-server-config.ts
+++ b/packages/yoshi-config/src/jest/validate-server-config.ts
@@ -1,0 +1,26 @@
+import chalk from 'chalk';
+
+export default function validateServerConfig(config: any) {
+  if (config?.server?.command && !config.server.port) {
+    console.error(
+      chalk.red(
+        `
+        ${chalk.bold(
+          `When using ${chalk.underline(
+            'server.command',
+          )} in jest config you must specify your port.`,
+        )}
+
+        ${chalk.underline('For example:')}
+        
+        module.exports = {
+          server: {
+            command: 'node dist/server.js',
+            ${chalk.bold('port: 3100')},
+          },
+        }`,
+      ),
+    );
+    process.exit(1);
+  }
+}
--- a/test/typescript/features/jest-yoshi-config/jest-yoshi-config.test.ts
+++ b/test/typescript/features/jest-yoshi-config/jest-yoshi-config.test.ts
@@ -1,0 +1,21 @@
+import Scripts from '../../../scripts';
+
+const scripts = Scripts.setupProjectFromTemplate({
+  templateDir: __dirname,
+  projectType: 'typescript',
+});
+
+jest.setTimeout(35000);
+
+describe.each(['prod', 'dev'] as const)('e2e [%s]', (mode) => {
+  it('run tests', async () => {
+    await scripts[mode](async () => {
+      try {
+        await scripts.test(mode);
+        throw new Error('Test should have failed but passed');
+      } catch (e) {
+        expect(e.message).toMatchSnapshot();
+      }
+    });
+  });
+});
--- a/packages/yoshi-config/src/jest/index.ts
+++ b/packages/yoshi-config/src/jest/index.ts
@@ -5,6 +5,7 @@ import importFresh from 'import-fresh';
 import { validate } from 'jest-validate';
 import validConfig from './validConfig';
 import { Config } from './config';
+import validateServerConfig from './validate-server-config';
 
 export default (): Config => {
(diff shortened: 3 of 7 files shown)


packages/yoshi-config/src/jest/validate-server-config.ts
```diff
@@ -0,0 +1,26 @@
...
+        
+        module.exports = {
+          server: {
+            command: 'node dist/server.js',
+            ${chalk.bold('port: 3100')},
+          },
+        }`,
+      ),
+    );
+    process.exit(1);
```
ranyitz, line 24: I prefer not to `process.exit` from this function, this is a node API which means that we don't know how the process that runs it prefer to handle the error.

I think we should do something similar to what `jest-validate` do, you can throw an error which doesn't have a stack trace/name and therefore looks nice. (The stack trace is not relevant in this case)

https://github.com/facebook/jest/blob/22b8e8514bc287de8b729c8f3e07a3210b5a0461/packages/jest-validate/src/utils.ts#L27-L38

test/typescript/features/jest-yoshi-config/jest-yoshi-config.test.ts
```diff
@@ -0,0 +1,21 @@
...
+});
+
+jest.setTimeout(35000);
+
+describe.each(['prod', 'dev'] as const)('e2e [%s]', (mode) => {
+  it('run tests', async () => {
+    await scripts[mode](async () => {
+      try {
+        await scripts.test(mode);
+        throw new Error('Test should have failed but passed');
```
ranyitz, line 15: It would be nicer to use https://jestjs.io/docs/en/expect#expectassertionsnumber in this case.

You assume that there is going to be a single assertion in this file, therefore you can write:

```ts
expect.assertions(1);
```

In case this passes, the test will fail as the assertion has never called.

packages/yoshi-config/src/jest/validate-server-config.ts
tzachbon: Unfortunately we can't do it.
When we throw an error we get this error message which generated by jest:
```
[16:21:27] Finished 'cdn' after 2060 ms
    ● Validation Error:
    
      Preset jest-yoshi-preset not found.
    
      Configuration Documentation:
      https://jestjs.io/docs/configuration.html
    
    jest failed with status code "1"
        at Scripts.test (/Users/tzachb/Desktop/Wix/yoshi/test/scripts.ts:370:13)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)

  console.log test/typescript/features/jest-yoshi-config/jest-yoshi-config.test.ts:17
    Error: [16:21:34] Starting 'cdn'...
        cdn is already running on 3200, skipping...
    [16:21:35] Finished 'cdn' after 624 ms
    ● Validation Error:
    
      Preset jest-yoshi-preset not found.
    
      Configuration Documentation:
      https://jestjs.io/docs/configuration.html
    
    jest failed with status code "1"
        at Scripts.test (/Users/tzachb/Desktop/Wix/yoshi/test/scripts.ts:370:13)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

After talking to @amitdahan we came to the understanding that this was resolved in version 25 and our users are using version 24

ranyitz: I understand. So in this case I prefer to do the same thing that we do with `jest-validate` which is to log the error and not to throw an error or exit the process.

tzachbon: And let it continue?

ranyitz: Yeah, it's currently the same thing we're doing with jest-validate. (We don't throw an error if there is a validation error but only logs it to the console)

We can change this decision and do it across the board, but for now I prefer not to deal with this. I think that we'll get the value from this PR either way

PR #2655, opened 2020-07-01
DOCS: Update migration-guide.md

sidoruk-sv:
### 🔦 Summary
Fix links and change link text

PR #2656, opened 2020-07-01
Update `migration-guide.md` (source)

sidoruk-sv:
### 🔦 Summary
Same as #2655 but now it is for the original file, not generated


amitdahan:
Thanks @sidoruk-sv 🥇

PR #2663, opened 2020-07-02
Add a BI logger support for editor flow

yavorsky:
### 🔦 Summary
This change adds BI logger to editor flow.

We are configuring BI logger under the hood and helping users just to consume it in controller, Settings panel and Widget (both viewer and editor).

We are also configuring BI logger according to user's role: owner or visitor.

### `.application.json`
Adding `bi` field enables automatic BI logger configuration.
Owner - passing to Settings panel consumer.
Visitor - controller and Widget.
In case of same loggers, you can just duplicate fields.

Screen Shot 2020-07-02 at 1 34 25 PM

Since in most cases owner and visitor is a different schemas and it is a `npm package` with some project specific methods.

### `controller.ts`
Here we are providing biLogger property with `flowAPI`. In controller we want to just log visitor events. We are setting some useful defaults taken from controller data. (`visitor_id`, `token`, 'appName', '_msid')

*controller.ts*
```ts
flowAPI.biLogger.logSomeEvent({ someBIField: 'someData' })

```

Moreover, we are allowing to extend defaults on user's side:
*controller.ts*
```ts
flowAPI.biLogger.util.updateDefaults({ someAdditionalField: 'someAdditionalData' })

```

### `viewer.app.ts`
The same as controller, you can access biLogger via `flowAPI` argument passed by the flow:
*controller.ts*
```ts
export initAppFoPage: InitAppForPageFn = (
   initParams,
   platformAPIs,
   namespaces,
   platformServices,
   viewerScriptFlowAPI
) => {
  flowAPI.biLogger.logSomeEvent({ someBIField: 'someData' })
}

```

### `Settings.tsx`
We are wrapping user's Settings component with BI provider and configured logger under the hood, so it's enough to just use `BILogger` consumer for it. It contains single argument - biLogger. For settings it's owner's one:

*Settings.tsx*
```tsx
<BILogger owner>{(bi) => (
  <Button onClick={() => {
    bi.logButtonClicked({ metaData: '123' })
  }} />
)}</BILogger>
```

### `Widget.tsx`
In Widget we want to log visitor BI events. To access it, we can use the same `BILogger` consumer as we have in Settings.

*Widget.tsx*
```tsx
<BILogger visitor>{(bi) => (
  <button onClick={() => {
    bi.logButtonClicked({ metaData: '123' })
  }} />
)}</BILogger>
```

### `BILoggerDefaults`
We want still to allow users to configure default values for logger in `Widget` and `Settings`.
```tsx
<BILoggerDefaults defaults={{ someDefault: 'value' }}>
  ...
  <BILogger visitor>{(bi) => (
    <button onClick={() => {
      bi.logButtonClicked({ metaData: '123' }) // `someDefault` will be also included to event
    }} />
  )}</BILogger>
```

### Types!
By just adding `bi` field to `.application.json`, you are receiving fully-typed logger according to your schema.

The implementation is tricky and requires some code generation by the flow, but it works perfectly. The only thing - you have to re-build the project when changing the bi-logger field.

Screen Shot 2020-07-02 at 1 33 25 PM


packages/yoshi-flow-editor-runtime/src/FlowAPI.ts
```diff
@@ -78,19 +90,39 @@ export class ControllerFlowAPI extends FlowAPI {
...
+    ) {
+      const biFactory = platformAPIs.biLoggerFactory();
+      const biOptions = {
+        visitor_id: platformBI.visitorId,
+        token: platformBI.biToken,
+        appName,
+        _msid: platformBI.metaSiteId,
+      };
+      this.biLogger = biLogger(biFactory)({});
+      if (this.biLogger) {
```
bodia-uz, line 121: in what case it can be not available?

packages/create-yoshi-app/templates/flow-editor/typescript/.application.json
```diff
@@ -11,6 +11,10 @@
   "translations": {
     "default": "en"
   },
+  "bi": {
+    "owner": "bi-logger-editor-flow-template",
```
bodia-uz, line 15: mb **widget & settings** instead of **visitor & owner**?
As for me it's more clear from the developer perspective

packages/yoshi-flow-editor-runtime/src/FlowAPI.ts
yavorsky: 👍

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
```diff
@@ -15,8 +15,10 @@ import {
   ExperimentsConfig,
   TranslationsConfig,
   DefaultTranslations,
+  BiConfig,
```
bordi, line 18: BiConfig  → BIConfig

packages/yoshi-flow-editor-runtime/src/i18next/index.ts
```diff
@@ -92,5 +93,7 @@ export default memoize(function i18n(params: I18nConfig) {
 });
 
 export const getLanguageWithInstance = memoize((Wix: IWixStatic) => {
-  return Wix.Utils.getLocale() || DEFAULT_LANGUAGE;
+  const queryParams: URLSearchParams = getQueryParams();
+  const multilingualLanguage = queryParams?.get('lang');
+  return multilingualLanguage || Wix.Utils.getLocale() || DEFAULT_LANGUAGE;
```
bordi, line 98: Fixed in another PR:
https://github.com/wix/yoshi/pull/2664

packages/yoshi-flow-editor/src/wrappers/commonViewerScriptWrapping.ts
```diff
@@ -38,13 +39,27 @@ const viewerScriptWrapper = (
     'viewerScript.js',
   );
 
+  let visitorBiLoggerPath = null;
+
+  if (model.biConfig?.visitor) {
+    visitorBiLoggerPath = resolveCwd.silent(model.biConfig.visitor);
+    if (!visitorBiLoggerPath) {
+      throw new Error(
+        `❌ Seems like you have \`bi.visitor\` specified, but didn't install it as a dependency.
```
bordi, line 48: 😍

packages/yoshi-flow-editor/src/wrappers/templates/SettingsAppEntryContent.ts
```diff
@@ -2,6 +2,7 @@ import {
   SentryConfig,
   TranslationsConfig,
   DefaultTranslations,
+  BiConfig,
```
bordi, line 5: BIConfig

packages/yoshi-flow-editor-runtime/src/i18next/index.ts
yavorsky: oops,  thanks..

packages/yoshi-flow-editor-runtime/src/EditorAppWrapper.tsx
yavorsky: What if I would say you that BiConfig is a config of 2 things - visitor and owner. The name is not related to Business intelligence

packages/yoshi-flow-editor/src/wrappers/biLoggerTypes.ts
```diff
@@ -0,0 +1,33 @@
...
+      fs.outputFileSync(resolvedDistConstants.biLoggerTypesFilename, template);
+    }
+  }
+};
+
+export const overrideBILoggerTypes = (biConfig: BIConfig) => {
+  if (biConfig?.owner || biConfig?.visitor) {
+    const template = biLoggerTypesTemplate({
+      visitor: biConfig.owner ? resolveCwd.silent(biConfig.owner) : null,
+      owner: biConfig.visitor ? resolveCwd.silent(biConfig.visitor) : null,
```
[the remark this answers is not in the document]
ranyitz: I see that it's already done [here](https://github.com/wix/yoshi/pull/2663/files#diff-15192b33a7288d23737d010cbf105ab7R48) maybe we can extract this into a function that handles both the loading and the correct error in case we haven't found it.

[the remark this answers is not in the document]
ranyitz: It will also remove this error handling logic from around 3-4 places in the application

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
```diff
@@ -43,54 +43,97 @@ export default () => (
 ```
 
 ## `BILogger`
+Renders a `children` function with `biLogger` relevant for current environment.
 
-Currently it consists of `BILoggerProvider` and `BILogger` components.
+You can configure 2 different kind of BI logger according to user's roles: owner and visitor:
```
ranyitz, line 48:
```suggestion
You can configure 2 different kinds of BI logger according to user's roles: `owner` and `visitor`:
```

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
```diff
@@ -43,54 +43,97 @@ export default () => (
...
 
-`BILoggerProvider` should be rendered in the root of you component and receive a `biLogger` prop.
+- **Owner** will be available in `Settings` panel.
```
ranyitz, line 50: I thought that the "site owner" is also the owner in the widget when it's in the editor. From the docs here it seems that it's not the case.

If the docs are describing it correctly, I agree with @bodia-uz regarding the naming (`settings`/`app`)

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
```diff
@@ -43,54 +43,97 @@ export default () => (
...
+- **Visitor** will be available in `Widget`, `controller` and `initAppForPage` in `viewer.app.ts`.
 
-`BILogger` is a consumer. It will render a `children` function with `biLogger` passed to provider.
+After you generate a project, a demo BI logger (`bi-logger-editor-flow-template`) will be added to `.application.json` configuration.
+It's a show case and should be finally repaced with user's BI logger schema.
```
ranyitz, line 54: We should probably mention to also remove this from `package.json` dependencies.

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
```diff
@@ -43,54 +43,97 @@ export default () => (
...
-
-By loading schema logger you've initialized and registered, you should use different loggers according to runtime environment:
-
-- Settings panel: `iframeAppBiLoggerFactory` imported from `@wix/iframe-app-bi-logger` package
-- Widget: `@wix/web-bi-logger` package
-
-_Settings.ts_
+**Settings.tsx**
+```tsx
+// Somewhere deep in the component
```
ranyitz, line 59:
```suggestion
// Somewhere deep in the component tree
```

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
```diff
@@ -43,54 +43,97 @@ export default () => (
...
 ```tsx
-import { WixSDK, BILogger, BILoggerProvider } from "yoshi-flow-editor-runtime";
-import { iframeAppBiLoggerFactory } from "@wix/iframe-app-bi-logger";
-import initSchemaLogger from "your-schema-logger-package";
+const createController = async ({ flowAPI, controllerConfig }) => {
+  const { setProps } = controllerConfig;
+
+  onSomeAction = async () => {
+    // Do something...
+    await flowAPI.biLogger?.somethigWasDone({});
```
ranyitz, line 80:
```suggestion
    await flowAPI.biLogger?.somethingWasDone({});
```

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
```diff
@@ -43,54 +43,97 @@ export default () => (
...
+      });
+      await flowAPI.biLogger?.templateWidgetLoaded({});
+    },
+  }
+}
+```
 
-const biLogger = initSchemaLogger(iframeAppBiLoggerFactory);
+### `BILoggerDefaults`
+To update defaults for each events being called from the `BILogger` render prop, you can update the context by wrapping your root component,
```
ranyitz, line 95:
```suggestion
To update defaults for each event being called from the `BILogger` render prop, you can update the context by wrapping your root component, 
```

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -132,16 +138,20 @@ function ooiControllerWrapper(
...
       ...userController,
       pageReady: async (...args: Array<any>) => {
-        // we are going to get rid of current setProps call and override original one with wrapper, where we can populate user's call with flow's fields.
+        // In future we are going to get rid of current setProps call and override original one with wrapper, where we can populate user's call with flow's fields.
         setProps({
           __publicData__: controllerConfig.config.publicData,
           _language: flowAPI.getSiteLanguage(),
           _translations: translations,
           _experiments: experiments.all(),
+          ...serializeBILogger(flowAPI.biLogger),
```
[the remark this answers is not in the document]
yavorsky: The issue here is that platform can only shallowly pass functions via `setProps`. We can't use some objects with it and have to use specific keys which will be grouped into object on Widget side

packages/yoshi-flow-editor/src/wrappers/biLoggerTypes.ts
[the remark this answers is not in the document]
yavorsky: I think it's 2 different use cases. We are trying to find it and save into entry-point during postinstall hook, start and build commands. But during npm install it's a lot of logs and just warn about missing dependency can be skipped by user.
But we want to warn user in console during runtime. If you think we need to break build if no bi loggers found, we can do it, but it could be a bit less consistent with other providers

website/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime.md
yavorsky: I prefer using owner / visitor naming b/c:
- It's naming used by BI devs and schema packages usually being separated by logging roles.
- In editor environment we're loading widget and controller and logging bi events the same we are doing it for viewer, but with editor scope. So in this case, in fact logs will be sent in `editor`, but package type will be `viewer`.
- Most of project's I've seen using this terminology instead of editor/viewer.
- We are trying to hide environment separation from user's side. For example when you are working with Widget, you are not maintaining 2 versions of it. The same for controller. I think we shouldn't spread any logic related to separating our app into 2 running environments.

As an alternative option, we can make it more module-oriented and allow pass any naming user want and then allow access it using this naming for controller or Render Prop: `<BILogger type="myLogger">`. It will add more responsibility, a bit more boilerplate on user's side and disable us from helping by passing needed provider, so we have to wrap everything in all providers for all loggers, which sounds not so cool also.

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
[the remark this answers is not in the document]
ronnyrin: you can have up to two levels of objects so you can do it

PR #2665, opened 2020-07-02
fix: requestfinished event -> check for response

tzachbon:
### 🔦 Summary
When there is no response on the request payload the `requestfinished` event breaks

[Slack](https://wix.slack.com/archives/CAL591CDV/p1593689270186900)


packages/jest-yoshi-preset/jest-environment-yoshi-puppeteer/index.js
```diff
@@ -52,14 +52,16 @@ module.exports = class PuppeteerEnvironment extends ParentEnvironment {
     });
 
     this.global.page.on('requestfinished', (request) => {
-      const hasError = [404, 503].includes(request.response().status());
+      const response = request.response();
```
ranyitz, line 55:
```suggestion
      // there are cases when the response object is undefined
      const response = request.response();
```

PR #2669, opened 2020-07-05
use experimentalMonorepoPublish

netanelgilad:
### 🔦 Summary
Try to move yoshi to use `experimentalMonorepoPublish`.

This will enable Deploy Preview publishes for yoshi packages, so easier testing of risky changes. It adds around 30-40 seconds to the PR build time.

Also, it should support everything that is done in the custom publish, so you'll be able to remove that code.

### 🗺️ Test Plan
N/A


ranyitz:
This is amazing!
Me and @netanelgilad talked about it and agreed that:

1. We want to create another parallel check in the CI for publishing. (currently all check perform publish)
2. We'll use a dedicated environment parameters to opt out from the publish process from the CI agents which aren't for publishing (which is not ideal but that way we won't need to perform a breaking change for other users using this feature)
3. We can remove our own script that was used for publishing and instead use [the CI's script for publishing monorepos](https://github.com/wix-private/wix-fed-scripts/blob/master/src/npm-ci-scripts/monorepoPublish.ts)

The only difference is that we used to verify that we can publish all of the packages before actually performing any publish which helped us with cases of partial publishes which lead to broken `npm install` on our users side.

If we'll see this in the future we'll need to perform a similar fix on the CI scripts.

ranyitz:
I added the `publish preview` check which is going to be the only check that publishes yoshi packages on PR

PR #2672, opened 2020-07-06
Deprecated Legacy Decorators Support

borislit:
### 🔦 Summary
After discussion with @ranyitz , this adds support for deprecated decorators proposal. This is required by Contacts, that uses Mobx, in order to be able to migrate to Yoshi 4

### 🗺️ Test Plan
...


ranyitz:
We prefer that people won't use syntax which is not part of the spec. Mostly because it diverge the codebases in Wix into 2 different types of syntax.

We understand that this is needed as a temporary thing. I really do hope that people won't abuse it to use legacy decorators in their projects.

PR #2673, opened 2020-07-06
Experimentally support skipping type check in monorepo flow

ronami:
### Summary

This PR adds support for not running `tsc -b` in monorepo flow's build script. This is an initial version of running TypScript's type-checks in a separate PR check.

Since the project I want to test this one consumes sources between packages, standard CI checks will simply build everything with Webpack's `ts-loader` which already skips type-checking. Only one CI check (publish check) will run `tsc -b` specifically for type-checking and for creating `dist` folders for every package before publishing them to `npm`. Because of technical difficulties, this project runs `tsc -b` on its own so we don't need to create a new  yoshi-monorepo-flow command specifically for type-checking yet.


packages/yoshi-flow-monorepo/src/scripts/build.ts
```diff
@@ -105,7 +105,9 @@ const build: cliCommand = async function (argv, rootConfig, { apps, libs }) {
     apps = apps.filter((app) => appNames.includes(stripOrganization(app.name)));
   }
 
-  await buildPkgs([...libs, ...apps]);
+  if (!process.env.SKIP_TYPE_CHECK) {
```
ranyitz, line 108: Not critical, but it's actually skipping build of the packages which also does TypeCheck.

Obviously not critical, but something to consider for the future

PR #2674, opened 2020-07-06
Add Sentry integration to BM Flow

amitdahan:
#### TODO:
- [x] Docs


packages/create-yoshi-app/templates/flow-bm/typescript/.module.json
```diff
@@ -3,6 +3,7 @@
   "experimentsScopes": [
     "{%projectName%}"
   ],
+  "sentryDsn": "{%sentryData.DSN%}",
```
yavorsky, line 6: maybe let's use the same config for both flows? we also have dsn, but also team name and few more useful fields:
https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/flow-editor/typescript/.application.json#L4

yavorsky, line 6: Maybe let's use the same configuration shape for both flows? We have few more auto-generated fields here that could be useful in future (for example for `open sentry` CLI command, etc)
https://github.com/wix/yoshi/blob/master/packages/create-yoshi-app/templates/flow-editor/typescript/.application.json#L4

packages/yoshi-flow-bm-runtime/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,14 @@
...
+  componentDidCatch(error: Error) {
+    this.props.sentryClient.captureException(error);
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
+
+export default withSentry(ErrorBoundary);
```
yavorsky, line 14: We have almost the same component for editor flow, would be cool to reuse it in future :)

website/versioned_docs/version-4.x/business-manager-flow/overview.md
```diff
@@ -188,6 +189,30 @@ export default () => {
...
+    } catch (error) {
+      sentryClient.captureException(error);
+    }
+  }
+
+  return (
+    <div />
+  );
+};
+```
```
yavorsky, line 214: I would not use a runtime usage example in place where we are trying to explain configuration.
Configuration example sounds more relevant to me in this place.
Also, we can add a part that this thing is going to be auto-generated

Moreover, I think `####` headers is not a part of right sidebar navigation, maybe it's worth some refactoring (not in this PR)

packages/yoshi-flow-bm-runtime/src/hooks/withSentry.tsx
```diff
@@ -0,0 +1,21 @@
...
+  sentryClient: BrowserClient;
+}
+
+const withSentry = <P extends InjectedSentryProps>(
+  Component: ComponentType<P>,
+) => {
+  const Wrapped: FC<Omit<P, keyof InjectedSentryProps>> = (props) => {
+    const sentryClient = useContext(SentryContext)!;
+
+    return <Component {...({ ...props, sentryClient } as P)} />;
```
yavorsky, line 15: `<Component {...props} sentryClient={sentryClient} />` should be a bit more performant and easier to read

Also, types should be auto-generated in this case, no?
(`props ` = `Omit<P, keyof InjectedSentryProps>`) & ( `sentryClient ` = `InjectedSentryProps['sentryClient']`)

packages/create-yoshi-app/src/createApp.ts
```diff
@@ -37,17 +45,27 @@ export default async ({
 
     // If we don't have template model injected, ask the user
     // to answer the questions and generate one for us
-    const runPrompt = require('./runPrompt').default;
+    const { default: runPrompt } = await import('./runPrompt');
```
yavorsky, line 48: 👏

packages/create-yoshi-app/templates/flow-bm/typescript/.module.json
amitdahan: My question is, what would a user do when he wants to opt-in to sentry integration if he didn't generate his project when sentry integration was built-in to the `prompts`?
I thought being able to just add your `DSN` would work for this, since we don't yet support a command for adding integrations after-the-fact (at least in BM).

packages/yoshi-flow-bm-runtime/src/hooks/withSentry.tsx
amitdahan: It _should_ work, but it errors :( 


Reproduced here: 
https://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAJQKYEMDGMA0cDecDCE4EAdkiTACoCeYS2AYvnAL5wBmURcA5FKhh4BuAFAjgFJFHbokcAMpEkMABYSA5rhFwdHCBABcvAEYoowkSzESYUmWjkBJEgCskGJABNFIZWpLqAApcYADOWjqhSqoaRj5+GqJWImikofAA7sCq8TEBcAC8cAA8gXBIAB62JJ7hzm4e3tH+QSGhAHwAFNoERJBkFEaExANUtEil7ZgiAJSF7RFwqSTpcADqUChgdJ5GTMUA8iDZpdgA1kjUEOxw9e62Tb55rRBh7QtFnWBtcwULOD1dEs0vAok8WoVcHojDxTOZWKJAbp+DAAK5QEglYb9cjwHAAOkJ31eoTYYISAQKOHJzzYAHp2qIdCxEToUejMRstjskkIgA

amitdahan: Also, found a few related issues:
- https://github.com/microsoft/TypeScript/issues/28884
- https://github.com/microsoft/TypeScript/issues/28938
(There are many duplicates / related issues, these are just examples)

amitdahan: I can change this to be:

```typescript jsx
return <Component {...(props as any)} sentryClient={sentryClient} />;
```
(`as P` could also work)
As it will keep runtime code more performant as you suggested.

After seeing similar issues, I'm sensing we won't be able to avoid _a_ workaround.

packages/create-yoshi-app/templates/flow-bm/typescript/.module.json
yavorsky: We can trigger
```
sentry: {
 dsn: value
}
```

to be valid enough to enable it.

And object structure will reduce breaking changes if we want to extend it with more fields

packages/yoshi-flow-bm-runtime/src/ErrorBoundary.tsx
amitdahan: Definitely!
I wonder if we already have an existing package that's a good fit to hold runtime commons or do we have a new use-case on our hands? 🤔

website/versioned_docs/version-4.x/business-manager-flow/overview.md
amitdahan: Hmm, so my takeaways from this are:
1. I'll extract all configurations to a separate `Configuration` page, which will not contain runtime examples.
1. I'll extract all runtime (examples mostly currently) related info to a `Runtimes` page similar to what you did in the Editor Flow.

These two will remove the unfavored runtime-example-in-config-doc usage, along with lowering the header-depth and solving the `####` headers not appearing in the sidebar.

I'll do this in a separate PR once this one is merged.

packages/create-yoshi-app/src/createApp.ts
amitdahan: Ended up reverting this :( It made the way we mock `./runPrompt` in `e2e.js` fail :(

PR #2675, opened 2020-07-06
Prevent cases of editor flow app generation w/o components.

yavorsky:
### 🔦 Summary
This PR fixes few issues related to generating a new project.

### Selecting empty app
If you generate a project and select 'Use already' registered app and it doesn't contain any components, currently it will just write `Aborting` since prompts w/o options will just exit the questions cycle. 
Here we are fixing it and allowing to call `next` method even if options list is empty.

So when you don't have any components and selected the app, we'll ask you to create new! Moreover, we will generate a platform component (with viewer and editor scripts if it doesn't exist).

### Throw an error instead of generating a default project if no components were selected
Since we are providing a lot of stuff, I would remove a fallback case if people just finished a registration w/o adding new components. In this case we'd rather throw an error instead of face cases when people trying to deal with unconfigured defaults.

### Ask to create a new component if nothing selected
In case user have selected already registered app, but didn't pick any of components in multi-select mode, we will ask to create new one. I can see cases when user want to use already created app w/o editor flow and add new components here instead of using current ones.


packages/create-yoshi-app/src/extended-prompts.ts
```diff
@@ -58,9 +58,12 @@ async function run<T>(
     if (question.getDynamicChoices) {
       question.choices = await question.getDynamicChoices(answers, context);
     }
-    const answer = await prompts([promptifyQuestion(question)], {
-      onCancel,
-    });
+    let answer = {};
+    if (!question.choices || question.choices.length) {
+      answer = await prompts([promptifyQuestion(question)], {
```
yavorsky, line 63: `prompts` will exit the chain by default if no choices provided.
Since we can have more steps via `next` method, which could return an array of next steps according to current selection, we will cover this case and won't allow `propmts` to exit the process.

packages/create-yoshi-app/src/dev-center-registration/runPrompt.ts
```diff
@@ -46,10 +47,19 @@ export default async (
       console.error(e);
       return fallback();
     }
+    console.log(e);
     console.log();
     console.log('Aborting ...');
     process.exit(0);
   }
 
+  if (!answers.components.length) {
```
yavorsky, line 56: We don't want to allow to create empty apps with empty `components`, `sled` directories and fedops config.

packages/create-yoshi-app/src/dev-center-registration/getQuestions.ts
```diff
@@ -406,6 +410,23 @@ export default (): Array<ExtendedPromptObject<string>> => {
...
+                );
+                const isEmptyApp =
+                  (isPlatformComponentRegistered &&
+                    context.app.components.length === 1) ||
+                  answers.components.length === 0;
+                if (
+                  isEmptyApp &&
+                  isOutOfIframe(context.templateDefinition.name)
+                ) {
+                  context.isPlatformComponentRegistered = isPlatformComponentRegistered;
```
yavorsky, line 425: Creating platform component in case of nothing created and skip if selected app already contains it

PR #2677, opened 2020-07-07
Remove lodash from the widget

yavorsky:
### 🔦 Summary
This PR removed lodash from the bundle.

Moreover we want to maintain same externals configuration and prevent duplicating lodash in node_modules for editor. So for this case we've added it as a `script` to vm file we are rendering (the same way we are doing for react, react dom).

The results are crazy:

From network:
Screen Shot 2020-07-07 at 4 12 11 PM
Screen Shot 2020-07-07 at 4 10 31 PM

From bundle analyze:
Screen Shot 2020-07-07 at 4 35 49 PM

Screen Shot 2020-07-07 at 3 15 32 PM


packages/yoshi-flow-editor/src/webpack.config.ts
```diff
@@ -48,29 +71,15 @@ export function createClientWebpackConfig(
     isHot,
     isAnalyze,
     forceEmitSourceMaps,
+    externalizeRelativeLodash: false,
```
ranyitz, line 74: Why false?

packages/yoshi-flow-editor/src/buildEntires.ts
```diff
@@ -13,12 +13,18 @@ const generatedWidgetEntriesPath = path.resolve(
...
-export const buildEditorPlatformEntries = (model: FlowEditorModel) => {
+export const buildViewerEntries = (model: FlowEditorModel) => {
   const componentEntries = writeComponentWrapping(
     generatedWidgetEntriesPath,
     model,
   );
 
+  return {
+    ...componentEntries,
+  };
```
amitdahan, line 24: Can this be just `return componentEntries;`?

PR #2678, opened 2020-07-08
Add module lifecycle hooks (`init`, `config`) to BM Flow

### LINKED ISSUE #2650: BM Flow doesn't support reacting to `config` events in module bundle
# Summary 
Same as we did for the `init` method in `BusinessManagerModule` bundles, we need to somehow support reacting to a `configModule` call, for our own module, using the [`BusinessManagerModule.config`](https://github.com/wix-private/business-manager/blob/master/business-manager-api/docs/business-manager-module.md#config) method.

This is currently not available to users in any way.

# Proposed Solution
1. Add a `moduleConfig.ts` convention that'll allow this to work exactly like it works for `moduleInit`.

2. A better approach, but probably more time-consuming would be to have a `module.ts` file that'll replace both `moduleInit.ts` and the mentioned `moduleConfig.ts` file, with 1 file that includes named-exports `init` & `config`:

```typescript
import { ModuleInitFn, ModuleConfigFn } from 'yoshi-flow-bm-runtime';

export const init: ModuleInitFn = ({ moduleParams }) => {
	// something something
};

export const config: ModuleConfigFn = ({ moduleParams }, sourceModuleId, configPayload) => {
	// something something
};
```

amitdahan:
### 🔦 Summary
Closes #2650.

This changes the `src/moduleInit` + `default` export convention for reacting to the module's `init` lifecycle hook to use named exports inside `src/module.ts`.

#### Example usage:

`src/module.ts`:

```typescript
import { ModuleInitFn, ModuleConfigFn } from 'yoshi-flow-bm-runtime';

export const init: ModuleInitFn = ({ module, moduleParams }) => {
  // ...
};

export const config: ModuleConfigFn = (
  { module, moduleParams },
  sourceModuleId,
  payload,
) => {
  // ...
};
```

#### TODO:
- [x] Implement
- [x] Docs

### DIFF
--- a/packages/yoshi-flow-bm-runtime/src/types.ts
+++ b/packages/yoshi-flow-bm-runtime/src/types.ts
@@ -1,15 +1,24 @@
-import { BusinessManagerModule } from '@wix/business-manager-api';
+import { ModuleId } from '@wix/business-manager-api';
 import { IBMModuleParams } from './moduleParams';
 
 export interface ModuleInitOptions {
-  module: BusinessManagerModule;
+  module: any;
   moduleParams: IBMModuleParams;
 }
 
 export type ModuleInitFn = (this: any, options: ModuleInitOptions) => void;
 
+export type ModuleConfigOptions = ModuleInitOptions;
+
+export type ModuleConfigFn = (
+  this: any,
+  options: ModuleConfigOptions,
+  sourceModuleId: ModuleId,
+  configPayload: any,
+) => void;
+
 export interface MethodOptions {
-  module: BusinessManagerModule;
+  module: any;
   moduleParams: IBMModuleParams;
 }
 
--- a/website/versioned_docs/version-4.x/business-manager-flow/overview.md
+++ b/website/versioned_docs/version-4.x/business-manager-flow/overview.md
@@ -288,18 +288,35 @@ For example, the `src/methods/some-method.ts` file, will be configured by `src/m
 
 Sets the method's `methodId`. Defaults to `<MODULE_ID>.methods.<FILE_NAME>`.
 
-### Run code in BM's `init()` phase
+### `init()` module lifecycle hook
 
-Create a file `src/moduleInit.{ts,js}`, for example:
+Create a file `src/module.{ts,js}`, for example:
 
 ```typescript
 import { ModuleInitFn } from "yoshi-flow-bm-runtime";
 
-const moduleInit: ModuleInitFn = ({ module, moduleParams }) => {
+export const init: ModuleInitFn = ({ module, moduleParams }) => {
   // ...
 };
-
-export default moduleInit;
 ```
 
 More info [here](https://github.com/wix-private/business-manager/blob/master/business-manager-api/docs/business-manager-module.md#init).
+
+### `config()` module lifecycle hook
+
+Create a file `src/module.{ts,js}`, for example:
+
+```typescript
+import { ModuleConfigFn } from "yoshi-flow-bm-runtime";
+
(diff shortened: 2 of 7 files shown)


amitdahan:
CC: @ofiratt @netanel89 @just-jeb

packages/yoshi-flow-bm-runtime/src/types.ts
```diff
@@ -1,15 +1,24 @@
-import { BusinessManagerModule } from '@wix/business-manager-api';
+import { ModuleId } from '@wix/business-manager-api';
 import { IBMModuleParams } from './moduleParams';
 
 export interface ModuleInitOptions {
-  module: BusinessManagerModule;
+  module: any;
```
ranyitz, line 5: Was this a mistake?

website/versioned_docs/version-4.x/business-manager-flow/overview.md
```diff
@@ -288,18 +288,35 @@ For example, the `src/methods/some-method.ts` file, will be configured by `src/m
 
 Sets the method's `methodId`. Defaults to `<MODULE_ID>.methods.<FILE_NAME>`.
 
-### Run code in BM's `init()` phase
+### `init()` module lifecycle hook
```
ranyitz, line 291: Not related to this PR, but I think we should separate all things related to `module` into another file, this overview is getting long :)

yarn.lock
```diff
@@ -5909,9 +5909,9 @@
     "@wix/wix-bi-logger-client" latest
 
 "@wix/web-bi-logger@^2.0.863":
```
ranyitz, line 5911: Feel a bit not related, its related to some lock file that wasn't updated?

packages/yoshi-flow-bm-runtime/src/types.ts
amitdahan: So actually - no, but I'm definitely interested to hear opinions about this, tho.
(This could have been a separate PR)

Currently we bind both `this` as well as `options.module` to be the module class instance (for most "function" conventions - methods, module init/config, ...).

Both should be typed the same. Ideally I'd use `BusinessManagerModule`, but the type itself doesn't really give anything useful since it's a class, with some private members.

Most use-cases I've seen, mutate `this` (or `module`). So `any` will allow them to not use hacks.

For example, [in `price-quotes-web`](https://github.com/wix-private/crm/blob/master/crm-financial/price-quotes-web/src/moduleInit.ts#L7) they use `this`, where they could easily just use `module`, but the `BusinessManagerModule` type doesn't allow the same code they used to run (since it's now external to the class instance).

This isn't a big deal as I believe we're slowly going to see less and less class instance mutations etc. These are mostly migrations from existing codebases where it happened.

website/versioned_docs/version-4.x/business-manager-flow/overview.md
amitdahan: Yes :) It's becoming more critical on each addition. I plan on splitting it very soon.

yarn.lock
amitdahan: I tried reverting the lock-file change and running `yarn`, it re-did these changes, so I'm guessing those were either pending from some `package.json` changes which didn't run `yarn`, or alternatively, some yarn version discrepancies between devs that may caused this.
I can omit it from the PR but I think it'll be annoying if it comes back in future-unrelated PRs.

ranyitz: no need to omit, I just wanted to verify that it's not from another branch of yours :)

PR #2679, opened 2020-07-08
Add a `files` module "hook" in BM Flow

### LINKED ISSUE #2649: BM Flow doesn't support loading external scripts in-parallel with component chunks
# Summary
In BM, users can specify external scripts to be loaded in parallel with the component's chunk/bundle using the `files` field for the [`ReactLoadableComponent`](https://github.com/wix/react-module-container/blob/master/src/react-loadable-component.js) constructor.

In BM flow, we abstract-away the call to `ReactLoadableComponent` (and the older `ReactLazyComponent`).
To support loading external scripts, we should provide some way of specifying an array of links, it should probably accept `moduleParams` as a parameter since most script links are dependant on the `moduleParams.config.topology` field.

If this was purely static configuration, I think it'd be best to have it in each page/component's `.json` config, but since it depends on the injected topology, it somehow needs to be a function.

# Proposed Solution
1. Support an additional _named_ export in each page/component's file, let's call it `getExternalScripts`, which will be a function accepting `moduleParams` and returning an array of strings:
```tsx
// src/pages/index.tsx
import React from 'react';
import { GetExternalScriptsFn } from 'yoshi-flow-bm-runtime';

export default () => <div />;

export const getExternalScripts: GetExternalScriptsFn = ({ config }) => [
  `${config.topology.somethingSomethingBaseUrl}/something.bundle.js`,
];

```

2. The same, bug with a separate file `index.externals.ts` (for example), in case splitting the exports into 2 separate bundles is too time-consuming to start with.

amitdahan:
### 🔦 Summary
Closes #2649.

Adds a new convention for extending pages & exported-components:

Adding a file `src/pages/foo.module.ts` next to an existing `src/pages/foo.tsx` page containing:

```typescript
import { FilesFn } from 'yoshi-flow-bm-runtime';

export const files: FilesFn = ({ moduleParams }) => ['..', '..', '...'];
```

Will load the returned array of strings as scripts (or styles) in parallel to loading the component chunk.


amitdahan:
Cc: @netanel89 @just-jeb

packages/yoshi-flow-bm-runtime/src/createModule.tsx
```diff
@@ -93,6 +111,34 @@ export default function createModule({
...
+            ReactLoadableComponent(
+              id,
+              loadComponent,
+              files.call(this, { module: this, moduleParams: props }),
+            ),
+          [],
+        );
+
+        return <Component {...props} />;
+      };
```
amitdahan, line 139: @just-jeb This is how I ended up doing it and it worked phenomenally! Thanks!

packages/yoshi-flow-bm/src/model.ts
```diff
@@ -125,11 +143,13 @@ export default function createFlowBMModel(cwd = process.cwd()): FlowBMModel {
     };
   };
 
-  const pages = globFiles(PAGES_PATTERN).map(getPageModel);
+  const pages = globFiles(PAGES_PATTERN)
```
ranyitz, line 146: See if we can use negate pattern (e.g. `!`)

PR #2680, opened 2020-07-09
Refactor translations. Add it to editor script.

yavorsky:
### 🔦 Summary
This is some refactor related to translations usage for controller. It includes additional features and some performance enhancements (reducing the bundle size a bit).

### Don't `await` for translations
Initially we thought it would a some performance win, but finally realized we awaiting it anyway before loading the widget, so these translations going to Widget i18Next provider.
*controller.ts*
```ts
const allTranslations = flowAPI.translations.all; // no `await`s
```

### Allowing using `t` in worker environment
Currently we are providing only object with all translations. So in controller you have to:
*controller.ts*
```ts
const translations = await flowAPI.getTranslations()
console.log(translations['app.widget.hi']);
```

It's not much convenient and we can initialize i18next instance. As an further addition would be good to include i18next instance to the platform side, so we won't include it in each viewer script.

So with proposed change controller will looks like:
*controller.ts*
```ts
const createController: CreateControllerFn = async ({
  flowAPI,
}: ControllerParams) => {
 const hey = flowAPI.translations.t('app.widget.hey');
 console.log(hey);
}
```

### Adding translations to `initAppForPage` and `editorScript` flow APIs
We had a feature request specifically for editor script, moreover we can initialize it once in `initAppForPage` and reuse in controllers.

*editor.app.ts*
```ts
export const editorReady: EditorReadyFn = async (
  sdk,
  appDefId,
  platformOptions,
  flowAPI,
) => {
  console.log(flowAPI.translations.t('app.widget.welcome'));
};
```

It could be useful for translations for gfpp button, page labels, etc, controller via editor script.

### Moved `flowAPI.mock` and `controllerConfig.mock` to editor flow side
Currently if we change something in flow API (even without breaking change), we need to update mock part also.
We had few cases when user complaining about changing it every time they update yoshi part.
Even here, we are adding `translation` prop, so it causes ts errors for spec files.

With this change, user just need to update their mocks to our ones:
```ts
import {
  controllerConfigMock,
  controllerFlowAPIMock,
} from 'yoshi-flow-editor-runtime/test';

const controllerConfig: IWidgetControllerConfig = controllerConfigMock({
      setProps: setPropsSpy,
      appParams: {
        instance: '1',
        instanceId: '1',
        appDefinitionId,
        baseUrls: {
          staticsBaseUrl: 'http://some-static-url.com',
        },
      },
    });
```

and we'll maintain it with any new changes.

### Split editor and viewer flow API to different modules
So we won't have something related to viewer script as a part of editor script bundle

> Note: We are not providing any breaking changes. `await getTranslations()` will still work for controller part


packages/yoshi-flow-editor-runtime/src/editorScript.ts
```diff
@@ -1,27 +1,49 @@
...
+  experimentsConfig,
+  artifactId,
+  translationsConfig,
+  defaultTranslations,
+}: EditorReadyOptions): PlatformEditorReadyFn => async (
   editorSDK,
   appDefinitionId,
   platformOptions,
 ) => {
+  const language = await editorSDK.environment.getLocale();
```
bordi, line 32: will it work with multilingual?

packages/yoshi-flow-editor-runtime/src/flow-api/Base.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (experimentsConfig) {
+      this.getExperiments = initExperimentsGetter(experimentsConfig);
+    } else {
+      this.getExperiments = initEmptyExperimentsGetter();
+    }
+  }
+
+  reportError: ReportError = (error) => {
+    console.warn(
+      "You are trying to report an error, but didn't configure it in `.application.json`",
```
ranyitz, line 26:
```suggestion
      "You are trying to report an error, but didn't configure sentry in `.application.json`\n\nPlease read the docs for more information https://wix.github.io/yoshi/docs/editor-flow/structure-api/app-configuration/#applicationjson",
```

It would be better if we would have headers for each property in the config so we'll be able to send to specific anchor

packages/yoshi-flow-editor-runtime/src/flow-api/EditorScript.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    this.appLoadStarted();
+  }
+
+  private appLoadStarted = () => {
+    const { appLoadStarted } = this.fedopsLogger;
+    appLoadStarted.call(this.fedopsLogger);
+    this.fedopsLogger.appLoadStarted = (...args) => {
+      console.warn(
+        "🥺 Seems like you're trying to call `fedopsLogger.appLoadStarted` and `fedopsLogger.appLoaded` in `editor.app.ts`.\nWe are already logging load events, so you can remove these calls from your project.",
```
ranyitz, line 86: This is great! 💯  A small modification

```suggestion
        "🥺 Seems like you're trying to call `fedopsLogger.appLoadStarted` and `fedopsLogger.appLoaded` in `editor.app.ts`.\nYoshi Flow Editor already logs load events, so you can remove these calls from your project.",
```

website/versioned_docs/version-4.x/editor-flow/runtime-api/flow-api.md
```diff
@@ -154,24 +154,50 @@ export default ({ flowAPI }) => {
...
-_Available for the `controller`._
-
-Returns a pending `Promise` which will be resolved with translations for the current site's language.
+To access translations object, use `flowAPI.translations`.
 
-_Since we want to start fetching translations as soon as possible, this call won't create a new network call.
-It will just return a promise for an already created call by the editor flow._
+### `t`
+Regular i18Next `t` helper already configured for your translations.
+You can specify either one key as a String or multiple keys as an Array of String. The first one that resolves will be returned.
```
ranyitz, line 165:
```suggestion
You can specify either one key as a String or multiple keys as an Array of Strings. The first one that resolves will be returned.
```

PR #2682, opened 2020-07-09
Yoshi Serverless - wait for GA

yanivefraim:
We want to wait for serverless GA before build is complete. This is for Sled test to run with the new Server that was GAed.


packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -44,5 +52,24 @@ export default async function publishServerless(config: Config) {
...
+  // Wait for Publish to Serverless to be finished
+  await retry(
+    async () => {
+      console.log('Ping to check if Service is up');
+      const res = await fetch(
+        `http://www.wix.com${getServerlessBase(getServerlessScope())}/_api_`,
+      );
+      const resText = await res?.text();
+
+      if (resText === '406') {
```
ronami, line 64: Once we have correct error pages we can test for the status code, also I think later we can have a dedicated `/health-check` route

packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -44,5 +52,24 @@ export default async function publishServerless(config: Config) {
...
+        return;
+      }
+      console.log('Service is not up, yet');
+      throw new Error('Service is not up, yet');
+    },
+    {
+      retries: 100,
```
ronami, line 71: Later we can adjust this number. Maybe lower to ~20-30 so it doesn't hold CI for too long of a time.

packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -44,5 +52,24 @@ export default async function publishServerless(config: Config) {
   await git.add('serverless/*');
   await git.commit(`deploy #serverless ${getServerlessScope()}`, '--no-verify');
   await git.push('origin', 'master');
+  // Wait for Publish to Serverless to be finished
+  await retry(
+    async () => {
+      console.log('Ping to check if Service is up');
+      const res = await fetch(
+        `http://www.wix.com${getServerlessBase(getServerlessScope())}/_api_`,
```
ronami, line 60: This might not work in `master`. We can work this out by having a dedicated `/health-check` route that can return the `SRC_MD5` it was built with.

packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -1,8 +1,13 @@
...
 import fs from 'fs-extra';
 import { Config } from 'yoshi-config/build/config';
-import { getServerlessScope } from 'yoshi-helpers/build/utils';
+import {
+  getServerlessScope,
+  getServerlessBase,
+} from 'yoshi-helpers/build/utils';
 
 export default async function publishServerless(config: Config) {
   if (!config.yoshiServer || !process.env.EXPERIMENTAL_YOSHI_SERVERLESS) {
```
ronami, line 13: It's not related to this PR but I think we should also call `publishServerless` only if we're in CI.

PR #2683, opened 2020-07-09
fix(flow-app): add isAnalyze to web worker webpack config

### LINKED ISSUE #2648: ״yoshi build --analyze״ does not work on webWorker entries
# 🙋 Feature Request

### 😯 Current Behavior

When running --analyze, it only analyzes the regular entries.

### 🤔 Expected Behavior

IF there are webWorker entries, --analyze should analyze those too.

kobiburnley:
fixes #2648 

@amitdahan can you please refer the right place to add tests for this?

Some additional notes:
* Similarity to https://github.com/wix/yoshi/pull/2295 maybe indicates that some abstraction is needed here?
* Maybe also creating a unified options interface that are passed to createWebackConfig functions


amitdahan:
Note, the `test:templates` failed because of an issue with:
- `test:templates:flow-library`
- `test:templates:library`

The issue is unrelated to this PR, we found out about it this morning.

Since those are the only checks that failed (and a missing PR label check, which I will fix by filing it as a "New Feature").

Regarding tests, not sure we'd insist on adding them in this case, but it may be easy if you want to pursue it, similar to:
- https://github.com/wix/yoshi/blob/master/test/javascript/features/analyze/analyze.test.ts
- https://github.com/wix/yoshi/blob/master/test/typescript/features/analyze/analyze.test.ts

(Maybe even use the same tests, but enhance the fixture to also have a worker bundle? Not sure)

In any case, very good job! 🥇

PR #2688, opened 2020-07-12
Yoshi serverless: handle errors

yanivefraim:
Handle errors for serverless


packages/yoshi-common/src/templates/serverless/serverless.js
```diff
@@ -17,7 +18,9 @@ const addOptionsCors = (functionsBuilder) => {
     return new FullHttpResponse({
       status: 204,
       body: {},
-      ...(process.env.NODE_ENV === 'development' ? accessControlHeaders : {}),
+      ...(process.env.NODE_ENV === 'development'
+        ? accessControlHeaders
```
yanivefraim, line 22: refactor?

packages/yoshi-serverless/src/routes/api.ts
```diff
@@ -74,14 +76,15 @@ export default route(async function () {
...
-    //   errors: [serializeError(error)],
-    // });
+    if (process.env.NODE_ENV === 'production') {
+      throw new Error('internal server error');
+    }
+    throw new Error(error);
   }
 });
+
+class CustomError extends Error {
```
yanivefraim, line 86: reuse customError?

packages/yoshi-common/src/templates/serverless/serverless.js
```diff
@@ -17,25 +15,50 @@ const addOptionsCors = (functionsBuilder) => {
...
     .addWebFunction('POST', '*', async (ctx, req) => {
-      const server = await Server.create(ctx);
-      const result = await server.handle(req);
-      const webResponse = {
-        body: result,
-        status: 200,
-        ...(isDevelopment ? accessControlHeaders : {}),
-      };
-      return new FullHttpResponse(webResponse);
+      try {
```
ronami, line 39: I think we need a similar `try {} catch {}` for `GET` requests too. WDYT?

packages/yoshi-serverless/src/routes/api.ts
```diff
@@ -49,20 +55,17 @@ export default route(async function () {
   const validation = requestPayloadCodec.decode(body);
 
   if (isLeft(validation)) {
-    return '406';
-    // return send(this.res, 406, {
-    //   errors: PathReporter.report(validation),
-    // });
+    throw new YoshiServerlessError(PathReporter.report(validation), 406);
```
ronami, line 58: Small naming comment: Since we're already in Yoshi Serverlss, I think that throwing an error called `YoshiServerlessError` doesn't mean much. Especially if we want to have other errors in the future. Wdyt about `HttpError`? Since it's like an error but also has status code.

test/typescript/features/yoshi-server/serverless-api-error-message/serverless-api-error-message.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  templateDir: __dirname,
+  projectType: 'yoshi-serverless-typescript',
+  ignoreWarnings: true,
+});
+
+describe.each(['dev'] as const)(
+  'yoshi-serverless api client to server - show an error message for development [%s]',
+  (mode) => {
+    let page: Page;
+    afterEach(() => page.close());
```
ronami, line 16: why do we need this and the `page = await browser.newPage();`?

yanivefraim: We have a global handler for page errors, which will fail the test when the error occur. This is similar to what we had in Yoshi Server error test

packages/yoshi-serverless/src/routes/api.ts
yanivefraim: We already have HttpError for serverless error, might be confusing...
Maybe `CustomHttpError`?

PR #2690, opened 2020-07-13
Fix `generate-bi-types` call

yavorsky:
### 🔦 Summary
Export a function for `generate-bi-types` module.
Currently it works, but logs an error since module has evaluation during require and exports nothing.


yavorsky:
Screenshot 2020-07-14 at 12 02 32
😎

PR #2699, opened 2020-07-15
Use the ci-build-info library

ohana54:
This PR replaces the current implementation which has knowledge about internal CI implementation details.
We now use the new `ci-build-info` library, that supports both current CI and Falcon CI, so there's no need to have that knowledge inside Yoshi.
Tests were sometimes tricky, since when running locally they behave in one way, and in CI behave in another way (due to the env pollution of CI env vars). I fixed it where possible (forced local env simulation).

cc @ranyitz

### DIFF
--- a/test/scripts.ts
+++ b/test/scripts.ts
@@ -93,7 +93,6 @@ export default class Scripts {
     testDirectory: string;
     isMonorepo: boolean;
     projectType: ProjectType;
-    packageName: string;
     yoshiBinToUse: string;
     ignoreWarnings: boolean;
   }) {
@@ -234,7 +233,6 @@ export default class Scripts {
       testDirectory: featureDir,
       isMonorepo,
       projectType,
-      packageName: newPackageJSONfileContents.name,
       yoshiBinToUse,
       ignoreWarnings,
     });
--- a/test/scripts.ts
+++ b/test/scripts.ts
@@ -4,7 +4,9 @@ import fetch from 'node-fetch';
 import fs from 'fs-extra';
 import defaultsDeep from 'lodash/defaultsDeep';
 import retry from 'async-retry';
-import { getServerlessScope } from '../packages/yoshi-helpers/build/utils';
+import globby from 'globby';
+import { testkit as ciBuildInfoTestkit } from '@wix/ci-build-info';
+import { getDevServerlessScope } from '../packages/yoshi-helpers/build/utils';
 import { ciEnv, localEnv } from '../scripts/utils/constants';
 import serve from '../packages/yoshi-common/serve';
 import writeJson from '../packages/yoshi-common/build/write-json';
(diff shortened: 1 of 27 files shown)


ohana54:
Thanks Ran, the test "complication" makes total sense for this type of project, we're in the same situation. I did love that there is great coverage for the use cases, it gave me a lot of confidence doing this refactor, so well done!

test/scripts.ts
```diff
@@ -203,6 +234,7 @@ export default class Scripts {
       testDirectory: featureDir,
       isMonorepo,
       projectType,
+      packageName: newPackageJSONfileContents.name,
```
yanivefraim, line 237: I think that this is redundant. We do not use it anywhere. Am I missing something? 😄

test/scripts.ts
```diff
@@ -433,7 +466,10 @@ export default class Scripts {
     callback: TestCallbackWithResult = async () => {},
     opts: ScriptOpts & { staticsDir?: string } = {},
   ) {
-    const buildResult = await this.build({ ...ciEnv, ...opts.env }, opts.args);
+    const buildResult = await this.build(
+      { ...ciEnv, ...opts.env, ...this.ciBuildInfoEnv },
```
yanivefraim, line 470: Not sure if I got this correct, but why do we add `this.ciBuildInfoEnv` only for `test` and for `build` inside `prod`? What if we  call build from within the test? Do we need this in `start`?

yanivefraim, line 470: I think that we might want to have build running with `ciBuildInfoEnv` by default, not only for `prod`. WDYT?

test/typescript/features/monorepo/basic.test.ts
```diff
@@ -38,7 +39,7 @@ describe('monorepo', () => {
   it('selective build', async () => {
     rimraf.sync(path.join(scripts.testDirectory, 'packages/*/dist'));
 
-    await scripts.build({}, ['monorepo-app']);
+    await scripts.build(localEnv, ['monorepo-app']);
```
yanivefraim, line 42: not sure about it, but I think that `build` was intended to be used as in CI env. If this is correct, maybe we should add it by default to `build` command

ranyitz: Yes, most builds runs in CI and it was intended to be tested there. For example, public path is being assigned in CI and we want to test that.

test/scripts.ts
ohana54: You're not missing anything, I forgot to remove it in one of the refactors :)

test/scripts.ts
ohana54: It depends on the context - `start` for example, is not relevant in a CI environment, so it might be confusing/wrong to apply it to everything by default. Does that make sense?

test/typescript/features/monorepo/basic.test.ts
ohana54: I thought that was the purpose of `prod`, which is basically `build` in CI env, no?
`build` seems like a low-level function being used by `dev` and `prod`, so let me know what you want to do with it - I think it's important that the env will be explicit, as before it was affected by the environment that it ran in (for example, I couldn't run this test locally in a simulated CI env).
Thoughts?

yanivefraim: I now see that we had `ciEnv` on build from `prod` only. So this might be a refactor for later. Let's leave it this way, and fix in another PR.

PR #2726, opened 2020-07-16
auto release

kobiburnley:
Adds a new npm script 'manual-release':

Bumps minor version: lerna version
Updates changelog
Creates 1 commit for version bump + changelog updates

`npm run manual-release`:

```
$ ts-node scripts/release.ts
Bumping version with lerna version...
Generating & prepending changelog with lerna-changelog...
✔ New version & changelog were created: v4.120.0
To publish this version push with tags: git push --follow-tags origin master
✨  Done in 9.90s.
```


scripts/release.ts
```diff
@@ -0,0 +1,62 @@
+import { promises as fs } from 'fs';
+import execa from 'execa';
+import chalk from 'chalk';
+import { prependChangelog } from 'changelog-prepend/lib/prependChangelog';
```
ranyitz, line 4:
```suggestion
// https://github.com/wix-a/changelog-prepend
import { prependChangelog } from 'changelog-prepend/lib/prependChangelog';
```

ranyitz: It's not critical but I would prefer to get the `main` and not a specific file from the library:

e.g. 

```
import { prependChangelog } from 'changelog-prepend';
```

You can do this by changing [the `main` field in `package.json`](https://github.com/wix-a/changelog-prepend/blob/4cbf58d8bfcaefe66840a293920a2714f3baccf9/packages/changelog-prepend/package.json#L5)

scripts/release.ts
```diff
@@ -0,0 +1,62 @@
...
+process.env.GITHUB_AUTH = process.env.GITHUB_AUTH || process.env.github_token;
+
+const LERNA_BACKUP = 'lerna.backup.json';
+const LERNA = 'lerna.json';
+
+main();
+
+async function main() {
+  try {
+    await execa('cp', [LERNA, LERNA_BACKUP], {});
```
amitdahan, line 15: Replace with `fs` call

scripts/release.ts
```diff
@@ -0,0 +1,62 @@
+import { promises as fs } from 'fs';
+import execa from 'execa';
+import chalk from 'chalk';
+import { prependChangelog } from 'changelog-prepend/lib/prependChangelog';
+
+process.env.GITHUB_AUTH = process.env.GITHUB_AUTH || process.env.github_token;
+
```
ranyitz, line 7: I don't know if this is handled in the library or not, but I would throw an error in case this wasn't supplied.

scripts/release.ts
```diff
@@ -0,0 +1,62 @@
...
+      `Generating & prepending changelog with ${chalk.bold(
+        'lerna-changelog',
+      )}...`,
+    );
+
+    await prependChangelog({
+      prevVersion,
+    });
+
+    await execa('rm', [LERNA_BACKUP]);
```
amitdahan, line 45: Same

CI_RELEASE.md
```diff
@@ -0,0 +1,19 @@
+## CI Release
+
+### npm run release --if-preset
+
+CI will run this npm script, which is resposible for:
+
+- Bumping version `lerna version --bump minor --exact --no-push --yes`
```
amitdahan, line 7: Move this to `CONTRIBUTING.md`

scripts/release.ts
kobiburnley: lerna-changelog exit code is 0, but the output is handled by the library, and it throws an error:
https://github.com/wix-a/changelog-prepend/blob/master/packages/changelog-prepend/src/generateNewChangelog.ts#L15

kobiburnley: done https://github.com/wix/yoshi/pull/2726/files#diff-1d7f42573d5c3507c48e94f5ed470cdfR16

package.json
```diff
@@ -22,6 +22,7 @@
     "// publishing": "-----------------------------------------------------------------------------------",
     "changelog": "lerna-changelog",
     "createVersion": "node scripts/createVersion",
```
amitdahan, line 24: Do we need to keep the old script?

scripts/release.ts
```diff
@@ -0,0 +1,61 @@
...
+      '--bump',
+      'minor',
+      '--no-push',
+      '--exact',
+      '--yes',
+    ]);
+
+    const { version: nextVersion } = JSON.parse(
+      (await fs.readFile('lerna.json')).toString('utf8'),
+    );
```
amitdahan, line 34: We can use `readJson` from `fs-extra` to simplify this, wdyt?

(Also, this isn't a `require` since we don't want a cache to be involved since we're re-reading the same file? Just to make sure I understand)

amitdahan:
Re-approved :) Nothing is urgent so feel free to merge!

PR #2728, opened 2020-07-16
require babel helpers instead of inlining them

roymiloh:
This PR makes [babel helpers](https://babeljs.io/docs/en/babel-plugin-transform-runtime#helper-aliasing) to be `require`ed instead of inlining them in every file, so the bundle size will be reduced. It's still experimental - we'd like to check it on a few apps before we open it to all.


kobiburnley:
@amitdahan any extra tests required here?

packages/babel-preset-yoshi/index.js
```diff
@@ -90,7 +90,7 @@ module.exports = function (api, opts = {}) {
         requireDefault('@babel/plugin-transform-runtime'),
         {
           // 2 options blow are usually handled by polyfill.io.
-          helpers: false,
+          helpers: process.env.EXPERIMENTAL_BABEL_LOADER === 'true',
```
amitdahan, line 93:
```suggestion
          helpers: process.env.EXPERIMENTAL_BABEL_SHARED_HELPERS === 'true',
```

packages/babel-preset-yoshi/index.js
```diff
@@ -90,7 +90,7 @@ module.exports = function (api, opts = {}) {
         requireDefault('@babel/plugin-transform-runtime'),
         {
           // 2 options blow are usually handled by polyfill.io.
-          helpers: false,
+          helpers: process.env.EXPERIMENTAL_BABEL_SHARED_HELPERS === 'true',
```
amitdahan, line 93:
```suggestion
          helpers: process.env.EXPERIMENTAL_BABEL_LOADER === 'true',
```

amitdahan:
Spoke with @roymiloh, env var needs to be named `EXPERIMENTAL_BABEL_LOADER` so it matches with the `babel-loader` experimental flag, so renamed back to this.

packages/babel-preset-yoshi/index.js
ranyitz: I agree with @amitdahan here

amitdahan: It was later explained they want to use the same flag `EXPERIMENTAL_BABEL_LOADER` for a few changes related to the babel preset.
I think that's fair, although it may be further improved by maybe extracting the env var check to a function that'll be better understood maybe?

ranyitz: This is very unclear. I'm totally cool with having 2 flags which are being combined on the userland.

This is (unfortunately) becoming an API of our babel preset that no-one can promise will go away. Let's use different names for this.

PR #2729, opened 2020-07-16
Fix BI integration for exported-components in BM flow

amitdahan:
### 🔦 Summary
I accidentally forgot to *duplicate* my work on the BI integration in the exported-component code generation template.

This fixes that and also prevents future incidents.


packages/yoshi-flow-bm/src/component.ts
```diff
@@ -0,0 +1,44 @@
...
+        )}),\n`
+      : ''
+  }
+  ${addSentry ? `createSentryProvider('${model.config.sentry?.DSN}'),` : ''}
+  ${addFedops ? `createFedopsProvider('${componentId}'),` : ''}
+  ${addBI ? `createBIProvider(initSchemaLogger),` : ''}
+]);`;
+};
+
+export default generateComponentCode;
```
ranyitz, line 44: This is very nit-picky comment, so feel free to merge without listen to it.

I think that a named export would work better here, as the name of the file is `component` and not `generateComponentCode`. In general named export are a good bet in very high percentage of the cases.

PR #2730, opened 2020-07-16
Add `@wix/wix-i18n-config` integration to BM flow

amitdahan:
### 🔦 Summary
Adds a way to opt-in to i18n integration.

By adding the following config to `.module.json`:

```json
{
	"translations": {
		"default": "en"
	}
}
```

Your pages & components will be wrapped with an initialized `<I18nextProvider />` (thanks @vybu 🥇!) to be consumed as follows:

```typescript jsx
import { useTranslation } from 'yoshi-flow-bm-runtime';

export default () => {
	const [t] = useTranslation();

	return <div>{t('translation.key')}</div>;
};
```

Translation files will be loaded by convention from `src/assets/locale/messages_{LOCALE}.json`.

Note: 
I enabled Suspense by default. I'm open to hear opinions.
It's on by default for `@wix/wix-experiments-react` so I wanted to match that.

```json
{
	"translations": {
		"default": "en",
		"suspense": false
	}
}
```
Will disable it.

### DIFF
--- a/packages/yoshi-flow-bm-runtime/src/wrapComponent.tsx
+++ b/packages/yoshi-flow-bm-runtime/src/wrapComponent.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentType, useMemo } from 'react';
+import React, { ComponentType, useMemo, Suspense } from 'react';
 import ModuleProvider, { IBMModuleParams } from './hooks/ModuleProvider';
 
 interface AdditionalProps {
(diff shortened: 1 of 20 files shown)


packages/yoshi-flow-bm-runtime/src/wrapComponent.tsx
```diff
@@ -78,14 +78,18 @@ export default function wrapComponent<P extends {}>(
...
     return (
-      <ModuleProvider moduleParams={moduleParams}>
-        {deps.reduce(
-          (children, Provider) => (
-            <Provider>{children}</Provider>
-          ),
-          <Component {...((restProps as unknown) as P)}>{children}</Component>,
-        )}
-      </ModuleProvider>
+      <Suspense fallback="SUSPENDEDDDDDDD">
```
amitdahan, line 81:
```suggestion
      <Suspense fallback="">
```

PR #2735, opened 2020-07-19
experimental babel unambiguous source type

roymiloh:
Our codebase (unfortunately) mix cjs and esm modules. This PR is going to [solve](https://babeljs.io/docs/en/options#sourcetype) that ambiguousness while we're working on converting cjs parts to esm, so we'll be able to remove this configuration in the near future.


ranyitz:
@roymiloh Out of interest, what project are you referring to?

I guess that it relates to https://github.com/wix/yoshi/pull/2728 that knows how to add `require` instead of `import`.

I think it worth understanding the amount of work required to remove the `cjs` modules, otherwise people would continue to use this ENV configuration and we'll found ourselves supporting this practically forever.

roymiloh:
We only use it in thunderbolt, it's not publicly exposed (I mean, in yoshi docs). That's the fastest solution, we want to see that everything goes well with the ts-loader -> babel-loader move. We'll later find all cjs modules in thunderbolt, convert them gradually and remove this config.

roymiloh:
**UPDATE:** unfortunately I found out that it's not thunderbolt modules only (that I already converted), but some santa ones as well (under `externalUnprocessedModules`), and we're not going to convert them soon.

roymiloh:
PR updated. See main description for explanation.

PR #2736, opened 2020-07-19
Add Fedops integration to BM Flow

amitdahan:
### 🔦 Summary
Adds ability to opt-in to Fedops integration, simply by "having" a `fedops.json`.

Initial template `fedops.json` will look like:

```json
[
  {
    "appName": "{%projectName%}.pages.index",
    "notifications": {
      "slack_channels": [
        "{%projectName%}-urgent"
      ]
    },
    "deploy_annotations": [
      "{%projectName%}"
    ],
    "ignore_first_request_roundtrip": true
  }
]
```

And usage will look like:

```typescript jsx
// src/pages/index.tsx
import { useFedops } from 'yoshi-flow-bm-runtime';

export default () => {
  const fedops = useFedops();

  useEffect(() => {
    fedops.appLoaded();
  }, []);

  return <div>...</div>;
};
```

### Caveats
1. We still don't support dynamic template generation like Editor Flow does, so this should be enough to get users started.

1. This has the potential to add potentially unwanted code into users' bundles.
For example, having a `fedops.json` file doesn't necessarily mean I have fedops calls from within one of my exported-components.

    I'm not sure whether this is an issue or not.

    This can maybe be improved upon by parsing the `fedops.json` file and seeing which `app_name`s are included, and only apply it selectively (we currently support fedops in pages & exported-components, but not all are necessarily mentioned in user-customizable `fedops.json` file).


packages/create-yoshi-app/templates/flow-bm/typescript/fedops.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "appName": "{%projectName%}.pages.index",
+    "notifications": {
+      "slack_channels": [
+        "{%projectName%}-urgent"
```
roymiloh, line 6: I think it'd be more accurate/relevant to put the team name there. I don't think you want to have a channel per project.

ranyitz: Good point. We currently don't have a place on the product where something like an "ownership" tag is being used. Does it worth asking for that information during project generation?

Because it a represent of a team and can be used for creating the slack channel in this case, and other things in the future. (For example, we can specify it as the default owner on petri configuration so new specs will be created using this)

amitdahan: It does make more sense, our other templates do 2 things:
1. Adding a slack channel per project
1. Use a default `"fedops-urgent-default"` slack channel

These are the alternatives if we don't want to add a prompt to the generation.

ranyitz: Let's do the same thing the editor-flow does and open an issue to discuss those ideas (either prompts, a different default etc...)

website/versioned_docs/version-4.x/business-manager-flow/runtimes.md
```diff
@@ -0,0 +1,26 @@
+---
+id: runtimes
```
ranyitz, line 2: Not critical, but I think "runtimes" is a bit too generic. "runtime-api" is bit better IMHO

https://wix.github.io/yoshi/docs/editor-flow/runtime-api/yoshi-flow-editor-runtime

amitdahan: Renamed to `Runtime API`, splitting to multiple "sources" of runtime will be done when we have more things here (or in The Overhaul™️).

PR #2752, opened 2020-07-20
Yoshi Serverless: deploy to Yoshi Serverless sandbox

yanivefraim:
Deploy new scopes directly to our Yoshi Serverless Sandbox environment


packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -50,8 +50,12 @@ export default async function publishServerless(config: Config) {
   }
   await git.cwd(path.resolve('temp'));
   await git.add('serverless/*');
-  await git.commit(`deploy #serverless ${getServerlessScope()}`, '--no-verify');
+  await git.commit(`deploy ${getServerlessScope()}`, '--no-verify');
   await git.push('origin', 'master');
+  // deploy to Yoshi Serverless sandbox
+  fetch(
```
ronami, line 56: I think we should wait for a response

ronami: We should also check the response code once we know it returns 200 when successful.

PR #2761, opened 2020-07-21
Support stylable manifest plugin [EXPERIMENTAL_STYLABLE_LOADER]

NitayRabi:
### 🔦 Summary

- Add support for StylableManifestPlugin
- Allows working with stylable-loader 

### Motivation 

- We want to support dynamic imports in stylable using `stylable-loader` (this is a blocker for migrating any more components using Stylable in Thunderbolt) 
- When we use `stylable-loader`, we lose the abillity to use StylableMetadata plugin which is crucial both for integration with the StylablePanel in the editor, and with applying customisations in the Viewer (Thunderbolt in this case) 
- StylableManifestPlugin supplies a way to collect all information about a stylesheet and package structure needed to support both the Editor and Viewer integrations. (outputs a JSON which is used by Viewer / Editor) 

### Convention Usage

- To support "per component" loading and gather the information, StylableManifestPlugin filters and collect component ids based on convention supplied by `filterComponents` & `getCompId` configurations.
- At this time while consulting with @ronami we decided that exposing this to consumer configuration is pre-mature and using hardcoded approach would achieve what we need while this feature is experimental. 

 

### 🗺️ Test Plan

Currently under `EXPERIMENTAL_STYLABLE_LOADER` flag

### DIFF
--- a/packages/yoshi-common/src/webpack.config.ts
+++ b/packages/yoshi-common/src/webpack.config.ts
@@ -172,6 +172,36 @@ const getCommonStylbleWebpackConfig = (name: string) => ({
   resolveNamespace: resolveNamespaceFactory(name),
 });
 
+const getStylableManifestPlugin = (name: string) => {
+  try {
+    const { StylableManifestPlugin } = require('@stylable/webpack-extensions');
+    // expected to be installed on the project that tests this experimental feature
+    // eslint-disable-next-line import/no-extraneous-dependencies
+    return new StylableManifestPlugin({
+      package: {
+        name,
+        version: process.env.ARTIFACT_VERSION || '0.0.0',
+      },
+      outputType: 'fs-manifest',
+      resolveNamespace: resolveNamespaceFactory(name),
+      filterComponents(resourcePath: string) {
+        return resourcePath.endsWith('.component.st.css');
+      },
+      getCompId(resourcePath: string) {
+        return path.basename(resourcePath).replace(/\.component\.st\.css$/, '');
+      },
+      getOutputFileName(contentHash: string) {
+        return `${name}.${contentHash}.metadata.json`;
+      },
+    });
+  } catch (e) {
+    console.error(
+      'Failed creating stylable manifest plugin. \n `@stylable/webpack-extensions` should be installed to support this feature',
+    );
+    throw e;
+  }
+};
+
 export const getStyleLoaders = ({
   name,
   embedCss = false,
@@ -340,14 +370,21 @@ export const getStyleLoaders = ({
                     },
                     // expected to be installed on the project that tests this experimental feature
                     // eslint-disable-next-line import/no-extraneous-dependencies
-                    require('@stylable/experimental-loader').stylableLoaders.transform(),
+                    require('@stylable/experimental-loader').stylableLoaders.transform(
+                      {
+                        resolveNamespace: resolveNamespaceFactory(name),
+                      },
+                    ),
                   ]
                 : // https://github.com/wix/stylable/tree/master/packages/experimental-loader#ssr-exportsonly
                   [
                     // expected to be installed on the project that tests this experimental feature
                     // eslint-disable-next-line import/no-extraneous-dependencies
                     require('@stylable/experimental-loader').stylableLoaders.transform(
-                      { exportsOnly: true },
+                      {
+                        exportsOnly: true,
+                        resolveNamespace: resolveNamespaceFactory(name),
+                      },
                     ),
                   ]),
             ],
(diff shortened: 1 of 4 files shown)


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -175,6 +175,43 @@ const getCommonStylbleWebpackConfig = (name: string) => ({
...
 
+const getStylableManifestPlugin = (name: string) => {
+  try {
+    // expected to be installed on the project that tests stylable-loader experimental feature
+    // eslint-disable-next-line import/no-extraneous-dependencies
+    const { StylableManifestPlugin } = require('@stylable/webpack-extensions');
+    return new StylableManifestPlugin({
+      package: {
+        name,
+        version: process.env.ARTIFACT_VERSION || '0.0.0',
```
ranyitz, line 186: We have a certain function to get the artifact version, https://github.com/wix/yoshi/blob/f6c120f98327fcbea879f76d8b02b436267d679d/packages/yoshi-helpers/src/utils.ts#L218

The CI team works to change their usage of environment parameters, we'll make this change only on this function.

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -175,6 +175,43 @@ const getCommonStylbleWebpackConfig = (name: string) => ({
...
+      },
+      outputType: 'fs-manifest',
+      resolveNamespace: resolveNamespaceFactory(name),
+      filterComponents(resourcePath: string) {
+        return resourcePath.endsWith('.component.st.css');
```
ranyitz, line 191: What does this means?

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -175,6 +175,43 @@ const getCommonStylbleWebpackConfig = (name: string) => ({
...
+      },
+      getCompId(resourcePath: string) {
+        return path.basename(resourcePath).replace(/\.component\.st\.css$/, '');
+      },
+      getOutputFileName(contentHash: string) {
+        return `${name}.${contentHash}.metadata.json`;
```
ranyitz, line 197: Is there a possibility the name would be as follows? `@wix/my-package-name.s0m3H&Sh.metadata.json`.

packages/yoshi-common/src/webpack.config.ts
```diff
@@ -175,6 +175,43 @@ const getCommonStylbleWebpackConfig = (name: string) => ({
...
+      },
+      /**
+       * TODO - this is a workaround for stylable panel
+       * It depends on `wix-ui-santa` string.
+       */
+      packageAlias: {
+        'wix-ui-santa': `/${name}`,
```
ranyitz, line 204: Can you explain that?

packages/yoshi-common/src/webpack.config.ts
NitayRabi: Stylable needs a convention to determine which stylesheet is an actual component. I discussed with @ronami the ability to extract this into configuration and we currently decided to go on a predetermined convention.

packages/yoshi-common/src/webpack.config.ts
NitayRabi: Maybe @barak007 could elaborate on the specific reason why this is so coupled into the panel and stylable-flatten, but currently there's a coupling to `wix-ui-santa` as some magic string - this is a workaround. 

(You can see this workaround it used here (https://github.com/wix-private/thunderbolt/blob/97a52918829b351190a8c7c6d1ba90ce1ed93a6f/packages/thunderbolt-becky/src/siteAssetsModules/stylable.ts#L29) - but this wouldn't "hold water" for more then 1 library, and makes a dependancy in thunderbolt for the magic string `editor-elements`).

There's a plan to fix the `wix-ui-santa` string dependancy in Stylable, and then we can remove thos.

packages/yoshi-common/src/webpack.config.ts
ranyitz: I see, I'm Just not sure that I understand the difference between a regular `*st.css` file to `.component.st.css` file. and also why we target just the `.component.st.css` files?

packages/yoshi-common/src/webpack.config.ts
NitayRabi: Good catch!

NitayRabi:
> I can't really review this without understanding the motivation of this change and some details on how it works. Please explain how will this affect the users that use it (I guess only you at the moment) and also please explain the reason this is needed

Correct - I spoke with @ronami and didn't elaborate in PR description - added now, we can also have a short talk to make sure we are in the right direction

PR #2777, opened 2020-07-22
Allow using separate controller bundles instead of single viewer script one.

yavorsky:
### 🔦 Summary
This PR adds support for using controller url as a separate controller for widget.

In this case to prevent any bugs for existing users we're adding `createControllersStrategy` option to `.application.json`. It's mostly for internal usage and will be deprecated in future.
`createControllersStrategy` accepts 2 options: `all` and `controller`.
- `all` - doesn't change anything and we have all controllers in single viewer script
- `controller` - will create `SomeWidgetController` bundle, so users can add it to dev center and use. I'll also add it to be by default from the template in separate PR. It will also register `controller url`.

We also wrapping controller function, so users can pass flowAPI to controller method or not. If not, we'll fetch experiments, bi, tranlsation, etc according to current app configuration. (will be added in separate PR).


packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -239,9 +224,15 @@ export const createControllers = (
 
 export const createControllersWithDescriptors = (
   controllerDescriptors: Array<ControllerDescriptor>,
-) => (controllerConfigs: Array<IWidgetControllerConfig>) => {
-  // It should be called inside initAppForPage
-
+) => (
+  controllerConfigs: Array<IWidgetControllerConfig>,
+  controllerInstance?: {
```
yavorsky, line 229: `controllerFunctions`

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -239,9 +224,15 @@ export const createControllers = (
...
+    [paramKey: string]:
+      | { default: CreateControllerFn }
+      | CreateControllerFn
+      | Function;
+  },
+) => {
   const wrappedControllers = controllerConfigs.map((controllerConfig) => {
     // [Platform surprise] `type` here, is a widgetId. :(
     const { type } = controllerConfig;
```
yavorsky, line 238: `type: widgetId`

packages/yoshi-flow-editor-runtime/src/viewerScript.ts
```diff
@@ -255,6 +246,15 @@ export const createControllersWithDescriptors = (
       );
     }
 
+    if (controllerInstance?.[type]) {
+      const method = controllerInstance?.[type];
+      if (typeof method === 'function') {
+        controllerDescriptor.method = method;
```
yavorsky, line 252: `method` -> `controllerFunction`

PR #2779, opened 2020-07-22
Yoshi Serverless: adjust to work with editor-flow

yanivefraim:
some small adjustments in order to work with editor-flow


packages/yoshi-common/src/webpack.config.ts
```diff
@@ -818,7 +818,7 @@ export function createBaseWebpackConfig({
               'process.env.ARTIFACT_ID': JSON.stringify(getProjectArtifactId()),
             }
           : {}),
-        ...(useYoshiServer && process.env.EXPERIMENTAL_YOSHI_SERVERLESS
+        ...(process.env.EXPERIMENTAL_YOSHI_SERVERLESS
```
ranyitz, line 821: Is there a case when `EXPERIMENTAL_YOSHI_SERVERLESS` is true and `useYoshiServer` is false?

packages/yoshi-flow-editor/package.json
```diff
@@ -36,7 +36,9 @@
     "yoshi-config": "4.122.0",
     "yoshi-flow-editor-runtime": "4.125.0",
     "yoshi-flow-legacy": "4.125.0",
-    "yoshi-helpers": "4.125.0"
+    "yoshi-helpers": "4.125.0",
+    "yoshi-server-tools": "4.125.0",
+    "yoshi-serverless-testing": "4.125.0"
```
ranyitz, line 41: It's a bit weird to see it here, because we only use the dev-server (which in this case only needed for working locally, and not really testing)

packages/yoshi-flow-editor/src/webpack.config.ts
```diff
@@ -75,6 +75,7 @@ export function createClientWebpackConfig(
     forceEmitStats,
     exportAsLibraryName: '[name]',
     ...defaultOptions,
+    useYoshiServer: config.yoshiServer,
```
ronami, line 78: Can probably remove

packages/yoshi-flow-editor/src/webpack.config.ts
```diff
@@ -96,12 +97,13 @@ export function createServerWebpackConfig(
     isDev,
     isHot,
     ...defaultOptions,
+    useYoshiServer: config.yoshiServer,
```
ronami, line 100: Can probably remove

packages/yoshi-flow-editor/src/commands/build.ts
```diff
@@ -143,6 +149,9 @@ const build: cliCommand = async function (argv, config, model) {
 
   const [, clientOptimizedStats, serverStats] = stats;
 
+  if (inTeamCity()) {
+    await publishServerless();
+  }
```
ranyitz, line 154: If this fails, will we see the build result printed? This not really need to be part of the build, so I would add this as additional step after the build results are reported.

ranyitz: You can add some log like `publishing to serverless`, if it's not already the behavior

packages/yoshi-flow-editor/src/webpack.config.ts
```diff
@@ -96,12 +97,13 @@ export function createServerWebpackConfig(
...
   });
 
   serverConfig.entry = async () => {
     const serverEntry = '../node_modules/yoshi-flow-editor/build/server/server';
 
-    let entryConfig = config.yoshiServer
+    let entryConfig = process.env.EXPERIMENTAL_YOSHI_SERVERLESS
```
ronami, line 106: I think we should consider changing to using a config opt in instead of an env var since projects will need to pass it to both, `yoshi start` and `yoshi test`

packages/yoshi-flow-editor/src/server/server.ts
```diff
@@ -80,7 +81,10 @@ app.use('/settings/:widgetName', (req, res) => {
 
 // Launch the server
 server.start().then(
-  () => {},
+  async () => {
+    const app = bootstrap();
```
ranyitz, line 85: We already have [express's app](https://github.com/wix/yoshi/pull/2779/files#diff-ee1a590eefe5d1fd6dabdcf9523cb771R42) in an upper scope, can we use a different variable name?

packages/yoshi-flow-editor/src/server/server.ts
```diff
@@ -80,7 +81,10 @@ app.use('/settings/:widgetName', (req, res) => {
...
+    await app.start();
+  },
```
ronami, line 87: We should probably check if this project uses Yoshi Serverless and bootstrap only if it is

packages/yoshi-flow-editor/src/server/server.ts
```diff
@@ -80,7 +81,10 @@ app.use('/settings/:widgetName', (req, res) => {
 
 // Launch the server
 server.start().then(
-  () => {},
+  async () => {
+    const app = bootstrap();
+    await app.start();
```
ranyitz, line 86: We start serverless testkit even if the users doesn't really needs it.

Either use `process.env.EXPERIMENTAL_YOSHI_SERVERLESS` or a config option to do that.

packages/yoshi-helpers/src/utils.ts
```diff
@@ -200,7 +200,7 @@ export const getServerlessScope = (cwd = process.cwd()) => {
   );
 };
 
-export const serverlessPort = '3000';
+export const serverlessPort = '7777';
```
ranyitz, line 203: Is there a possibility to use a random port?

packages/yoshi-server-tools/src/serverless-publish.ts
```diff
@@ -3,14 +3,13 @@ import simpleGit from 'simple-git';
...
-import { Config } from 'yoshi-config/build/config';
 import {
   getServerlessScope,
   getServerlessBase,
 } from 'yoshi-helpers/build/utils';
 
-export default async function publishServerless(config: Config) {
-  if (!config.yoshiServer || !process.env.EXPERIMENTAL_YOSHI_SERVERLESS) {
+export default async function publishServerless() {
+  if (!process.env.EXPERIMENTAL_YOSHI_SERVERLESS) {
```
ranyitz, line 12: I believe that this logic should be visible from outside of this function, it will make it more readable. In this case I prefer readability over a small duplication, even if we'll need to have a bit more code on `build` where we call it, as it makes the `build` file much clearer without the need to step into this function and understand that it would be a `noop` in case you don't use some environment parameter.

packages/yoshi-flow-editor/src/webpack.config.ts
ranyitz: I agree. We're not using yoshi-config (for external API) but instead have [`.application.json`](https://wix.github.io/yoshi/docs/editor-flow/structure-api/application) which I think can work here. @yavorsky what do you think?

packages/yoshi-common/src/webpack.config.ts
yanivefraim: no, this is the reason I removed `useYoshiServer` in all Yoshi Serverless places

packages/yoshi-flow-editor/src/commands/build.ts
yanivefraim: we do have logs about the publish steps:
```
[12:48:59][npm run build --if-present] Publishing to Serverless
[12:48:59][npm run build --if-present] Deploy command:
[12:48:59][npm run build --if-present] deploy #serverless yoshi-serverless-poc-1-51-0
[12:48:59][npm run build --if-present] cloning git@github.com:wix-a/yoshi-serverless.git
[12:49:21][npm run build --if-present] clone complete
[12:49:26][npm run build --if-present] Ping to check if Service is up
[12:49:26][npm run build --if-present] Service is not up, yet
[12:49:28][npm run build --if-present] Ping to check if Service is up
[12:49:28][npm run build --if-present] Service is not up, yet
[12:49:30][npm run build --if-present] Ping to check if Service is up
[12:49:30][npm run build --if-present] Service is not up, yet
[12:49:36][npm run build --if-present] Ping to check if Service is up
[12:49:36][npm run build --if-present] Publish to Serverless complete
```

yanivefraim: Also, moved to be after build results. Thanks!

packages/yoshi-flow-editor/package.json
yanivefraim: hmmm, this is the testkit for running serverless server locally (in dev mode). The question is how can we avoid installing this dependency when we do not have a server...

packages/yoshi-flow-editor/src/server/server.ts
yanivefraim: Thanks, we figured this out yesterday, didn't had the time to fix it. 
Fixed now

packages/yoshi-helpers/src/utils.ts
yanivefraim: Yes, we plan to have something like that. Will do it in a different PR. Thanks

packages/yoshi-flow-editor/src/webpack.config.ts
yanivefraim: Yes, I agree. I prefer having another PR with a refactor for this env var (it will affect all flows)

packages/yoshi-common/src/webpack.config.ts
yanivefraim: I will add a  PR for refactoring all those into config

packages/yoshi-flow-editor/src/webpack.config.ts
yavorsky: yep. we have `.application.json` config here, so we can have an config option here. We are reading this module in `model.ts`, so it will be available in `webpack.config.ts` as a result. 

But I think we can try to enable it automatically if user added `yoshi-server` as a dep or maybe created some files under `api` ?

packages/yoshi-flow-editor/package.json
yavorsky: I think it's not critical on current stage

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  A deploy preview of your PR was successfully released 🖼.

  You can find it under one of the following URLs:
  * https://bo.wix.com/pages/yoshi.pr_2779/

PR #2782, opened 2020-07-22
Mobile settings support for editor flow

yavorsky:
### 🔦 Summary
This PR adds ability to maintain separate settings panel components for Dekstop and Mobile.

Screenshot 2020-07-22 at 19 47 52

By switching the mode in editor user should receive the specific settings panel for mobile part.

So with editor flow users can easily maintain both Dekstop settings and Mobile settings.

The main addition here - `Settings.mobile` module addition to the convention. So we'll generate 2 bundles - for mobile and for dekstop and then by checking the query param passed by editor, we can load needed bundle.

I also considered using `React.lazy`, but finally decided to change approach, since with React.lazy we'll have additional loading moment. In this case we need to handle loading screen by ourselves with React Suspense. I think maintaining the same loader that editor has when TPAs are loading is not optimal, so with current approach we'll just need to load needed bundle and platform will show a loader until full iframe document will be loaded, which increase UX in general.


yavorsky:
@bordi I think it's a good idea in general. But it could be a part of PR, where we'll have `getSettingsValues` on the flow side and provide it as a `flowAPI`. This proposal also should include some convention related to `componentSettings` / `componentStyles` naming, so flow will take these values by itself.

bordi:
> @bordi I think it's a good idea in general. But it could be a part of PR, where we'll have `getSettingsValues` on the flow side and provide it as a `flowAPI`. This proposal also should include some convention related to `componentSettings` / `componentStyles` naming, so flow will take these values by itself.

Got it. Cool!
The best engineers from `tpa-settings` took this to consideration and will work on it. 

image

yavorsky:
When 'the best engineers from `tpa-settings`' took something to consideration

image

packages/yoshi-flow-editor-runtime/src/SettingsWrapper.tsx
```diff
@@ -52,26 +52,25 @@ const getBiLoggerInstance = memoize(
...
-    appName: string | null;
-    translationsConfig: TranslationsConfig | null;
-    defaultTranslations: DefaultTranslations | null;
-    experimentsConfig: ExperimentsConfig | null;
-    projectName: string;
-    biLogger: OwnerBILoggerFactory;
-  },
-) => (props: SettingsWrapperProps) => {
+const SettingsWrapper = ({
+  Settings,
```
yavorsky, line 56: Just internal change, I decided it will be more convenient to have single object

packages/yoshi-flow-editor/src/__tests__/utils.test.ts
```diff
@@ -47,7 +48,7 @@ describe('addOverrideQueryParamsWithModel', () => {
     const urlWithParams = overrideParams('https://mysite.com', 'editorUrl');
 
     expect(urlWithParams).toBe(
-      `https://mysite.com/?tpaWidgetUrlOverride=WIDGET_ID=${serverUrl}/editor/comp&tpaSettingsUrlOverride=WIDGET_ID=${serverUrl}/settings/comp&widgetsUrlOverride=WIDGET_ID=${cdnUrl}compViewerWidget.bundle.js&viewerPlatformOverrides=APP_DEF_ID=${cdnUrl}viewerScript.bundle.js&editorScriptUrlOverride=APP_DEF_ID=${cdnUrl}editorScript.bundle.js&overridePlatformBaseUrls=APP_DEF_ID={"staticsBaseUrl":"${cdnUrl}"}`,
+      `https://mysite.com/?tpaWidgetUrlOverride=WIDGET_ID=${serverUrl}/editor/comp&tpaMobileUrlOverride=WIDGET_ID=https://localhost:5004/editor/comp&tpaSettingsUrlOverride=WIDGET_ID=${serverUrl}/settings/comp&widgetsUrlOverride=WIDGET_ID=${cdnUrl}compViewerWidget.bundle.js&viewerPlatformOverrides=APP_DEF_ID=${cdnUrl}viewerScript.bundle.js&editorScriptUrlOverride=APP_DEF_ID=${cdnUrl}editorScript.bundle.js&overridePlatformBaseUrls=APP_DEF_ID={"staticsBaseUrl":"${cdnUrl}"}`,
```
yavorsky, line 51: Already supported by the platform 🙂

packages/yoshi-flow-editor/src/wrappers/settingsWrapping.ts
```diff
@@ -22,31 +59,24 @@ const settingsWrapper = (
...
-        );
-
-        acc[`${component.name}SettingsPanel`] = generatedWidgetEntryPath;
+        acc[`${component.name}SettingsPanel.mobile`] = generateSettingsEntry({
+          generatedWidgetEntriesPath,
+          model,
+          component,
+          ownerBiLoggerPath,
+          componentFileName:
+            component.settingsMobileFileName || component.settingsFileName,
```
yavorsky, line 77: fallback to the original Settings for mobile mode, if user didn't configured mobile settings

website/versioned_docs/version-4.x/editor-flow/structure-api/component.md
```diff
@@ -100,4 +100,35 @@ const Settings = () => {
...
+            articleId="78c122c7-ef06-47c1-a137-07abfae7ed89"
+            Component={() => <MobileTab />}
+          />
+        </SettingsTabLayout>
+      </TpaSettingsProvider>
+    )}</WixSDK>
+  );
+};
+
+export default MobileSettings;
```
yavorsky, line 133: add a note that this feature is not enaled in dev center

packages/yoshi-flow-editor/src/server/templates/settingsApp.vm
```diff
@@ -31,6 +31,6 @@
         window.__USE_PRIVATE_SDK_MOCK__ = true;
       #end
     </script>
-    <script src="${clientTopology.staticsBaseUrl}${widgetName}SettingsPanel.bundle#if(!${debug}).min#{end}.js"></script>
+    <script src="${clientTopology.staticsBaseUrl}${widgetName}SettingsPanel#if(${isMobile}).mobile#{end}.bundle#if(!${debug}).min#{end}.js"></script>
```
yavorsky, line 34: support html file also

packages/yoshi-flow-editor/src/wrappers/settingsWrapping.ts
```diff
@@ -22,31 +59,24 @@ const settingsWrapper = (
...
+          entryFilename: `${component.name}SettingsPanel.js`,
         });
 
-        fs.outputFileSync(
-          generatedWidgetEntryPath,
-          generateSettingsEntryContent,
-        );
-
-        acc[`${component.name}SettingsPanel`] = generatedWidgetEntryPath;
+        acc[`${component.name}SettingsPanel.mobile`] = generateSettingsEntry({
```
ranyitz, line 71: I would start adding those convention into some constants file.

Constant strings like `SettingsPanel.mobile` can be duplicated and are being used from many places in the code so there is no single place when we can see them.

It's not necessarily important to do this in this PR, but just a note.

website/versioned_docs/version-4.x/editor-flow/structure-api/component.md
bordi: Good catch btw! 
Maybe add this to the docs? It will be quite useful to understand how to add support of Mobile SP. Cos now, seems this knowledge goes from father to son at Wix :) 
Basically, to show Settings GFPP in mobile mode user should create experiment, add this experiment to `client-spec-map` which overrides component fields. For example: 
https://bo.wix.com/client-spec-map-experiment-server/experiments/specs.challenges.OOIEnableMobileSettingsPanel/2936472a-a1ed-4ae5-9f71-614313a9f4e7

Kudos to @aarnoldaas

yavorsky: Maybe we can use our REST API here to update the app service model for it?
For example, we can check if mobile entry exists and update it during build on CI?

yavorsky: Added some explanation to docs. But I think it would be really cool to have some guide / article how to add new things to the current app deployed to production via client-spec-map-experiments

PR #2787, opened 2020-07-23
fixed a small bug with certificates on legacy cdn

amiryonatan:
### 🔦 Summary

Fixed a small bug in a flow used only by editor-x.
If the certificate didn't exist, the cdn server would throw error, instead of launching without a certificate


ranyitz:
My commits are verified, you can either sign your commits or instead use this to open a PR - https://github.com/wix/yoshi/compare/amiryonatan/master

PR #2789, opened 2020-07-23
restart server on r keypress

roymiloh:
This PR introduces a new functionality - when app server is up, there's a "usage" section that lets the developer to press "r" and then the server will be (hard) restarted (kill & start process).

Screen Shot 2020-07-23 at 17 22 50


packages/yoshi-common/src/dev-environment-logger.ts
```diff
@@ -147,6 +154,14 @@ export type DevEnvironmentLogger = (opts: {
   suffix: string;
 }) => void;
 
+function logUsageInstructions(state: State) {
+  if (state.AppServer) {
+    console.log(chalk.bold('Usage'));
+    console.log(` › Press ${chalk.bold('r')} to hard restart the app server`);
```
ranyitz, line 160:
```suggestion
    console.log(` › Press ${chalk.bold('r')} to restart the app server`);
```

I know that technically `hard restart` may be more correct, but I guess people are going to see this quite often and I think that `restart` will do the trick

packages/yoshi-common/src/dev-environment.ts
```diff
@@ -383,14 +384,24 @@ export default class DevEnvironment {
 
       openBrowser(actualStartUrl);
     }
+
+    createUserActionsWatcher().on('r', async () => {
+      logRestarting();
```
ranyitz, line 389: I know that the dev-environment logger is not in a good shape. But generally it would be preferred to update a state that would render this log to the screen. The reason for that is that we'll be able to replace the logging in the future.

You'll also won't need to clear the screen because the screen clearing is handled in the dev-environment-logger. Also you won't need to to take care of the `DEBUG` case when we don't want to clear the screen.

roymiloh: It requires some refactor as `State` is now defined as a map of `string` -> `ProcessState`. @ronami and I agreed to leave it as it is and refactor in the future. Wdyt?

roymiloh: I eventually added it to the `State`, recheck please.

packages/yoshi-common/src/user-actions-watcher.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    process.stdin.setRawMode(true);
+    process.stdin.setEncoding('utf8');
+
+    const CTRL_C = '\u0003';
+
+    process.stdin.on('keypress', (str) => {
+      if (str === CTRL_C) {
+        process.stdin.setRawMode(false);
+        process.exit(0);
```
ranyitz, line 23: Obviously not critical, but this should be `SIGINT` code (128+2)

```suggestion
        process.exit(130);
```

Issue #2479, opened 2020-05-26
publish yoshi's documentation under bo.wix.com using Janet


ranyitz:
After seeing https://github.com/wix-private/editor-elements/pull/822

It's seems to be quite easy to integrate [Janet](https://github.com/wix-private/janet) with our documentation site and host it under bo.wix.com.

We need to contact @illBeRoy and see how hard it's going to be. I think we should start publishing it in parallel to our current GitHub pages based solution so it'll be ready.

PR #2790, opened 2020-07-23
publish website to janet from CI

kobiburnley:
Not sure what approach to take regrading migration. It looks like we cannot publish 2 websites unless we also build twice (with different baseUrl)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  A deploy preview of your PR was successfully released 🖼.

  You can find it under one of the following URLs:
  * https://bo.wix.com/pages/yoshi.pr_2790/

amitdahan:
This closes #2479, right?

kobiburnley:
> This closes #2479, right?

Yes, thanks. I changed `publish-gh-pages` to only publish to janet, because the `baseUrl` is not the same when publishing to Gitgub pages. If it's required to publish to both, we can implement something that modified baseUrl before running the deploy command.

website/docusaurus.config.js
```diff
@@ -1,16 +1,11 @@
...
 const versions = require('./versions.json');
 const flows = require('./flows.json');
 
 module.exports = {
   title: 'Yoshi Universe',
   tagline: 'A Galaxy of toolkits to develop applications at Wix',
-  url: 'https://wix.github.io',
-  baseUrl: isPr ? '/' : '/yoshi/',
+  url: 'https://bo.wix.com/pages/yoshi',
+  baseUrl: '/pages/yoshi/',
```
ranyitz, line 8: How does Janet PR published version look like? I think that it changes the baseurl. See https://github.com/wix-private/wix-indicator/pull/7#issuecomment-657226697 for example

website/package.json
```diff
@@ -6,8 +6,9 @@
     "start": "docusaurus start",
     "build": "docusaurus build",
     "version": "docusaurus docs:version",
-    "publish-gh-pages": "docusaurus deploy",
-    "publish-surge": "npx teamcity-surge-autorelease --dist=build --verbose"
+    "publish-gh-pages": "npm run publish-janet",
+    "publish-surge": "npm run publish-janet",
```
ranyitz, line 10: I understand this was done for compatibility reasons. But let's just have a single `publish` command.

e.g. 
```
"publish": "janet deploy --ci"
```

kobiburnley: done (I named it `publish-janet` not to override npm built in `publish` command)

website/docusaurus.config.js
kobiburnley: fixed

kobiburnley:
> Regrading publish versions on PRs, I know that Janet supports it, maybe we can just add it on a different PR.

@ranyitz we should change current or add another PR build that uses the `publish-janet` script

website/docusaurus.config.js
```diff
@@ -1,16 +1,28 @@
...
+  }
+
+  const prNumber = gitRefFromEnvVar.match(prHeadMatcher)[1];
+  return parseInt(prNumber, 10);
+};
+
+const baseUrl =
+  process.env.BRANCH == null
+    ? '/pages/yoshi'
+    : `/pages/yoshi.pr_${getPrNumber()}`;
```
amitdahan, line 16: Maybe we should file an issue / let the guys @ Janet know that this would be nice to have exposed on their end, wdyt?

kobiburnley: Opened an issue on janet repo. Until it's done I think we should go with inlining this logic on our end.

PR #2791, opened 2020-07-24
Extend controller with flowAPI features.

yavorsky:
Follow-up for https://github.com/wix/yoshi/pull/2777

### 🔦 Summary
Currently we are supporting bundles per `controller`, so it's possible to add it as a `Controller URL` and consumer in `viewerScript` in different package.
But solution we have is not supporting `flowAPI` entities (experiments, bi, translations, fedops, sentry).

In this PR we are extending `flowAPI` to use real values fetched from the user.

Moreover, we want to prevent double work for cases when user already fetched experiments, initialized fedops or bi in custom `createControllers`. In this case users have to pass `appData` with `__prepopulated` property. It could contain `experiments`, `translations`, `sentryMonitor`, `fedopsLogger`, `biLogger`:

*IPrepopulatedData*
```ts
interface IPrepopulatedData {
  experiments?: ExperimentsBag;
  translations?: Record<string, string>;
  sentryMonitor?: ReturnType<IPlatformServices['monitoring']['createMonitor']>;
  fedopsLogger?: BaseLogger<string>;
  biLogger?: VisitorBILoggerFactory;
}
```

So in `createControllers` you can just pass it as an argument:
```ts
export const createControllers = (controllerConfigs, controllerFunctions) => {
  return controllerConfigs.map((controllerConfig) => {
    const { type: widgetId } = controllerConfig;
    const appData = {
      __prepopulated: {
       experiments: fetchedExperiments,
       biLogger: initedBiLogger,
       translations: fetchedTranslations,
      }
    }
    controllerFunctions[widgetId]({ controllerConfig, appData });
  })
};
```

This change will allow gradual migration, even if the viewer script was built with different from yoshi editor flow configuration.


packages/yoshi-flow-editor-runtime/src/utils.ts
```diff
@@ -127,3 +135,78 @@ export const biLoggerFromProps = ({
     util: biUtil,
   };
 };
+
+let isCSRLoaded = false;
```
ranyitz, line 139: I don't like the fact that we have stateful `utils` module. What do you say about exporting `wrapUserController` including this function to another file?

packages/yoshi-flow-editor/src/wrappers/templates/ControllerEntry.ts
```diff
@@ -1,12 +1,66 @@
...
+  var biConfig = ${({ biConfig }) =>
+    biConfig ? JSON.stringify(biConfig) : 'null'};
+
+  var defaultTranslations = ${({ defaultTranslations }) =>
+    defaultTranslations ? JSON.stringify(defaultTranslations) : 'null'};
+
+  ${({ visitorBiLoggerPath }) =>
+    visitorBiLoggerPath
+      ? `import biLogger from '${visitorBiLoggerPath}'`
+      : 'var biLogger = null'};
```
ranyitz, line 54: I'm pretty sure that TypeScript knows how to handle this import even though it's not a top level import. But it still looks a bit weird.

yavorsky: I'm not sure I got it. Can you please explain a bit, so I'll add it to notes and update in future ?

PR #2801, opened 2020-07-27
Add `react-custom-scroll` to be transpiled for editor flow projects.

yavorsky:
### 🔦 Summary
Currently, `react-custom-scroll` doesn't have a ES6 code in the output `dist`. But it use to have. The issue were detected after releasing the `4.x` version and fixed: https://github.com/wix-playground/react-custom-scroll/commit/b3b5c4daa053ee44e6425254877be1cbbb182934

`wix-base-ui` has pretty modern ASCII art on their README 😎, but still using a bit outdated `react-custom-scroll` version: https://github.com/wix-private/wix-base-ui/blob/master/package.json#L134 🙇‍♂️
They can't change it to `4.x` b/c of breaking changes, so in fact using the version with a bug.

Users usually fixing the bug by excluding the specific package from transpile ignore. Since we control yoshi config and user's have nothing to do with it, we can provide a fix specifically for this module.

I won't add some project specific workarounds, but if we'll think in the scope of `react-custom-scroll` - is something that all OOI apps have, we just providing a fix out of the box for everyone!


yavorsky:
@bordi Worked on it for the last decade!

bordi:
> @bordi Worked on it for the last decade!

No pain - no gain, bro

PR #2802, opened 2020-07-27
Algolia search bar eject

kobiburnley:
depends on #2797


website/dev-server-proxy-plugin.js
```diff
@@ -0,0 +1,27 @@
...
+      }
+
+      return {
+        devServer: {
+          ...config.devServer,
+          after: function (app) {
+            app.use(
+              '/1/indexes/:aestrik/queries',
+              createProxyMiddleware({
+                target: 'http://localhost:3001',
```
amitdahan, line 18: Is this `doc-search`'s default port when running locally?

kobiburnley: @amitdahan it's not the default, but usually janet local will run on 3000, and then doc-search will use automatically use 3001. I wonder what I can do here to guarantee an address where doc-search is at. I can only think on using some local service discovery like consul.io

PR #2803, opened 2020-07-27
Add controller url to dev center registration

yavorsky:
### 🔦 Summary
This PR just adds 2 things:
- Controller URL to the widget url (local overrides already working in platform)
- `"createControllersStrategy": "controller"` - enable our viewer script to use controllers from the 2nd `createControllers` argument instead of including it into a bundle. We are going to ignore this option in future and use `controller` strategy by default.


yavorsky:
But in general it's a good practice to use `controller url`. We won't include all controllers into viewer script bundle and moreover all new apps will be configured correctly.
For bolt it's not optimised well and we'll fetch all controller bundles anyway, but for TB it is and for example if you'll have a page and a widget, you won't load a page controller bundle until you really need.

It's not bringing any cons in any case and just adding us a way to configure apps in dev center correctly on early stage

ranyitz:
@yavorsky I totally understand and agree with you on that.

My point is, that if we plan on the default to be this. Maybe instead of adding it to the templates (effectively creating project that will contain this in the future) we can just test it for now and just change the default later?

I'm not sure if we'll be able to change the default under the hood as it requires dev-center configuration, but if we can it's preferred.

